PR #6, opened 2020-02-21
CLOUDP-57719: Add Basic Automation Config Struct

chatton:
This PR introduces the starting point for the Automation Config struct and builder.


antonlisovenko:
Good stuff, we don't need to merge anything here as in case with Ops Manager so we can just create a struct we need 👍

pkg/automationconfig/automation_config_builder.go
```diff
@@ -0,0 +1,89 @@
...
+	b.name = name
+	return b
+}
+
+func (b *Builder) SetMongoDBVersion(version string) *Builder {
+	b.mongodbVersion = version
+	return b
+}
+
+func (b *Builder) SetAutomationConfigVersion(version string) *Builder {
```
rodrigovalin: version should be an int, even if it will be transformed into a string by the automationconfig, that's a detail that the caller don't care about. a numeric attribute should be represented by a number

pkg/automationconfig/automation_config_builder.go
```diff
@@ -0,0 +1,89 @@
...
+	b.version = version
+	return b
+}
+
+func (b *Builder) Build() AutomationConfig {
+	hostnames := make([]string, b.members)
+	for i := 0; i < b.members; i++ {
+		hostnames[i] = fmt.Sprintf("%s-%d.%s", b.name, i, b.domain)
```
LouisPlisso: should we have a function to create hostname: we're already duplicating that code

pkg/automationconfig/automation_config_builder.go
```diff
@@ -0,0 +1,89 @@
...
+		Version:   b.version,
+		Processes: processes,
+		ReplicaSets: []ReplicaSet{
+			{
+				Id:              b.name,
+				Members:         members,
+				ProtocolVersion: "1",
+			},
+		},
+		Auth: DisabledAuth(),
```
LouisPlisso: do we plan to have this implemented later?

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -79,6 +81,22 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	// TODO: Read current automation config version from config map
```
rodrigovalin: I would not even care about doing this in the controller. This automation config implementation is completely independent of kubernetes and as such we should be able to get to a comfortable level with it only by unit testing it.

In fact the automationconfig Builder should return a struct that we will pass to the next "stage". That "next stage" will be in charge of creating the ConfigMap, as we have discussed... but that's should be left for a future patch

pkg/automationconfig/automation_config_builder.go
chatton: yes! This is just to provide a valid disabled configuration to start with

pkg/automationconfig/automation_config_builder.go
chatton: yes good point, will add a small helper function

pkg/automationconfig/automation_config_builder.go
chatton: very good point, will update 👍

PR #7, opened 2020-02-24
Create config map from automation config

chatton:
This PR takes the fully constructed Automation Config and places it into a ConfigMap.

This PR also adds a small mocking client which can hold any `runtime.Object`


deploy/crds/mongodb.com_mongodbs_crd.yaml
```diff
@@ -31,19 +31,18 @@ spec:
         spec:
           description: MongoDBSpec defines the desired state of MongoDB
           properties:
-            kind:
-              description: Kind defines which kind of MongoDB deployment the resource
-                should create
-              type: string
```
LouisPlisso: is this a manual edit or automatically generated?

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -107,9 +108,46 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
+	}
+
+	return configmap.Builder().
+		SetName(mdb.ConfigMapName()).
+		SetNamespace(mdb.Namespace).
+		SetField(automationConfigKey, string(acBytes)).
+		Build(), nil
+}
+
 func getDomain(service, namespace, clusterName string) string {
```
LouisPlisso: do we want to export this function so that we can use it in enterprise package?

pkg/kube/configmap/configmap.go
```diff
@@ -42,6 +44,21 @@ func (c Client) Create(cm corev1.ConfigMap) error {
 	return nil
 }
 
+// CreateOrUpdate will either Create the configmap if it doesn't exist, or Update it
+// if it does
+func (c Client) CreateOrUpdate(cm corev1.ConfigMap) error {
```
LouisPlisso: no unit tests?

deploy/crds/mongodb.com_mongodbs_crd.yaml
chatton: automatically generated with `operator-sdk generate crds`

rodrigovalin:
where are the tests?

pkg/kube/client/client.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+}
+
+type Client interface {
+	k8sClient.Client
+	CreateOrUpdate(obj runtime.Object) error
+
+}
+
+type client struct {
```
rodrigovalin: Too many client, Client, NewClient, NewMockedClient

pkg/kube/client/mocked_client.go
```diff
@@ -0,0 +1,88 @@
...
+
+func alreadyExistsError() error {
+	return &errors.StatusError{ErrStatus: metav1.Status{Reason: metav1.StatusReasonAlreadyExists}}
+}
+
+func NewMockedClient() k8sClient.Client {
+	return &mockedClient{backingMap: map[reflect.Type]map[k8sClient.ObjectKey]runtime.Object{}}
+}
+
+func getObjectKey(obj runtime.Object) k8sClient.ObjectKey {
```
rodrigovalin: Can we use the kubernetes client implementation of this?

pkg/controller/mongodb/mongodb_controller.go
chatton: this small function might be overkill, but it could worth while. I think we should wait until we have more relevant functions that can be grouped into a package before we export it though.

PR #10, opened 2020-02-26
Fix logger and stateful set construction

chatton:
This PR fixes errors while reconciling the MongoDB resource

1. The logger needed to be configured globally to display error messages in the reconciliation loop
2. podSpec.labels need to match label selectors

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -2,11 +2,10 @@ package main
 
 import (
 	"fmt"
-	"go.uber.org/zap"
-	"os"
-
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/apis"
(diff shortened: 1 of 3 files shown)


cmd/manager/main.go
```diff
@@ -2,11 +2,10 @@ package main
 
 import (
 	"fmt"
-	"go.uber.org/zap"
-	"os"
-
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/apis"
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/controller"
+	"go.uber.org/zap"
```
LouisPlisso: we need to have a gofmt and goimports commit hook, so that these don't move around at each PR

PR #12, opened 2020-02-26
CLOUDP-58090: Create StatefulSet that starts agent image

chatton:
This PR configures the stateful set to use an agent image specified by the `AGENT_IMAGE` environment variable.

The stateful set is configured with default resource requirements. Sample StatefulSet configuration

```bash
  spec:
    podManagementPolicy: OrderedReady
    replicas: 3
    revisionHistoryLimit: 10
    selector:
      matchLabels:
        dummy: label
    serviceName: ""
    template:
      metadata:
        creationTimestamp: null
        labels:
          dummy: label
      spec:
        containers:
        - command:
          - agent/mongodb-agent
          - -cluster=/var/lib/automation/config/automation-config.json
          image: <agent-image>
          imagePullPolicy: Always
          name: mongodb-agent
          resources:
            limits:
              cpu: "1"
              memory: 500M
            requests:
              cpu: 500m
              memory: 400M
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -6,7 +6,6 @@ import (
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/controller"
 	"go.uber.org/zap"
 	"os"
-
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
@@ -31,6 +30,13 @@ func main() {
 	if err != nil {
 		os.Exit(1)
 	}
+
+	// TODO: implement mechanism to specify required/optional environment variables
+	if _, agentImageSpecified := os.LookupEnv("AGENT_IMAGE"); !agentImageSpecified {
+		log.Error("required environment variable AGENT_IMAGE not found")
+		os.Exit(1)
+	}
+
 	// get watch namespace from environment variable
 	namespace, nsSpecified := os.LookupEnv("WATCH_NAMESPACE")
 	if !nsSpecified {
@@ -49,7 +55,7 @@ func main() {
 	mgr, err := manager.New(cfg, manager.Options{
 		Namespace: namespace,
 	})
-	
+
 	if err != nil {
 		os.Exit(1)
 	}
--- a/pkg/kube/resourcerequirements/resource_requirements.go
+++ b/pkg/kube/resourcerequirements/resource_requirements.go
@@ -10,9 +10,20 @@ const (
 	resourceCpu    = "cpu"
 )
 
-// New returns a new corev1.ResourceRequirements with the specified arguments, and an error
+// Defaults returns the default resource requirements for a container
+func Defaults() corev1.ResourceRequirements {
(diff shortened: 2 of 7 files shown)


pkg/kube/resourcerequirements/resource_requirements.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+const (
+	resourceMemory = "memory"
+	resourceCpu    = "cpu"
+)
+
+// New returns a new corev1.ResourceRequirements with the specified arguments, and an error
+// which indicates if there was a problem parsing the input
+func New(limitsCpu, limitsMemory, requestsCpu, requestsMemory string) (corev1.ResourceRequirements, error) {
```
LouisPlisso: is this worth passing a struct here so that it's easier to understand what value you pass, and maybe address the use-case of passing no value and taking the default?

LouisPlisso: or maybe a builder (again) that comes with default values and you can call .setLimitsCpu(xxx) and so on

cmd/manager/main.go
```diff
@@ -6,7 +6,6 @@ import (
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/controller"
 	"go.uber.org/zap"
 	"os"
-
```
LouisPlisso: when are we getting the commit hook for gofmt and goimports?

chatton: @rodrigovalin  is working on it now I believe!

pkg/kube/resourcerequirements/resource_requirements.go
chatton: I'm on a little bit on the fence about this one. When we're not doing much (just 4 values). I like the simplicity of a single factory-style function. It might make sense to introduce a builder when we need to start . configuring non-defaults here.

cmd/manager/main.go
rodrigovalin: I am, doing it right now

pkg/kube/resourcerequirements/resource_requirements.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	return corev1.ResourceRequirements{
+		Limits:   limits,
+		Requests: requests,
+	}, nil
+}
+
+// Default returns the default resource requirements for a container
+func Default() (corev1.ResourceRequirements, error) {
+	return New("1.0", "500M", "0.5", "400M")
```
rodrigovalin: Discard the returning errors of this function, `Default` should not fail as the values are known

rodrigovalin: Possibly add a unit test that checks that `Default()` return these values so there's not way we introduce errors in the `Default` impl.

I mention this because I see a few lines up there checking the resulting value of the Default function

chatton: good point about removing the error in this case 👍

pkg/kube/resourcerequirements/resource_requirements.go
LouisPlisso: as discuss, un-export New for now, and we'll see later if we need a builder

PR #17, opened 2020-02-28
Creating a proper Automation Config

rodrigovalin:
### All Submissions:


agent/Dockerfile
```diff
@@ -15,4 +15,8 @@ RUN mkdir -p agent \
     && mkdir -p /var/lib/automation/config \
     && chmod -R +r /var/lib/automation/config
 
+RUN curl -LO http://downloads.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1604-4.0.6.tgz && \
+    tar xfz mongodb-linux-x86_64-ubuntu1604-4.0.6.tgz && \
+    mv mongodb-linux-x86_64-ubuntu1604-4.0.6/bin/* /usr/bin
```
LouisPlisso: I'd keep only the mongo shell binary, not all the binaries

agent/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:16.04
```
LouisPlisso: why?

pkg/automationconfig/automation_config_builder.go
```diff
@@ -50,13 +53,23 @@ func (b *Builder) SetName(name string) *Builder {
 	return b
 }
 
+func (b *Builder) AddVersion(version MongoDbVersionConfig) *Builder {
+	for idx, _ := range version.Builds {
+		if version.Builds[idx].Modules == nil {
+			version.Builds[idx].Modules = make([]string, 0)
+		}
+	}
+	b.versions = append(b.versions, version)
```
LouisPlisso: this needs to be inside the loop

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -194,16 +236,52 @@ func buildStatefulSet(mdb mdbv1.MongoDB) (appsv1.StatefulSet, error) {
 		},
 	}
 
+	dataVolume, dataVolumeClaim := buildDataVolumeClaim()
+	v := statefulset.CreateVolumeFromConfigMap("automation-config", "example-mongodb-config")
```
LouisPlisso: are we not just using an emptyDir volume for sharing the config file between the side car and the main container?

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -194,16 +236,52 @@ func buildStatefulSet(mdb mdbv1.MongoDB) (appsv1.StatefulSet, error) {
...
 		SetPodTemplateSpec(podSpecTemplate).
 		SetNamespace(mdb.Namespace).
 		SetName(mdb.Name).
 		SetReplicas(mdb.Spec.Members).
 		SetLabels(labels).
 		SetMatchLabels(labels).
+		AddVolumeMount(agentName, vm).
+		AddVolumeMount(mongodbName, vm).
+		AddVolume(v).
+		AddVolumeMount(agentName, dataVolume).
```
LouisPlisso: oh I see, you are using the dataVolume as the shared dir for agent and mongod

I think it's better to use an emptyDir because this is guaranteed to be a local mount, and if the data directory is full, then the agent will still be able to change the config

This looks like a blocker to me

agent/Dockerfile
rodrigovalin: At one point I thought that the agent might get confused by the image it was running on, and the image that corresponds to the mongod that was expected to be running, so I changed everything to ubuntu16. Now I'm not sure if this is a situation we should care about, and I'll remove it for now.

pkg/automationconfig/automation_config_builder.go
rodrigovalin: The automation config has many `MongoDbVersionConfig`s, which in turn, have multiple `Builds`.

This function only receives 1 `MongoDbVersionConfig`, that need to be appended to the ones we have already built, by previous calls to `AddVersion`.

For example:

```golang
builder := NewBuilder()
builder.AddVersion("1.2.3").AddVersion("4.2.3").Build()
```

agent/Dockerfile
rodrigovalin: actually I think it is better if we use ubuntu16 everywhere, just to make sure this is not a variable we should care about, for now. And then we can be specific on the base images.

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -132,17 +132,46 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
 }
 
 func buildAutomationConfig(mdb mdbv1.MongoDB) automationconfig.AutomationConfig {
-	domain := getDomain(mdb.ServiceName(), mdb.Namespace, mdb.Name)
+	domain := getDomain(mdb.ServiceName(), mdb.Namespace, "")
```
LouisPlisso: maybe good for now, but we many need a TODO here

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -157,17 +186,31 @@ func buildAutomationConfigConfigMap(mdb mdbv1.MongoDB) (corev1.ConfigMap, error)
...
+		Resources:       resourcerequirements.Defaults(),
+		Command:         agentCommand,
 	}
 
+	// mongoDbCommand := []string{
+	// 	"/bin/sh",
+	// 	"-c",
+	// 	`while [ ! -f /var/lib/automation/config/mongodb-automation.conf ]; do echo "[$(date)] waiting" ; sleep 10; done; mongod -f /var/lib/automation/config/mongodb-automation.conf`,
+	// }
+	mongoDbCommand := []string{"mongod"}
```
LouisPlisso: maybe we want to put that back in

pkg/controller/mongodb/mongodb_controller.go
rodrigovalin: Actually not, because that parameter is the "cluster domain"; when passed the empty string, then the default one will be used ("cluster.local")

pkg/controller/mongodb/mongodb_controller.go
rodrigovalin: That's part of the work I'm doing next.

PR #18, opened 2020-03-02
CLOUDP-58264: Add Simple Replica Set Test

chatton:
### All Submissions:

This PR creates a simple test using the operator-sdk test framework which ensures that the operator creates the ConfigMap and StatefulSet.

The go.mod was updated to work with the testing framework but will be reverted as soon as possible.


LouisPlisso:
@chatton can you please check why the unit tests are failing: it looks like a gopath issue

go.mod
```diff
@@ -3,13 +3,46 @@ module github.com/mongodb/mongodb-kubernetes-operator
...
 	github.com/stretchr/testify v1.4.0
 	go.uber.org/zap v1.13.0
-	k8s.io/api v0.15.9
-	k8s.io/apimachinery v0.15.9
+	golang.org/x/net v0.0.0-20191028085509-fe3aa8a45271
+	k8s.io/api v0.17.2
+	k8s.io/apimachinery v0.17.2
 	k8s.io/client-go v12.0.0+incompatible
-	sigs.k8s.io/controller-runtime v0.3.0
+	sigs.k8s.io/controller-runtime v0.5.0
```
LouisPlisso: is this intentional?

we should stick to 1.16 max
also client-go is now following k8s versioning

go.mod
```diff
@@ -3,13 +3,46 @@ module github.com/mongodb/mongodb-kubernetes-operator
...
 )
+
+// TODO: These replace statements need to be completely removed. These exist as the operator-sdk currently requires
+// them. Once they are fully removed, we should only include "require" statements in go.mod
```
LouisPlisso:
```suggestion
// them. Once they are fully removed, we should not include "require" statements in go.mod
```

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 const (
-	automationConfigKey   = "automation-config"
+	AutomationConfigKey   = "automation-config"
 	agentName             = "mongodb-agent"
 	mongodbName           = "mongod"
 	agentImageEnvVariable = "AGENT_IMAGE"
```
LouisPlisso: we need to be consistent: either all capitalised or none, no?

scripts/ci/download.go
```diff
@@ -10,7 +10,7 @@ import (
 
 // download.go uses the following environment variables:
 //   URL: The url of the file to download
-//   BINDIR: The directory which the newly downloaded file will be placed
+//   DIR: The directory which the newly downloaded file will be placed
 //   FILENAME: The name the file should have after being downloaded
```
LouisPlisso: why are we using environment variables: argument passing is much more explicit and you can always pass env variables as argument if you really really need it

scripts/ci/generate_k8.sh
```diff
@@ -6,4 +6,4 @@ set -xeo pipefail
 # GOROOT needs to be explicitly set otherwise we get the error
 # time="2020-02-24T17:52:08Z" level=info msg="Running deepcopy code-generation for Custom Resource group versions: [mongodb:[v1], ]\n"
 # deepcopy.go:885] Hit an unsupported type invalid type for invalid type, from ./pkg/apis/mongodb/v1.MongoDB
-GOROOT="/opt/golang/go1.13" operator-sdk generate k8s
+operator-sdk generate k8s
```
LouisPlisso: please adjust the comment if GOROOT is no more needed

pkg/controller/mongodb/mongodb_controller.go
chatton: this one in particular is being imported in the e2e test.

PR #26, opened 2020-03-13
CLOUDP-58891: Copy Agent Readiness Probe in InitContainer

chatton:
### All Submissions:

This PR adds an init. container which copies the readiness probe into a shared volume which the agent uses.

A new E2E test was added which scales up/down a replica set and ensures the database is reachable throughout the scaling up/down process

Todo:
* [X] Update Unit Tests


chatton:
@LouisPlisso we can leave unit Containers for now but I think we should merge the scale replica set test as it should still work just fine with readiness probe inside the agent container. That be can be pulled out into a separate pr maybe and we can leave this one for now.

PR #27, opened 2020-03-13
CLOUDP-59033: MongoDB URI Added To Status

chatton:
### All Submissions:

This PR is based on top of #26 
This PR adds a Status Subresource to the MongoDB Resource.

There were issues with concurrent modification failures in the tests so I've added some retry logic to account for this.

The E2E tests were updated to test the status was updated correctly

EVG run: https://evergreen.mongodb.com/version/5e7a091ed1fe077fc3e379f1


chatton:
https://evergreen.mongodb.com/version/5e7a091ed1fe077fc3e379f1

chatton:
@LouisPlisso  can you take a quick look when you get a chance?

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -31,6 +37,9 @@ type MongoDBSpec struct {
 
 // MongoDBStatus defines the observed state of MongoDB
 type MongoDBStatus struct {
+	MongoURI string `json:"mongoUri"`
+	Members  int    `json:"members"`
```
rodrigovalin: If this field is going to be a copy of what we have in `spec`, then it should be removed.

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -124,19 +126,52 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
 
 	if err = r.client.CreateOrUpdate(&sts); err != nil {
-		log.Warnf("error creating/updating StatefulSet: %s", err)
+		log.Infof("Error creating/updating StatefulSet: %s", err)
 		return reconcile.Result{}, err
+	} else {
+		log.Infof("StatefulSet successfully Created/Updated")
+	}
+
+	log.Infof("waiting for StatefulSet %s/%s to reach ready state", mdb.Namespace, mdb.Name)
```
rodrigovalin: Debug message maybe?

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -124,19 +126,52 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
+	set := appsv1.StatefulSet{}
+	timedOut, err := r.client.WaitForCondition(types.NamespacedName{Name: mdb.Name, Namespace: mdb.Namespace}, time.Second*3, time.Second*30, &set, func() bool {
+		return statefulset.IsReady(set)
+	})
+
+	if timedOut {
```
rodrigovalin: This kind of information is only relevant in a debugging scenario. Moreover, I think we can go with a different strategy here. We should rely on reconcile loop for the mechanics of object polling & waiting, not on our own loops with sleeps in between.

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -124,19 +126,52 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
+		return reconcile.Result{}, nil
 	}
 
 	log.Info("Successfully finished reconciliation", "MongoDB.Spec:", mdb.Spec, "MongoDB.Status", mdb.Status)
 	return reconcile.Result{}, nil
 }
 
+func (r ReplicaSetReconciler) updateStatusSuccess(mdb *mdbv1.MongoDB) error {
+	mdb.UpdateSuccess()
+	if err := r.client.Status().Update(context.TODO(), mdb); err != nil {
```
rodrigovalin: Will this mean that the object will go to error state if the Status object was not updated?

test/e2e/e2eutil.go
```diff
@@ -29,15 +29,28 @@ func RegisterTypesWithFramework(newTypes ...runtime.Object) error {
...
 }
 
-func CreateOrUpdateMongoDB(mdb *mdbv1.MongoDB, ctx *f.TestCtx) error {
-	err := f.Global.Client.Get(context.TODO(), types.NamespacedName{Name: mdb.Name, Namespace: mdb.Namespace}, &mdbv1.MongoDB{})
-	if err != nil {
-		if errors.IsNotFound(err) {
-			return f.Global.Client.Create(context.TODO(), mdb, &f.CleanupOptions{TestContext: ctx})
+// UpdateMongoDBResource applies the provided function to the most recent version of the MongoDB resource
+// and retries when there are conflicts
+func UpdateMongoDBResource(original *mdbv1.MongoDB, updateFunc func(*mdbv1.MongoDB)) error {
```
rodrigovalin: Can you add some sort of exponential backoff here?

pkg/controller/mongodb/mongodb_controller.go
chatton: The idea here was a relatively short wait period before retrying. We could of course shorten it to almost no wait period, this would end up with a lot of reconciliations consistently requeued (maybe that's not a problem though!)

pkg/controller/mongodb/mongodb_controller.go
chatton: No, at the moment there's no error state for the resource

test/e2e/e2eutil.go
chatton: I don't think we need the backoff here, this loop is solely to account for period occurrences of apiserver errors that we get when the underlying resources are being modified elsewhere.

pkg/controller/mongodb/mongodb_controller.go
rodrigovalin: In the Kubernetes environment, a blocking controller is a problem.

test/e2e/e2eutil.go
rodrigovalin: In a scenario where apiserver errors are occurring, the less we want is to overwhelm the apiserver with a burst of requests.

test/e2e/e2eutil.go
```diff
@@ -29,15 +29,28 @@ func RegisterTypesWithFramework(newTypes ...runtime.Object) error {
...
+	for i := 0; i < 3; i++ {
+		err := f.Global.Client.Get(context.TODO(), types.NamespacedName{Name: original.Name, Namespace: original.Namespace}, original)
+		if err != nil {
+			return err
+		}
+
+		updateFunc(original)
```
rodrigovalin: The original object is passed to the function, then it is being fetched from the API, then it is modified in place, then it is updated...

Unfortunatelly this means that the same object could be updated multiple times, and the updated object will be fetched from the API...

The "original" `original` object will have to be saved so every iteration is independant from the next.

test/e2e/e2eutil.go
rodrigovalin: I think the loop is not needed here. Why would the object be modified by multiple actors?

test/e2e/e2eutil.go
rodrigovalin: resolved in person

test/e2e/e2eutil.go
rodrigovalin: discussed:

- understand why there are multiple concurrent modifications, and fix that. but this 0..3 tries is tech debt.

chatton:
@rodrigovalin 

* Adjusted the get/update function to perform a single operation
*  No longer polling for stateful set (also removed polling function as it isn't used anywhere else)

test/e2e/e2eutil.go
chatton: replaced with a single attempt, will keep track of any errors from this

PR #28, opened 2020-03-23
CLOUDP-59487: E2E Test to Scale a Replica Set

chatton:
### All Submissions:

This PR adds a new E2E test which scales a replica set up and down and ensures basic connectivity throughout. The readiness probe remains in the agent image.

This test comes from #26 but without the initContainer change


test/e2e/e2eutil.go
```diff
@@ -28,8 +29,15 @@ func RegisterTypesWithFramework(newTypes ...runtime.Object) error {
 	return nil
 }
 
-func CreateRuntimeObject(obj runtime.Object, ctx *f.TestCtx) error {
-	return f.Global.Client.Create(context.TODO(), obj, &f.CleanupOptions{TestContext: ctx})
+func CreateOrUpdateMongoDB(mdb *mdbv1.MongoDB, ctx *f.TestCtx) error {
```
rodrigovalin: I think we discussed this at some point.

The proposed process is:

a) creation
1. Check if the resource exists
2. If is not found create

b) update
1. Check if the process exists
2. If found, Update

There's always 2 requests in this case. What if we do:

1. Create
2. If error with conflict, Update instead

With this strategy, create is always 1 request, while Update is always 2.

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -68,32 +71,81 @@ func DeletePod(mdb mdbv1.MongoDB, podNum int) func(*testing.T) {
...
-			res, err = collection.InsertOne(ctx, bson.M{"name": "pi", "value": 3.14159})
-			if err != nil {
-				t.Logf("error inserting document: %+v", err)
-				return false, err
-			}
-			return true, nil
-		})
+// IsReachable performs a check by initializing a mongo client
+// and inserting a document into the MongoDB resource
+func IsReachable(mdb *mdbv1.MongoDB) (bool, error) {
```
rodrigovalin: This function has 2 return parameters that are dependant and I think one of them should be removed.

Under no circumstances the following can be observed, `true, err != nil` and that indicates me that we might get all we need by just returning `bool`.

So basically the truth value of `bool` is an alias (almost) of the value of error. And it is actually masking it on line #99, making an actual error be returned as a `false` value.

Can this be made into a 1-parameter returning function instead?

test/e2e/e2eutil.go
chatton: @rodrigovalin the issue with this approach is we run into "resource version does not match" during update, as we first need to GET the resource, which populates the resourceVersion field. (empty string is invalid for the resource)

test/e2e/mongodbtests/mongodbtests.go
chatton: yes you're right! We don't need to have the multi return here at all, nice catch 👍

chatton: ah just seeing the one use case, the error provides the information about why it failed which is logged, in this case what do you think about just returning the error. Maybe with a different name. "Connect" or something similar which returns just an error

rodrigovalin: I think it makes sense if you want to surface the error to the caller

test/e2e/e2eutil.go
rodrigovalin: Ok, then you need to consider that case because from what I see, after calling `Get` you are not using that value (`&mdbv1.MongoDB{}`) that is forgotten right after the function call. This is, it is still the "original" unmodified `mdb` object the one you are passing to the API.

chatton: I addressed this issue in PR: #27

I didn't realise it was a problem until I hit it there.

PR #32, opened 2020-04-14
CLOUDP-60686: Create HaveEqualSpec function and tests

chatton:
### All Submissions:

This PR adds a function which will compare only the changed fields on StatefulSet that we are building. This allows us to construct a stateful set struct that does not include all of the fields that are populated by the api server when doing a comparison.

### DIFF
--- a/pkg/kube/statefulset/statefulset.go
+++ b/pkg/kube/statefulset/statefulset.go
@@ -82,9 +82,8 @@ func IsReady(sts appsv1.StatefulSet) bool {
 	return allUpdated && allReady
 }
 
-// HaveEqualSpec accepts a StatefulSet built, and the StatefulSet that existed already
-// in the API server. HaveEqualSpec returns true if all the changes made in the built StatefulSet
-// are different in any way from the StatefulSet that exists in the API server
+// HaveEqualSpec accepts a StatefulSet builtSts, and a second existingSts, and compares
+// the Spec of both inputs but only comparing the fields that were specified in builtSts
 func HaveEqualSpec(builtSts appsv1.StatefulSet, existingSts appsv1.StatefulSet) (bool, error) {
 	stsToMerge := *existingSts.DeepCopyObject().(*appsv1.StatefulSet)
 	if err := mergo.Merge(&stsToMerge, builtSts, mergo.WithOverride); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/kube/statefulset/statefulset.go
```diff
@@ -78,3 +81,14 @@ func IsReady(sts appsv1.StatefulSet) bool {
 	allReady := replicas == sts.Status.ReadyReplicas
 	return allUpdated && allReady
 }
+
+// HaveEqualSpec accepts a StatefulSet built, and the StatefulSet that existed already
+// in the API server. HaveEqualSpec returns true if all the changes made in the built StatefulSet
```
rodrigovalin, line 86: can you check this function's doc? the function does not take into account Sts existing in the api server or being built, only cares about 2 `appsv1.StatefulSet` objects that are being compared. It could read something like: "checks if both arguments have identical `Spec`s."

You could also mention that the compare function will mask existingSts to only consider the attributes in `builtSts` and not every attribute posible. (am I right on this?)

pkg/kube/statefulset/statefulset_test.go
```diff
@@ -203,3 +204,73 @@ func TestCreateVolumeMountWithMultipleOptions(t *testing.T) {
...
+		areEqual, err := HaveEqualSpec(builtSts, existingSts)
+		assert.NoError(t, err)
+		assert.True(t, areEqual, "When both stateful sets are identical, these should be considered equivalent")
+	})
+	t.Run("Built StatefulSet is different from existing StatefulSet", func(t *testing.T) {
+		builtSts, _ := defaultStatefulSetBuilder().SetUpdateStrategy(appsv1.OnDeleteStatefulSetStrategyType).Build()
+		existingSts, _ := defaultStatefulSetBuilder().Build()
+		areEqual, err := HaveEqualSpec(builtSts, existingSts)
+		assert.NoError(t, err)
+		assert.False(t, areEqual, "We have specified a field that is different from the existring StatefulSet, so these should be considered different")
```
rodrigovalin, line 221:
```suggestion
		assert.False(t, areEqual, "We have specified a field that is different from the existing StatefulSet, so these should be considered different")
```

PR #42, opened 2020-05-11
Cloudp 59610 pre stop hook

rodrigovalin:
### All Submissions:


pkg/controller/mongodb/replicaset_controller_test.go
```diff
@@ -136,29 +136,30 @@ func TestChangingVersion_ResultsInRollingUpdateStrategyType(t *testing.T) {
...
+//	})
+//	t.Run("On Version Change", func(t *testing.T) {
+//		mdb := newTestReplicaSet()
+//		mdb.Spec.Version = "4.0.0"
+//		mdb.Annotations[mdbv1.ReachedVersionAnnotationKey] = "4.2.0"
+//		sts, err := buildStatefulSet(mdb)
+//		assert.NoError(t, err)
+//		assert.Equal(t, appsv1.OnDeleteStatefulSetStrategyType, sts.Spec.UpdateStrategy.Type)
+//	})
+//}
```
chatton, line 165: Have you tested these since? These might be obsolete tests now, but we should still test for the change in strategy type with some unit tests I think.

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -175,12 +194,13 @@ func (r *ReplicaSetReconciler) resetStatefulSetUpdateStrategy(mdb mdbv1.MongoDB)
 
 // isStatefulSetReady checks to see if the stateful set corresponding to the given MongoDB resource
 // is currently in the ready state
-func (r *ReplicaSetReconciler) isStatefulSetReady(mdb mdbv1.MongoDB) (bool, error) {
+func (r *ReplicaSetReconciler) isStatefulSetReady(mdb mdbv1.MongoDB, expectedUpdateStrategy appsv1.StatefulSetUpdateStrategyType) (bool, error) {
+	time.Sleep(time.Second * 5)
```
chatton, line 198: we should add a comment and TODO here, maybe even an issue so this doesn't stay as a long term solution.

cmd/prestop/main.go
```diff
@@ -0,0 +1,231 @@
...
+
+	prettyPrint(health)
+	shouldDelete, err := shouldDeletePod(health)
+	fmt.Printf("shouldDeletePod=%t\n", shouldDelete)
+	logger.Debugf("shouldDeletePod=%t", shouldDelete)
+	if err != nil {
+		logger.Errorf("Error in shouldDeletePod: %s", err)
+	}
+
+	if shouldDelete {
```
chatton, line 181: to keep the level of nesting to a minimum, maybe we should do

```golang
if !shouldDelete {
   return
}

// delete pod
```

cmd/prestop/main.go
```diff
@@ -0,0 +1,231 @@
...
+		zap.S().Errorf("Error building logger config: %s", err)
+		os.Exit(1)
+	}
+	logger = log.Sugar()
+	health, err := getAgentHealthStatus()
+	if err != nil {
+		logger.Errorf("Error getting the agent health file: %s", err)
+	}
+
+	prettyPrint(health)
```
chatton, line 173: this was a temporary debugging print, although getting the full health status file at the time this hook was called could still be useful.

cmd/prestop/main.go
```diff
@@ -0,0 +1,231 @@
...
+	shouldDelete, err := shouldDeletePod(health)
+	fmt.Printf("shouldDeletePod=%t\n", shouldDelete)
```
chatton, line 175: again `fmt` used as pod gets killed/restarted before the log file is ever written

cmd/prestop/main.go
```diff
@@ -0,0 +1,231 @@
...
+	logger.Debugf("shouldDeletePod=%t", shouldDelete)
+	if err != nil {
+		logger.Errorf("Error in shouldDeletePod: %s", err)
```
chatton, line 178: should we continue in this case? This will likely be that the file didn't exist.

cmd/prestop/main.go
```diff
@@ -0,0 +1,231 @@
...
+	for _, m := range lastPlan.Moves {
+
+		// The next conditions are based on observations on the outcome
+		// of the agent after they have stopped the mongo server.
+
+		if m.Move == "WaitFeatureCompatibilityVersionCorrect" {
+			// First condition observed. This is the Plan reported by the
+			// agent on the first MongoD stopped.
+			for _, s := range m.Steps {
+				if s.Step == "WaitFeatureCompatibilityVersionCorrect" &&
```
chatton, line 217: I've seen this stage coming up, why should we delete the pod when seeing `WaitFeatureCompatibilityVersionCorrect`. Does this require a restart?

Also maybe instead of 2 ifs, we could switch on `m.Move`

deploy/operator.yaml
```diff
@@ -32,3 +32,5 @@ spec:
               value: "mongodb-kubernetes-operator"
             - name: AGENT_IMAGE # The MongoDB Agent the operator will deploy to manage MongoDB deployments
               value: quay.io/chatton/mongodb-agent
+            - name: PRE_STOP_HOOK_IMAGE
+              value: quay.io/mongodb/community-operator-pre-stop-hook
```
chatton, line 36: this image still needs to be built in evg, so far it was built manually.

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -117,6 +141,18 @@ func Scale(mdb *mdbv1.MongoDB, newMembers int) func(*testing.T) {
 	}
 }
 
+func Upgrade(mdb *mdbv1.MongoDB, newVersion string) func(*testing.T) {
```
chatton, line 144: Maybe, `ChangeVersion`, this doesn't have to be an upgrade

test/e2e/replica_set_upgrade/replica_set_test.go
```diff
@@ -0,0 +1,44 @@
...
+		mdbv1.MongoDBStatus{
+			MongoURI: mdb.MongoURI(),
+			Phase:    mdbv1.Running,
+		}))
+	t.Run("MongoDB is reachable", mongodbtests.IsReachableDuring(&mdb, time.Second*10,
+		func() {
+			t.Run("Test Version can be upgraded", mongodbtests.Upgrade(&mdb, "4.0.8"))
+			t.Run("Stateful Set Reaches Ready State, after Upgrading", mongodbtests.StatefulSetIsUpdated(&mdb))
+		},
+	))
```
chatton, line 42: should we add a downgrade straight after and ensure connectivity throughout?

We could also check that at some point the MongoDB resource has been configured to`OnDelete` update strategy.

scripts/dev/dockerutil.py
```diff
@@ -20,8 +21,27 @@ def push_image(tag: str):
...
+    to_skip = ("Preparing", "Waiting", "Layer already exists")
+    if "status" in line:
+        if line["status"] in to_skip:
+            return ""
+        if line["status"] == "Pushing":
+            current = int(line["progressDetail"]["current"])
+            total = int(line["progressDetail"]["total"])
+            return "Complete: {:.2f}\n".format(current / total)
+
+    return ""
```
chatton, line 44: thanks for cleaning up the output 👍

pkg/controller/mongodb/mongodb_controller.go
rodrigovalin: commented on the code

https://jira.mongodb.org/browse/CLOUDP-62444

cmd/prestop/main.go
rodrigovalin: we should continue, now, should we exit with a status code of 1 or 0? It means that the container will be restarted after a crash, which might not necessary be the case.

cmd/prestop/main.go
rodrigovalin: this is only seen when the first mongod is stopped, my theory is that the the agent already moved to the next plan, which is "start" the mongodb now that it has been stopped.

After observing the agent's output and plan, this was consistent. Evey time the mongod was restarted, then the first one to be upgraded had this plan, while the following ones, had the other case in question: "ChangeVersion"

cmd/prestop/main.go
chatton: maybe for now we can leave it to not exit 1

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -123,47 +135,54 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	r.log.Debug("Building service")
```
chatton, line 138: maybe we should add a little more context to these debug logs, "Building service: namespace/name"

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -123,47 +135,54 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
 	svc := buildService(mdb)
 	if err = r.client.CreateOrUpdate(&svc); err != nil {
 		r.log.Infof("The service already exists... moving forward: %s", err)
 	}
 
+	r.log.Debug("Creating/Updating StatefulSet")
```
chatton, line 144: same comment as above

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -123,47 +135,54 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
 	if err := r.createOrUpdateStatefulSet(mdb); err != nil {
 		r.log.Infof("Error creating/updating StatefulSet: %+v", err)
 		return reconcile.Result{}, err
 	}
 
-	if ready, err := r.isStatefulSetReady(mdb); err != nil {
+	r.log.Debug("Ensuring StatefulSet is ready")
+	r.log.Debugf("StrategyType: %s", getUpdateStrategyType(mdb))
```
chatton, line 151: we could maybe condense these into a single `Debugw`

pkg/controller/mongodb/mongodb_controller.go
rodrigovalin: this is indicated by the zap context `r.log = zap.S().With("ReplicaSet", request.NamespacedName)` at the beginning of the function.

rodrigovalin:
I have added the downgrade part to the test

PR #44, opened 2020-05-13
Fix service update failure

rodrigovalin:
### All Submissions:

### DIFF
--- a/pkg/controller/mongodb/mongodb_controller.go
+++ b/pkg/controller/mongodb/mongodb_controller.go
@@ -136,9 +136,9 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 	}
 
 	r.log.Debug("Creating/Updating Service")
-	svc := buildService(mdb)
-	if err = r.client.CreateOrUpdate(&svc); err != nil {
+	if err := r.createOrUpdateService(mdb); err != nil {
 		r.log.Infof("Error creating/updating Service: %+v", err)
+		return reconcile.Result{}, err
 	}
 
 	r.log.Debug("Creating/Updating StatefulSet")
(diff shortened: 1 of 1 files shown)


pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -135,7 +135,7 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	r.log.Debug("Building service")
+	r.log.Debug("Building or updating service")
```
chatton, line 138: maybe "Creating/Updating Service"

pkg/kube/client/client.go
```diff
@@ -37,7 +37,7 @@ func (c client) CreateOrUpdate(obj runtime.Object) error {
 		}
 		return err
 	}
-	return c.Update(context.TODO(), obj)
+	return c.Update(context.TODO(), objCopy)
```
chatton, line 40: 👍

rodrigovalin: This solution does not work of course. It always updates the object that was just retrieved by the API!

rodrigovalin:
I'm leaving this solution for later, the message observed only happens with some type of objects (Services for instance). There's a reference to this issue [here](https://github.com/kubernetes/kubernetes/issues/70674).

I'll come back to this later.

chatton:
@rodrigovalin did you test to see the behaviour if we fetch the latest service before attempting the update? I think this should do the trick for this particular issue.

rodrigovalin:
@chatton yes, the fetched service can be updated. The updated service can't be used for updates because it is missing the resourceVersion.

What needs to be done is to obtain the current `resourceVersion` from the API, and pass it along the new updated Service (https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#metadata). The client is not responsible for updating this resourceVersion on the client side.

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -220,6 +220,21 @@ func (r *ReplicaSetReconciler) createOrUpdateStatefulSet(mdb mdbv1.MongoDB) erro
...
 
+func (r *ReplicaSetReconciler) createOrUpdateService(mdb mdbv1.MongoDB) error {
+	svc := buildService(mdb)
+	currentSvc := corev1.Service{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}, &currentSvc); err == nil {
+		// For the particular case of the Service we need to provide the resourceVersion when updating the object.
+		// From the docs (https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#metadata)
+		// > "This value MUST be treated as opaque by clients and passed unmodified back to the server"
+		svc.ResourceVersion = currentSvc.ResourceVersion
+	} else if !errors.IsNotFound(err) {
```
LouisPlisso, line 231: wouldn't it be better if we follow usual error handling pattern: if err!= nil return err

and in the rest of the code override the ResourceVersion unconditionally

rodrigovalin: resourceVersion can't be update unconditionally, it needs to be fetched from an existing service. If the resource does not exist (this is a service that will be just created) then resourceVersion is not updated.

chatton: At the moment, what's the use case for updating the service? Currently, once the service is created, it never needs to be be updated. This might change in the future, but can we simply do the creation, and never try and update.

LouisPlisso: this code reminds me the one where we had some error not handled properly with imbricated elseif

what i undrestand is that if err is nil we override ResourceVersion and if it is not nil we return the error

please make the behaviour more explicit (I think starting by sending the error when it's not nil is the best start and it's the most common pattern)

PR #49, opened 2020-05-27
Cloudp 62906 set owner reference

bznein:
### All Submissions:

* [] Have you opened an Issue before filing this PR?
* [x] Have you signed our [CLA](r)?

### DIFF
--- a/pkg/controller/mongodb/mongodb_controller.go
+++ b/pkg/controller/mongodb/mongodb_controller.go
@@ -449,7 +449,7 @@ func buildStatefulSet(mdb mdbv1.MongoDB) (appsv1.StatefulSet, error) {
 		},
 	}
 
-	ownerReference := []metav1.OwnerReference{
+	ownerReferences := []metav1.OwnerReference{
 		*metav1.NewControllerRef(&mdb, schema.GroupVersionKind{
 			Group:   mdbv1.SchemeGroupVersion.Group,
 			Version: mdbv1.SchemeGroupVersion.Version,
@@ -466,7 +466,7 @@ func buildStatefulSet(mdb mdbv1.MongoDB) (appsv1.StatefulSet, error) {
 		SetMatchLabels(labels).
 		SetServiceName(mdb.ServiceName()).
 		SetUpdateStrategy(getUpdateStrategyType(mdb)).
-		SetOwnerReference(ownerReference)
+		SetOwnerReference(ownerReferences)
 
 	// TODO: Add this section to architecture document.
 	// The design of the multi-container and the different volumes mounted to them is as follows:
--- a/pkg/controller/mongodb/mongodb_controller.go
+++ b/pkg/controller/mongodb/mongodb_controller.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -149,6 +150,7 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 
 	r.log.Debugf("Ensuring StatefulSet is ready, with type: %s", getUpdateStrategyType(mdb))
 	if ready, err := r.isStatefulSetReady(mdb, getUpdateStrategyType(mdb)); err != nil {
+
 		r.log.Infof("error checking StatefulSet status: %+v", err)
 		return reconcile.Result{}, err
 	} else if !ready {
@@ -447,6 +449,14 @@ func buildStatefulSet(mdb mdbv1.MongoDB) (appsv1.StatefulSet, error) {
 		},
 	}
 
+	ownerReference := []metav1.OwnerReference{
+		*metav1.NewControllerRef(&mdb, schema.GroupVersionKind{
+			Group:   mdbv1.SchemeGroupVersion.Group,
+			Version: mdbv1.SchemeGroupVersion.Version,
+			Kind:    mdb.Kind,
+		}),
+	}
+
 	builder := statefulset.NewBuilder().
 		SetPodTemplateSpec(podSpecTemplate).
 		SetNamespace(mdb.Namespace).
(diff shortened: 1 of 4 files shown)


test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -37,6 +38,35 @@ func StatefulSetIsReady(mdb *mdbv1.MongoDB) func(t *testing.T) {
...
+				Group:   mdbv1.SchemeGroupVersion.Group,
+				Version: mdbv1.SchemeGroupVersion.Version,
+				Kind:    mdb.Kind,
+			})
+
+		for _, ownerReference := range sts.GetOwnerReferences() {
+			if ownerReference.APIVersion == ownerReferenceMdb.APIVersion &&
+				ownerReference.Kind == "MongoDB" &&
+				ownerReference.Name == ownerReferenceMdb.Name &&
+				ownerReference.UID == ownerReferenceMdb.UID {
```
chatton, line 60: Rather than checking for all of the conditions in a single if statement, I think it would be more beneficial to do an individual assertion on each field. In this case, we know either all of the fields are correct, or at least one of them isn't, we won't know which one is wrong.

We could do something like 

```golang
assert.Equal(t, ownerReferenceMdb.APIVersion, ownerReference.APIVersion)
assert.Equal(t, "MongoDB", ownerReference. Kind)
...
```

The assert library will give us a more detailed breakdown of what went wrong, what do you think?

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -447,6 +449,14 @@ func buildStatefulSet(mdb mdbv1.MongoDB) (appsv1.StatefulSet, error) {
 		},
 	}
 
+	ownerReference := []metav1.OwnerReference{
```
chatton, line 452: [nit] as it's a slice, we should use the plural form for naming `ownerReferences`

test/e2e/replica_set/replica_set_test.go
```diff
@@ -24,6 +24,7 @@ func TestReplicaSet(t *testing.T) {
 	t.Run("Create MongoDB Resource", mongodbtests.CreateMongoDBResource(&mdb, ctx))
 	t.Run("Config Map Was Correctly Created", mongodbtests.AutomationConfigConfigMapExists(&mdb))
 	t.Run("Stateful Set Reaches Ready State", mongodbtests.StatefulSetIsReady(&mdb))
+	t.Run("Stateful Set has OwnerReference", mongodbtests.StatefulSetHasOwnerReference(&mdb))
```
chatton, line 27: I believe we have a "BasicConnectivity" or "BasicFunctionality" test that incorporates a lot of the common things we test frequently. This new test would be a great one to add there!

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -37,6 +38,35 @@ func StatefulSetIsReady(mdb *mdbv1.MongoDB) func(t *testing.T) {
 	}
 }
 
+func StatefulSetHasOwnerReference(mdb *mdbv1.MongoDB) func(t *testing.T) {
+	return func(t *testing.T) {
+		sts := appsv1.StatefulSet{}
+		err := f.Global.Client.Get(context.TODO(), types.NamespacedName{Name: mdb.Name, Namespace: f.Global.OperatorNamespace}, &sts)
+		if err != nil {
+			t.Fatal(fmt.Errorf("%s", err))
```
chatton, line 46: in this case there's no real difference between `fmt.Errorf("%s", err)` and `err`

we should just be able to do `t.Fatal(err)`

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -37,6 +38,35 @@ func StatefulSetIsReady(mdb *mdbv1.MongoDB) func(t *testing.T) {
...
+			if ownerReference.APIVersion == ownerReferenceMdb.APIVersion &&
+				ownerReference.Kind == "MongoDB" &&
+				ownerReference.Name == ownerReferenceMdb.Name &&
+				ownerReference.UID == ownerReferenceMdb.UID {
+				t.Logf("StatefulSet %s/%s has the correct OwnerReference!", mdb.Namespace, mdb.Name)
+				return
+			}
+		}
+
+		t.Fatal(fmt.Errorf("StatefulSet has not a correct OwnerReference"))
```
chatton, line 66: bear in mind that this will be the output we get when trying to investigate what went wrong, we should try and include the actual and expected values to get better insight.

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -37,6 +38,35 @@ func StatefulSetIsReady(mdb *mdbv1.MongoDB) func(t *testing.T) {
...
 
+func StatefulSetHasOwnerReference(mdb *mdbv1.MongoDB) func(t *testing.T) {
+	return func(t *testing.T) {
+		sts := appsv1.StatefulSet{}
+		err := f.Global.Client.Get(context.TODO(), types.NamespacedName{Name: mdb.Name, Namespace: f.Global.OperatorNamespace}, &sts)
+		if err != nil {
+			t.Fatal(fmt.Errorf("%s", err))
+		}
+
+		ownerReferenceMdb :=
```
chatton, line 49: one thing that to consider is that we have no way currently of testing for a different owner reference (we do only have one so it's not that big a deal) but a possible option here is to add it as a function argument so in the test we can decide which owner reference to look for e.g.

```golang
StatefulSetHasOwnerReference(mdb *mdbv1.MongoDB, expectedOwnerReference metav1.OwnerReference)
```

will leave this one up to you!

test/e2e/mongodbtests/mongodbtests.go
bznein: That makes absolutely sense, I'll do it!

bznein: Now that I think of it, since `getOwnerReferences()` return a slice and we are checking if at least one element of it its equal to `ownerReferenceMdb`, that won't work, or am I missing something? We can't assert it on one particular element.

test/e2e/mongodbtests/mongodbtests.go
bznein: Related to the comment above: since `getOwnerReferences()` returns a slice, we would have to include every ownerReference returned? In this case we only have one but we might potentially have more

test/e2e/mongodbtests/mongodbtests.go
chatton: well in our case we know there should be exactly 1 element, so either we could pass in a slice of owner references, or assert a single element and compare index 0

PR #52, opened 2020-06-03
CLOUDP-62351: update service

### LINKED ISSUE #41: replica set could not be created successfully following readme
Hi 
Thanks for your sharing of mongodb operator. I'm learning how to build k8s operator by mongodb these days. I tried to create mongodb operator following readme insturctions. But it failed for :

[failure descripiton]
There is an error in operator while checking satus of replica set. When Reconcile is called again, and an error was returned by `r.client.CreateOrUpdate(&svc)`
` The service already exists... moving forward: Service "example-mongodb-svc" is invalid: metadata.resourceVersion: Invalid value: "": must be specified for an update`

image

I tried to fix it by [Kubernetes Service invalid clusterIP or resourceVersion](https://www.timcosta.io/kubernetes-service-invalid-clusterip-or-resourceversion/). But it didn't work.

my k8s version:
`Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.5", GitCommit:"20c265fef0741dd71a66480e35bd69f18351daea", GitTreeState:"clean", BuildDate:"2019-10-15T19:16:51Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:48:36Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
`
Any response/help will be much appreciated.

bznein:
closes #41 

### All Submissions:

* [] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
* [x] Put `closes #xxxx` in your comment to auto-close the issue that your PR fixes (if such).

### DIFF
--- a/pkg/controller/mongodb/mongodb_controller.go
+++ b/pkg/controller/mongodb/mongodb_controller.go
@@ -136,7 +136,7 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	r.log.Debug("Ensuring the service esists")
+	r.log.Debug("Ensuring the service exists")
 	if err := r.ensureService(mdb); err != nil {
 		r.log.Infof("Error ensuring the service exists: %s", err)
 		return reconcile.Result{}, err
--- a/pkg/controller/mongodb/mongodb_controller.go
+++ b/pkg/controller/mongodb/mongodb_controller.go
@@ -137,8 +137,7 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 	}
 
 	r.log.Debug("Building service")
-	svc := buildService(mdb)
-	if err = r.client.CreateOrUpdate(&svc); err != nil {
+	if err = r.createOrUpdateService(mdb); err != nil {
 		r.log.Infof("The service already exists... moving forward: %s", err)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -205,9 +208,20 @@ func (r *ReplicaSetReconciler) isStatefulSetReady(mdb mdbv1.MongoDB, expectedUpd
 	return statefulset.IsReady(set, mdb.Spec.Members) && expectedUpdateStrategy == set.Spec.UpdateStrategy.Type, nil
 }
 
+func (r *ReplicaSetReconciler) createService(mdb mdbv1.MongoDB) error {
+	svc := buildService(mdb)
+	svcCopy := svc.DeepCopyObject()
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: mdb.Name, Namespace: mdb.Namespace}, svcCopy)
```
chatton, line 214: watch out! the name of the service is not the same as the name of the MongoDB resource, it is `<mdb-name>-svc`, there is a function `mdb.ServiceName()` we can use here.

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -205,9 +208,20 @@ func (r *ReplicaSetReconciler) isStatefulSetReady(mdb mdbv1.MongoDB, expectedUpd
 	return statefulset.IsReady(set, mdb.Spec.Members) && expectedUpdateStrategy == set.Spec.UpdateStrategy.Type, nil
 }
 
+func (r *ReplicaSetReconciler) createService(mdb mdbv1.MongoDB) error {
```
chatton, line 211: I think maybe we're making this a little more complicated that it needs to be, if the goal is to create the service but not worry about errors during creation if the service exists already, we're guarding from those errors by checking for its existence before creating. I think in this case we can replace this function with a call to 

`r.client.Get(context.TODO(), types.NamespacedName{Name: mdb.ServiceName(), Namespace: mdb.Namespace}), svc)`

this will either fail with a conflict error (which we don't consider an error right now), or a different kind of error which we do want to worry about

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -136,10 +136,13 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
-	r.log.Debug("Building service")
-	svc := buildService(mdb)
-	if err = r.client.CreateOrUpdate(&svc); err != nil {
-		r.log.Infof("The service already exists... moving forward: %s", err)
+	r.log.Debug("Building and creating service")
+	if err = r.createService(mdb); err != nil {
+		if errors.IsAlreadyExists(err) {
+			r.log.Infof("The service already exists... moving forward: %s", err)
+		} else {
+			return reconcile.Result{}, err
```
chatton, line 144: One thing to keep in mind here, returning `reconcile.Result{}, err` will requeue a new reconcilliation, but we aren't actually getting any info about this error. We should make sure to log it before hand.

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -136,10 +136,13 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	r.log.Debug("Building service")
-	svc := buildService(mdb)
-	if err = r.client.CreateOrUpdate(&svc); err != nil {
-		r.log.Infof("The service already exists... moving forward: %s", err)
+	r.log.Debug("Building and creating service")
+	if err = r.createService(mdb); err != nil {
```
chatton, line 140: as mentioned below, we may not need this `createService` function at all, it should be safe to use `r.client.Create` directly

pkg/controller/mongodb/mongodb_controller.go
bznein: Ouch! Thanks for spotting this!

PR #55, opened 2020-06-05
CLOUDP-64640: fix deploy operator

bznein:
### All Submissions:

### DIFF
--- a/scripts/dev/build_and_deploy_operator.py
+++ b/scripts/dev/build_and_deploy_operator.py
@@ -39,7 +39,7 @@ def load_yaml_from_file(path: str) -> Optional[Dict]:
         return yaml.full_load(f.read())
     return None
 
-def _ensure_crds() -> bool:
+def _ensure_crds():
     """
     ensure_crds makes sure that all the required CRDs have been created
     """
@@ -53,9 +53,8 @@ def _ensure_crds() -> bool:
     # Make sure that the CRD has being deleted before trying to create it again
     if not wait_for_condition(
             lambda: crdv1.list_custom_resource_definition(field_selector = "metadata.name==mongodb.mongodb.com"),
-            lambda crd_list : len(crd_list.items)==0, timeout=5, sleep_time=0.1):
-        print("Execution timed out while waiting for the CRD to be deleted")
-        return False
+            lambda crd_list : len(crd_list.items)==0, timeout=5, sleep_time=0.5):
+        raise Exception("Execution timed out while waiting for the CRD to be deleted")
 
 
     # TODO: fix this, when calling create_custom_resource_definition, we get the error
@@ -67,7 +66,6 @@ def _ensure_crds() -> bool:
         pass
 
     print("Ensured CRDs")
-    return True
 
 
 def build_and_push_operator(repo_url: str, tag: str, path: str):
@@ -88,9 +86,8 @@ def deploy_operator():
     rbacv1 = client.RbacAuthorizationV1Api()
 
     dev_config = load_config()
-    if not _ensure_crds():
-        return
-
+    _ensure_crds()
+    
     ignore_if_already_exists(
         lambda: rbacv1.create_namespaced_role(
             dev_config.namespace, _load_operator_role()
--- a/scripts/dev/build_and_deploy_operator.py
+++ b/scripts/dev/build_and_deploy_operator.py
@@ -9,7 +9,7 @@ from dev_config import DevConfig, load_config
 from dockerfile_generator import render
 from dockerutil import build_and_push_image
 
-from k8sutil import wait_for_k8s_api_condition,ignore_if_already_exists, ignore_if_doesnt_exist
+from k8sutil import wait_for_condition,ignore_if_already_exists, ignore_if_doesnt_exist
 
 
 def _load_operator_service_account() -> Optional[Dict]:
@@ -51,7 +51,7 @@ def _ensure_crds() -> bool:
     )
 
     # Make sure that the CRD has being deleted before trying to create it again
-    if not wait_for_k8s_api_condition(
+    if not wait_for_condition(
             lambda: crdv1.list_custom_resource_definition(field_selector = "metadata.name==mongodb.mongodb.com"),
             lambda crd_list : len(crd_list.items)==0, timeout=5, sleep_time=0.1):
         print("Execution timed out while waiting for the CRD to be deleted")
--- a/scripts/dev/build_and_deploy_operator.py
+++ b/scripts/dev/build_and_deploy_operator.py
@@ -1,40 +1,16 @@
 import io
 import os
-import time
 from typing import Dict, Optional
 
 import yaml
 from kubernetes import client, config
-from kubernetes.client.rest import ApiException
 
 from dev_config import DevConfig, load_config
 from dockerfile_generator import render
 from dockerutil import build_and_push_image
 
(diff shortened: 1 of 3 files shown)


scripts/dev/build_and_deploy_operator.py
```diff
@@ -1,16 +1,16 @@
...
 
 import yaml
 from kubernetes import client, config
-from kubernetes.client.rest import ApiException
 
 from dev_config import DevConfig, load_config
 from dockerfile_generator import render
 from dockerutil import build_and_push_image
 
+from k8sutil import wait_for_condition,ignore_if_already_exists, ignore_if_doesnt_exist
```
chatton, line 12: Is this formatted properly? We need to set up a pre-commit hook for this. Generally we use [black](https://github.com/psf/black) to format our python

scripts/dev/e2e.py
```diff
@@ -4,10 +4,13 @@
 from build_and_deploy_operator import (
     build_and_push_operator,
     deploy_operator,
-    ignore_if_doesnt_exist,
-    ignore_if_already_exists,
     load_yaml_from_file,
 )  # TODO: put these function somewhere else
```
chatton, line 8: looks like we can get rid of this TODO! 🎉

scripts/dev/e2e.py
```diff
@@ -222,7 +218,13 @@ def main():
...
     pod = create_test_runner_pod(args.test)
     corev1 = client.CoreV1Api()
-    wait_for_pod_to_be_running(corev1, TEST_RUNNER_NAME, dev_config.namespace)
+
+    print("Waiting for pod to be running")
+    if not wait_for_condition(
+            lambda: corev1.read_namespaced_pod(TEST_RUNNER_NAME, dev_config.namespace),
+            lambda pod : pod.status.phase=="Running",
+            sleep_time=5, timeout=50, exceptions_to_ignore=ApiException):
+        raise Exception("Pod never got into Running state!")
```
chatton, line 227: what do you think of keeping the function `wait_for_pod_to_be_running` but just change the implementation to use `wait_for_condition`. I think we get better readability this way instead of needing to parse the different lambdas.

scripts/dev/k8sutil.py
```diff
@@ -0,0 +1,63 @@
...
+
+# time to sleep between retries
+SLEEP_TIME = 2
+# no timeout (loop forever)
+INFINITY = -1
+
+def _current_milliseconds():
+    return int(round(time.time() * 1000))
+
+def wait_for_condition(fn, condition, exceptions_to_ignore=[], codes_to_ignore=[],sleep_time=SLEEP_TIME, timeout=INFINITY) -> bool:
```
chatton, line 13: watch out for using mutable objects as default arguments! It's counter intuitive, but it will use the same instance each time.

```python

def add(to_add, my_list=[]):
    my_list.append(to_add)
    return my_list


>>> print(add(1))
[1]
>>> print(add(2))
[1, 2]
>>> print(add(3))
[1, 2, 3]
>>> print(add(4))
[1, 2, 3, 4]
>>> print(add(5))
[1, 2, 3, 4, 5]
```

for collections and other mutable objects, we should do the following

```python

def add(to_add, my_list=None):
    if my_list is None:
       my_list = []
    my_list.append(to_add)
    return my_list
```

This ensures a new instance scoped to the function each time

```python
>>> print(add(1))
[1]
>>> print(add(2))
[2]
>>> print(add(3))
[3]
>>> print(add(4))
[4]
>>> print(add(5))
[5]
```

scripts/dev/k8sutil.py
```diff
@@ -0,0 +1,63 @@
+import time
+
+from kubernetes.client.rest import ApiException
+
+# time to sleep between retries
+SLEEP_TIME = 2
+# no timeout (loop forever)
+INFINITY = -1
+
+def _current_milliseconds():
```
chatton, line 10:
```suggestion
def _current_milliseconds() -> int:
```

scripts/dev/k8sutil.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    start_time = _current_milliseconds()
+    end = start_time + (timeout * 1000)
+
+
+    while _current_milliseconds() < end or timeout <= 0:
+        res = None
+        try:
+            res =  _ignore_error_codes(fn,codes_to_ignore)
+        except exceptions_to_ignore:
```
chatton, line 30: [q] does excepting a list of exceptions work? Is it the same as `(ExceptionTypeA, ExceptionTypeB)` ?

scripts/dev/build_and_deploy_operator.py
bznein: Oh I didn't know this, thanks :)

scripts/dev/build_and_deploy_operator.py
```diff
@@ -105,7 +88,8 @@ def deploy_operator():
     rbacv1 = client.RbacAuthorizationV1Api()
 
     dev_config = load_config()
-    _ensure_crds()
+    if not _ensure_crds():
```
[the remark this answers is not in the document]
bznein: Agree, updated!

scripts/dev/e2e.py
bznein: Makes sense and makes it clearer, updated!

scripts/dev/k8sutil.py
bznein: Thanks I was not aware of that!

scripts/dev/k8sutil.py
bznein: It actually works if passing a tuple, not a list. 

E.g. with this code:
```def f(i):
    if i==1:
        raise BufferError("Buffer")
    else:
        raise LookupError("Lookup")

def ignore_exc(ex = None):
    for i in range (0,2):
        try:
            f(i)
        except ex as e:
            print("EXCEPT: " + format(e))
```
the call to ` ignore_exc((BufferError,LookupError))` would work and print 

```
EXCEPT: Lookup
EXCEPT: Buffer
```

But the call to `ignore_exc([BufferError,LookupError])` would raise an exception:

```
TypeError: catching classes that do not inherit from BaseException is not allowed
```

I modified the comment to the function to make it clearer that a tuple should be used

scripts/dev/k8sutil.py
```diff
@@ -0,0 +1,70 @@
...
+    start_time = _current_milliseconds()
+    end = start_time + (timeout * 1000)
+
+    while _current_milliseconds() < end or timeout <= 0:
+        res = None
+        try:
+            res = _ignore_error_codes(fn, codes_to_ignore)
+        except exceptions_to_ignore:
+            pass
+        if res is not None and condition(res) == True:
```
chatton, line 40: [nit]

```suggestion
        if res is not None and condition(res):
```

scripts/dev/e2e.py
```diff
@@ -140,9 +143,35 @@ def create_test_runner_pod(test: str):
     dev_config = load_config()
     corev1 = client.CoreV1Api()
     pod_body = _get_testrunner_pod_body(test)
+
+    if not wait_for_condition(
+        lambda: corev1.list_namespaced_pod(
+            dev_config.namespace, field_selector="metadata.name==" + TEST_RUNNER_NAME
```
chatton, line 149: [nit] we could use string.format or f-strings

`"metadata.name=={}".format(TEST_RUNNER_NAME)`

`f"metadata.name=={TEST_RUNNER_NAME)"`

PR #60, opened 2020-06-10
CLOUDP-64701: ensure automation config version bump

bznein:
### All Submissions:

### DIFF
--- a/pkg/controller/mongodb/mongodb_controller.go
+++ b/pkg/controller/mongodb/mongodb_controller.go
@@ -9,7 +9,6 @@ import (
 	"reflect"
 	"time"
 
-	"github.com/kylelemons/godebug/pretty"
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/controller/predicates"
 
 	mdbv1 "github.com/mongodb/mongodb-kubernetes-operator/pkg/apis/mongodb/v1"
--- a/pkg/controller/mongodb/mongodb_controller.go
+++ b/pkg/controller/mongodb/mongodb_controller.go
@@ -25,7 +25,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
-	k8sClient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -275,20 +274,15 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
 func buildAutomationConfig(mdb mdbv1.MongoDB, mdbVersionConfig automationconfig.MongoDbVersionConfig, client mdbClient.Client) (automationconfig.AutomationConfig, error) {
 	domain := getDomain(mdb.ServiceName(), mdb.Namespace, "")
 
-	currentAc, err := getCurrentAutomationConfig(client, mdb)
-	if err != nil {
-		return automationconfig.AutomationConfig{}, err
-	}
 	newAc, err := automationconfig.NewBuilder().
 		SetTopology(automationconfig.ReplicaSetTopology).
 		SetName(mdb.Name).
 		SetDomain(domain).
 		SetMembers(mdb.Spec.Members).
-		SetPreviousAutomationConfig(currentAc).
 		SetMongoDBVersion(mdb.Spec.Version).
 		SetFCV(mdb.GetFCV()).
 		AddVersion(mdbVersionConfig).
-		Build()
+		Build(client, mdb, AutomationConfigKey)
 
 	if err != nil {
 		return automationconfig.AutomationConfig{}, err
@@ -328,20 +322,6 @@ func buildService(mdb mdbv1.MongoDB) corev1.Service {
 		SetClusterIP("None").
 		SetPort(27017).
 		Build()
-}
-
-func getCurrentAutomationConfig(client mdbClient.Client, mdb mdbv1.MongoDB) (automationconfig.AutomationConfig, error) {
-	currentCm := corev1.ConfigMap{}
-	currentAc := automationconfig.AutomationConfig{}
-	if err := client.Get(context.TODO(), types.NamespacedName{Name: mdb.ConfigMapName(), Namespace: mdb.Namespace}, &currentCm); err != nil {
-		// If the AC was not found we don't treat it as an error
-		return automationconfig.AutomationConfig{}, k8sClient.IgnoreNotFound(err)
-
-	}
-	if err := json.Unmarshal([]byte(currentCm.Data[AutomationConfigKey]), &currentAc); err != nil {
-		return automationconfig.AutomationConfig{}, err
-	}
-	return currentAc, nil
 }
 
 func (r ReplicaSetReconciler) buildAutomationConfigConfigMap(mdb mdbv1.MongoDB) (corev1.ConfigMap, error) {
--- a/pkg/controller/mongodb/mongodb_controller.go
+++ b/pkg/controller/mongodb/mongodb_controller.go
@@ -6,8 +6,10 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"reflect"
 	"time"
 
+	"github.com/kylelemons/godebug/pretty"
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/controller/predicates"
 
 	mdbv1 "github.com/mongodb/mongodb-kubernetes-operator/pkg/apis/mongodb/v1"
@@ -130,7 +132,6 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	// TODO: Read current automation config version from config map
 	if err := r.ensureAutomationConfig(mdb); err != nil {
 		r.log.Infof("error creating automation config config map: %s", err)
 		return reconcile.Result{}, err
@@ -265,24 +266,43 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
 	if err != nil {
 		return err
 	}
+
 	if err := r.client.CreateOrUpdate(&cm); err != nil {
 		return err
 	}
 	return nil
 }
 
-func buildAutomationConfig(mdb mdbv1.MongoDB, mdbVersionConfig automationconfig.MongoDbVersionConfig) automationconfig.AutomationConfig {
+func buildAutomationConfig(mdb mdbv1.MongoDB, mdbVersionConfig automationconfig.MongoDbVersionConfig, currentAc automationconfig.AutomationConfig) automationconfig.AutomationConfig {
 	domain := getDomain(mdb.ServiceName(), mdb.Namespace, "")
-	return automationconfig.NewBuilder().
+	newAc := automationconfig.NewBuilder().
 		SetTopology(automationconfig.ReplicaSetTopology).
 		SetName(mdb.Name).
 		SetDomain(domain).
 		SetMembers(mdb.Spec.Members).
 		SetMongoDBVersion(mdb.Spec.Version).
-		SetAutomationConfigVersion(1). // TODO: Correctly set the version
+		SetAutomationConfigVersion(currentAc.Version).
 		SetFCV(mdb.GetFCV()).
 		AddVersion(mdbVersionConfig).
 		Build()
+
+	zap.L().Sugar().Infof("NewAc: %+v", newAc)
+
+	zap.L().Sugar().Infof("OlcdA: %+v", currentAc)
+
+	b1, _ := json.Marshal(newAc)
+	b2, _ := json.Marshal(currentAc)
+
+	zap.L().Sugar().Infof("Bytes: %s", reflect.DeepEqual([]byte(b1), []byte(b2)))
+	zap.L().Sugar().Infof(pretty.Compare(newAc, currentAc))
+	if !reflect.DeepEqual(newAc, currentAc) {
+		zap.L().Info("Not equal")
+		newAc.Version += 1
+	}
+
+	zap.L().Sugar().Infof("NewAcModified: %s", newAc)
+
+	return newAc
 }
 
 func readVersionManifestFromDisk() (automationconfig.VersionManifest, error) {
@@ -323,7 +343,12 @@ func (r ReplicaSetReconciler) buildAutomationConfigConfigMap(mdb mdbv1.MongoDB) 
 	if err != nil {
 		return corev1.ConfigMap{}, fmt.Errorf("error reading version manifest from disk: %+v", err)
 	}
-	ac := buildAutomationConfig(mdb, manifest.BuildsForVersion(mdb.Spec.Version))
+	currentCm := &corev1.ConfigMap{}
+	currentAc := automationconfig.AutomationConfig{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: mdb.ConfigMapName(), Namespace: mdb.Namespace}, currentCm); err == nil {
+		json.Unmarshal([]byte(currentCm.Data[AutomationConfigKey]), &currentAc)
+	}
+	ac := buildAutomationConfig(mdb, manifest.BuildsForVersion(mdb.Spec.Version), currentAc)
 	acBytes, err := json.Marshal(ac)
 	if err != nil {
 		return corev1.ConfigMap{}, err
--- a/pkg/controller/mongodb/mongodb_controller.go
+++ b/pkg/controller/mongodb/mongodb_controller.go
@@ -1,12 +1,12 @@
 package mongodb
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
-	"reflect"
 	"time"
 
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/controller/predicates"
@@ -272,7 +272,7 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
 	return nil
 }
 
-func buildAutomationConfig(mdb mdbv1.MongoDB, mdbVersionConfig automationconfig.MongoDbVersionConfig, currentAc automationconfig.AutomationConfig) automationconfig.AutomationConfig {
+func buildAutomationConfig(mdb mdbv1.MongoDB, mdbVersionConfig automationconfig.MongoDbVersionConfig, currentAc automationconfig.AutomationConfig) (automationconfig.AutomationConfig, error) {
 	domain := getDomain(mdb.ServiceName(), mdb.Namespace, "")
 	newAc := automationconfig.NewBuilder().
 		SetTopology(automationconfig.ReplicaSetTopology).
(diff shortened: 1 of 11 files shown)


pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -265,24 +265,34 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
...
 		SetMembers(mdb.Spec.Members).
 		SetMongoDBVersion(mdb.Spec.Version).
-		SetAutomationConfigVersion(1). // TODO: Correctly set the version
+		SetAutomationConfigVersion(currentAc.Version).
 		SetFCV(mdb.GetFCV()).
 		AddVersion(mdbVersionConfig).
 		Build()
+
+	b1, _ := json.Marshal(newAc)
+	b2, _ := json.Marshal(currentAc)
```
chatton, line 289: we should never ignore the errors! It looks like we'll need to update the method signature here to also return an error.

We can also name them a little better. Even `newAcBytes` and `existingAcBytes`

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -265,24 +265,34 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
...
+
+	if !reflect.DeepEqual([]byte(b1), []byte(b2)) {
```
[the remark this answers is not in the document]
chatton: Also, the `[]byte(...)` is redundant, both b1 and b2 are of type `[]byte` already!

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -323,7 +333,12 @@ func (r ReplicaSetReconciler) buildAutomationConfigConfigMap(mdb mdbv1.MongoDB)
 	if err != nil {
 		return corev1.ConfigMap{}, fmt.Errorf("error reading version manifest from disk: %+v", err)
 	}
-	ac := buildAutomationConfig(mdb, manifest.BuildsForVersion(mdb.Spec.Version))
+	currentCm := &corev1.ConfigMap{}
+	currentAc := automationconfig.AutomationConfig{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: mdb.ConfigMapName(), Namespace: mdb.Namespace}, currentCm); err == nil {
+		json.Unmarshal([]byte(currentCm.Data[AutomationConfigKey]), &currentAc)
```
chatton, line 339: [!] `json.Umarshal` returns an error, we should be sure to handle it.
```golang
if err := json.Unmarshal(...); err != nil {
    return err
}
```

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -323,7 +333,12 @@ func (r ReplicaSetReconciler) buildAutomationConfigConfigMap(mdb mdbv1.MongoDB)
 	if err != nil {
 		return corev1.ConfigMap{}, fmt.Errorf("error reading version manifest from disk: %+v", err)
 	}
-	ac := buildAutomationConfig(mdb, manifest.BuildsForVersion(mdb.Spec.Version))
+	currentCm := &corev1.ConfigMap{}
+	currentAc := automationconfig.AutomationConfig{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: mdb.ConfigMapName(), Namespace: mdb.Namespace}, currentCm); err == nil {
```
chatton, line 338: A good practice around error handling/indentation in go is to make sure that the happy path of the code is always at the left most side of the function. This basically translates to check for `err != nil` rather than `err == nil`

I think this is a good pattern to follow, as it reduces the scope of where we need to look to see what a function is doing when reading it.

so basically

```golang
if err := GetConfigMap(); err != nil {
    return err
}
// do thing with configmap
```

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -323,7 +333,12 @@ func (r ReplicaSetReconciler) buildAutomationConfigConfigMap(mdb mdbv1.MongoDB)
 	if err != nil {
 		return corev1.ConfigMap{}, fmt.Errorf("error reading version manifest from disk: %+v", err)
 	}
-	ac := buildAutomationConfig(mdb, manifest.BuildsForVersion(mdb.Spec.Version))
+	currentCm := &corev1.ConfigMap{}
+	currentAc := automationconfig.AutomationConfig{}
```
chatton, line 337: [nit] there's a stylistic inconsistency here

at first we get a reference, then pass the reference directly, but then we create the struct, and pass it by reference

I would prefer

```golang
currentCm := corev1.ConfigMap{}
currentAc := automationconfig.AutomationConfig{}

// use &currentCm
// use &currentAc
```

pkg/controller/mongodb/replicaset_controller_test.go
```diff
@@ -181,5 +182,56 @@ func TestService_isCorrectlyCreatedAndUpdated(t *testing.T) {
...
+}
+
+func getCurrentAutomationConfig(mdb mdbv1.MongoDB, mgr *client.MockedManager, t *testing.T) automationconfig.AutomationConfig {
+	currentCm := &corev1.ConfigMap{}
+	currentAc := automationconfig.AutomationConfig{}
+	err := mgr.GetClient().Get(context.TODO(), types.NamespacedName{Name: mdb.ConfigMapName(), Namespace: mdb.Namespace}, currentCm)
+
+	assert.NoError(t, err)
+
+	json.Unmarshal([]byte(currentCm.Data[AutomationConfigKey]), &currentAc)
```
chatton, line 194:
```suggestion
	err = json.Unmarshal([]byte(currentCm.Data[AutomationConfigKey]), &currentAc)
	assert.NoError(t, err)
```

pkg/controller/mongodb/replicaset_controller_test.go
```diff
@@ -181,5 +182,56 @@ func TestService_isCorrectlyCreatedAndUpdated(t *testing.T) {
...
+	mgr := client.NewManager(&mdb)
+	r := newReconciler(mgr, mockManifestProvider(mdb.Spec.Version))
+	res, err := r.Reconcile(reconcile.Request{NamespacedName: types.NamespacedName{Namespace: mdb.Namespace, Name: mdb.Name}})
+	assertReconciliationSuccessful(t, res, err)
+
+	currentAc := getCurrentAutomationConfig(mdb, mgr, t)
+	assert.Equal(t, 1, currentAc.Version)
+
+	mdbRef := &mdb
+	mdbRef.Spec.Members += 1
```
chatton, line 210: [nit] do we need to get the ref like this?

pkg/controller/mongodb/replicaset_controller_test.go
```diff
@@ -181,5 +182,56 @@ func TestService_isCorrectlyCreatedAndUpdated(t *testing.T) {
...
+	_ = mgr.GetClient().Update(context.TODO(), &mdb)
+	res, err = r.Reconcile(reconcile.Request{NamespacedName: types.NamespacedName{Namespace: mdb.Namespace, Name: mdb.Name}})
+	assertReconciliationSuccessful(t, res, err)
+
+	currentAc = getCurrentAutomationConfig(mdb, mgr, t)
+	assert.Equal(t, 2, currentAc.Version)
+
+}
+
+func TestAutomationConfig_versionIsNotBumpedWithNoChanges(t *testing.T) {
```
chatton, line 221: 👍

pkg/controller/mongodb/mongodb_controller.go
bznein: I see your point. In this case we would still need to do things with configmap inside a branch since not all errors from the get of the config map need to terminate the function (is the error is IsNotFound we are ok with that). So it would looke like this:

```golang
if err := r.client.Get(context.TODO(), types.NamespacedName{Name: mdb.ConfigMapName(), Namespace: mdb.Namespace}, currentCm); err != nil {
	if !errors.IsNotFound(err) {
		return corev1.ConfigMap{}, err
	}
} else {
	if err := json.Unmarshal([]byte(currentCm.Data[AutomationConfigKey]), &currentAc); err != nil {
		return corev1.ConfigMap{}, err
	}
}
```

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -265,24 +265,44 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
...
 		SetDomain(domain).
 		SetMembers(mdb.Spec.Members).
 		SetMongoDBVersion(mdb.Spec.Version).
-		SetAutomationConfigVersion(1). // TODO: Correctly set the version
+		SetAutomationConfigVersion(currentAc.Version).
 		SetFCV(mdb.GetFCV()).
 		AddVersion(mdbVersionConfig).
 		Build()
+
+	/* Here we compare the bytes of the two automationconfigs,
```
chatton, line 288: generally multi line comments in go use multiple single line comments (even though multi line comments exist!)

```golang
// this is
// a multiline
// comment!
```

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -323,9 +343,23 @@ func (r ReplicaSetReconciler) buildAutomationConfigConfigMap(mdb mdbv1.MongoDB)
...
+	currentAc := automationconfig.AutomationConfig{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: mdb.ConfigMapName(), Namespace: mdb.Namespace}, &currentCm); err != nil {
+		if !errors.IsNotFound(err) {
+			return corev1.ConfigMap{}, err
+		}
+	} else {
+		if err := json.Unmarshal([]byte(currentCm.Data[AutomationConfigKey]), &currentAc); err != nil {
+			return corev1.ConfigMap{}, err
+		}
+	}
```
chatton, line 356: we don't actually need this to be in an `else` clause. (another thing that keeps the happy path as left as possible)

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -265,24 +265,44 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
...
+		SetAutomationConfigVersion(currentAc.Version).
 		SetFCV(mdb.GetFCV()).
 		AddVersion(mdbVersionConfig).
 		Build()
+
+	/* Here we compare the bytes of the two automationconfigs,
+	 we can't use reflect.DeepEqual() as it treats nil entries as different from empty ones,
+	and in the AutomationConfig Struct we use omitempty to set empty field to nil */
+	newAcBytes, err1 := json.Marshal(newAc)
+	if err1 != nil {
```
chatton, line 292: generally, we name all error values `err`. Even if there is name shadowing. If possible, we like to reduce the scope of the variable using the construct

```golang
if err := SomeFunc(); err != nil {
   // err is only in scope here
}
```

but this is not always possible with multi value return functions. I would prefer just re-using the same error variable and re-assigning rather than multiple similarly named error variables

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -265,24 +265,44 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
...
+		return automationconfig.AutomationConfig{}, err1
+	}
+
+	currentAcBytes, err2 := json.Marshal(currentAc)
+	if err2 != nil {
+		return automationconfig.AutomationConfig{}, err2
+	}
+
+	if bytes.Compare(newAcBytes, currentAcBytes) != 0 {
```
chatton, line 301: 👍

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -265,24 +265,46 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
...
+		return automationconfig.AutomationConfig{}, err
+	}
+
+	currentAcBytes, err := json.Marshal(currentAc)
+	if err != nil {
+		return automationconfig.AutomationConfig{}, err
+	}
+
+	if bytes.Compare(newAcBytes, currentAcBytes) != 0 {
+		newAc.Version += 1
```
LouisPlisso, line 304: another approach would be the builder to take in the previous version and upon calling build decides if it bumps the number: thus the AA comparisons will happen inside the AA builder code which is more relevant

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -318,12 +340,33 @@ func buildService(mdb mdbv1.MongoDB) corev1.Service {
 		Build()
 }
 
+func getCurrentAutomationConfig(client mdbClient.Client, mdb mdbv1.MongoDB) (automationconfig.AutomationConfig, error) {
+	currentCm := corev1.ConfigMap{}
+	currentAc := automationconfig.AutomationConfig{}
+	if err := client.Get(context.TODO(), types.NamespacedName{Name: mdb.ConfigMapName(), Namespace: mdb.Namespace}, &currentCm); err != nil {
+		if !errors.IsNotFound(err) {
+			return automationconfig.AutomationConfig{}, err
+		}
```
LouisPlisso, line 349: so we discard is not found errors?

I find this code convoluted: can you please make it simpler?

LouisPlisso: I think making removing the 1st if would make this clearer

LouisPlisso: right now I think that if you get a is not found error while fetching the config map, you will return an empty automation config with no error
if this is intended: make it explicit and remove the else part

bznein: Yes @LouisPlisso you are right we are returning an empty AC with no error in this case.

A more explicit way to express it would look like this:

```go
	if err := client.Get(context.TODO(), types.NamespacedName{Name: mdb.ConfigMapName(), Namespace: mdb.Namespace}, &currentCm); err != nil {
		if errors.IsNotFound(err) {
			return automationconfig.AutomationConfig{}, nil
		} else {
			return automationconfig.AutomationConfig{}, err
		}

	}
	if err := json.Unmarshal([]byte(currentCm.Data[AutomationConfigKey]), &currentAc); err != nil {
		return automationconfig.AutomationConfig{}, err
	}
	return currentAc, nil
```

would you consider this (with appropriate comments) a better solution?

pkg/controller/mongodb/mongodb_controller.go
bznein: Would it make more sense to get rid of the function `setAutomationConfigVersion()` and replace it with a function that passes the previous version of the AC? Since the version should only be bumped in case of changes, this would prevent from setting the versions manually in unwanted ways. It would be set automatically to the one of the previous AC (0 if none is passed) and increased if something changed

pkg/controller/mongodb/mongodb_controller.go
LouisPlisso: way easier to read: I'd just remove the else (just return unindented as the if is returning), but it's whatever you prefer

pkg/controller/mongodb/mongodb_controller.go
LouisPlisso: yes that what i think

pkg/controller/mongodb/mongodb_controller.go
LouisPlisso: or maybe if errors.IsNotFound { err = nil } and then return AutomationConfig{}, err unconditionnaly

maybe that's the clearest?

LouisPlisso: and this one only needs comment on why we hide the not found error, everything else is self explanatory

pkg/controller/mongodb/mongodb_controller.go
LouisPlisso: and it's the build method that would call getCurrentAutomationConfig

pkg/controller/mongodb/mongodb_controller.go
bznein: I like the idea of having `if errors.isNotFound {err = nil }`

 I think it might be even taken a step further and have something analogous to what we have in [k8s_condition.py](https://github.com/mongodb/mongodb-kubernetes-operator/blob/master/scripts/dev/k8s_conditions.py#L48): we could have a function which takes a number of statuses (in this case it would be _metav1.StatusReasonNotFound_) and returns `nil` if the error is one of the provided ones.

So that would look like in our case:
```go
if err := client.Get(context.TODO(), types.NamespacedName{Name: mdb.ConfigMapName(), Namespace: mdb.Namespace}, &currentCm); err != nil {
         err  = makeErrorNilIfStatusIs(err, metav1.StatusReasonNotFound)
	return automationconfig.AutomationConfig{}, err
}
```

This is probably outside the scope of this PR though

chatton: this `makeErrorNilIfStatusIs` functionality already exists, in the form of `client.IgnoreNotFound(err)` (`"sigs.k8s.io/controller-runtime/pkg/client"`)

LouisPlisso: nice so in the end we could have

```
if err := xxx; err != nil {
    return automationconfig.AutomationConfig{}, client.IgnoreNotFound(err)
}
```
no more else

so much clearer

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -318,12 +330,30 @@ func buildService(mdb mdbv1.MongoDB) corev1.Service {
...
+	currentAc := automationconfig.AutomationConfig{}
+	if err := client.Get(context.TODO(), types.NamespacedName{Name: mdb.ConfigMapName(), Namespace: mdb.Namespace}, &currentCm); err != nil {
+		// If the AC was not found we don't treat it as an error
+		return automationconfig.AutomationConfig{}, k8sClient.IgnoreNotFound(err)
+
+	}
+	if err := json.Unmarshal([]byte(currentCm.Data[AutomationConfigKey]), &currentAc); err != nil {
+		return automationconfig.AutomationConfig{}, err
+	}
+	return currentAc, nil
```
LouisPlisso, line 344: so much clearer :)

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -265,24 +265,36 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
...
+	currentAc, err := getCurrentAutomationConfig(client, mdb)
+	if err != nil {
+		return automationconfig.AutomationConfig{}, err
+	}
+	newAc, err := automationconfig.NewBuilder().
 		SetTopology(automationconfig.ReplicaSetTopology).
 		SetName(mdb.Name).
 		SetDomain(domain).
 		SetMembers(mdb.Spec.Members).
+		SetPreviousAutomationConfig(currentAc).
```
LouisPlisso, line 287: I don't want to block on this, but maybe we can do the call to getCurrentAutomationConfig directly in the build function: as we always want to increment the version if it's different, the call should be by default and the closest possible to build time

you can do it later/never if you don't think it's important

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -318,12 +330,30 @@ func buildService(mdb mdbv1.MongoDB) corev1.Service {
 		Build()
 }
 
+func getCurrentAutomationConfig(client mdbClient.Client, mdb mdbv1.MongoDB) (automationconfig.AutomationConfig, error) {
+	currentCm := corev1.ConfigMap{}
+	currentAc := automationconfig.AutomationConfig{}
+	if err := client.Get(context.TODO(), types.NamespacedName{Name: mdb.ConfigMapName(), Namespace: mdb.Namespace}, &currentCm); err != nil {
+		// If the AC was not found we don't treat it as an error
```
LouisPlisso, line 337:
```suggestion
		// If the AC was not found we don't surface it as an error
```
nit

PR #62, opened 2020-06-16
Fix broken setup script for kind 0.8

fabianlindfors:
### All Submissions:

[kind 0.8](https://github.com/kubernetes-sigs/kind/releases/tag/v0.8.0) changed the way kind sets up networking. Previously, the kind container would be attached to the default Docker bridge network and our registry container would be attached to that by default as well. With 0.8, kind now creates its own network called `kind` which breaks our script as our registry can no longer be reached.

This PR fixes the issue by creating the `kind` network before kind does so itself and attaching the registry to that on creation. These changes are not backwards compatible and will require an update.

### DIFF
--- a/scripts/dev/setup_kind_cluster.sh
+++ b/scripts/dev/setup_kind_cluster.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -Eeou pipefail
 
+# create the kind network early unless it already exists.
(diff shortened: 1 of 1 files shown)


fabianlindfors:
Before running the script again the existing registry container has to be removed. Alternatively the existing container can be connected to the kind network using: `docker network connect kind kind-registry`

PR #63, opened 2020-06-18
CLOUDP-62444: Adopt functional approach for StatefulSet construction

chatton:
### All Submissions:

This PR replaces the StatefulSet builder with a functional approach.

This allows us to determine if a change is actually taking place without needing to perform an arbitrary time.Sleep (waiting for client cache to clear). We can now use a combination of comparing the before and after of the StatefulSet with out modification function to see if the StatefulSet is ready or not.

(there will be a separate PR to remove the builder and refactor tests if we go ahead with this design)

Side Notes:
* Before, we were creating the builder, then adding volumes, then ultimately calling build. Now we pass the volumes to the containers as required and keep everything in one place.
*  the `statefulset. HaveEqualSpec` function is no longer required and was removed


pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -195,15 +206,26 @@ func (r *ReplicaSetReconciler) resetStatefulSetUpdateStrategy(mdb mdbv1.MongoDB)
 }
 
 // isStatefulSetReady checks to see if the stateful set corresponding to the given MongoDB resource
-// is currently in the ready state
-func (r *ReplicaSetReconciler) isStatefulSetReady(mdb mdbv1.MongoDB, expectedUpdateStrategy appsv1.StatefulSetUpdateStrategyType) (bool, error) {
-	// TODO: This sleep will be addressed as part of https://jira.mongodb.org/browse/CLOUDP-62444
-	time.Sleep(time.Second * 5)
```
chatton, line 201: @LouisPlisso

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -366,36 +385,62 @@ func (r ReplicaSetReconciler) buildAutomationConfigConfigMap(mdb mdbv1.MongoDB)
...
+func getUpdateStrategyType(mdb mdbv1.MongoDB) appsv1.StatefulSetUpdateStrategyType {
+	if !mdb.IsChangingVersion() {
+		return appsv1.RollingUpdateStatefulSetStrategyType
 	}
+	return appsv1.OnDeleteStatefulSetStrategyType
+}
 
+// buildStatefulSet takes a MongoDB resource and converts it into
+// the corresponding stateful set
+func buildStatefulSet(mdb mdbv1.MongoDB) (appsv1.StatefulSet, error) {
```
LouisPlisso, line 399: 🥇

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -366,36 +385,62 @@ func (r ReplicaSetReconciler) buildAutomationConfigConfigMap(mdb mdbv1.MongoDB)
...
+	sts := appsv1.StatefulSet{}
+	buildStatefulSetModificationFunction(mdb)(&sts)
+	return sts, nil
```
LouisPlisso, line 402: if we can never return an error, then we can just get rid of it

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -148,12 +151,20 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	r.log.Debugf("Ensuring StatefulSet is ready, with type: %s", getUpdateStrategyType(mdb))
-	if ready, err := r.isStatefulSetReady(mdb, getUpdateStrategyType(mdb)); err != nil {
+	currentSts := appsv1.StatefulSet{}
+	if err := r.client.Get(context.TODO(), mdb.NamespacedName(), &currentSts); err != nil {
+		r.log.Warnf("Error getting StatefulSet: %s", err)
```
bznein, line 156: [nit] do we want to use `r.log.WarnF` here? (Almost) everywhere else we use `r.log.InfoF`(except one case in which we use `ErrorF`) Maybe we should use the same everywhere for consistency?

(btw I would be more in favour of using `WarnF` instead of `InfoF`

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -217,19 +239,16 @@ func (r *ReplicaSetReconciler) ensureService(mdb mdbv1.MongoDB) error {
 }
 
 func (r *ReplicaSetReconciler) createOrUpdateStatefulSet(mdb mdbv1.MongoDB) error {
-	sts, err := buildStatefulSet(mdb)
+	set := appsv1.StatefulSet{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: mdb.Name, Namespace: mdb.Namespace}, &set)
```
bznein, line 243: Here we can use the new `NamespacedName()` function you introduced

pkg/kube/persistantvolumeclaim/pvc.go
```diff
@@ -0,0 +1,33 @@
+package persistantvolumeclaim
```
bznein, line 1: The package is called `persistAntvolumeclaim` instead of `persistEntvolumeclaim`, which is the term used in k8s. Is this a typo?

pkg/kube/statefulset/statefulset.go
```diff
@@ -81,12 +80,94 @@ func IsReady(sts appsv1.StatefulSet, expectedReplicas int) bool {
...
+	return func(set *appsv1.StatefulSet) {
+		template := &set.Spec.Template
+		templateFunc(template)
+	}
+}
+
+func WithVolumeClaim(name string, f func(*corev1.PersistentVolumeClaim)) Modification {
+	return func(set *appsv1.StatefulSet) {
+		idx := findVolumeClaimIndexByName(name, set.Spec.VolumeClaimTemplates)
+		if idx == -1 {
```
bznein, line 149: We could have a `not_found` constant as in `podspec_template` here.

pkg/controller/mongodb/mongodb_controller.go
chatton: Yep we need to make sure we're consistent with our logging levels. 👍

pkg/kube/persistantvolumeclaim/pvc.go
chatton: almost definitely, nice catch!

PR #68, opened 2020-06-23
CLOUDP-65806: Add TLS/SSL settings to automation config

fabianlindfors:
### All Submissions:

This PR adds the settings necessary to implement TLS/SSL to the automation config. It only adds the old `ssl` fields and not the new `tls` fields in 4.2. This will be fixed in a later PR.

### DIFF
--- a/pkg/automationconfig/automation_config_builder.go
+++ b/pkg/automationconfig/automation_config_builder.go
@@ -26,7 +26,7 @@ type Builder struct {
 	previousAC        AutomationConfig
 	tlsCAFile         string
 	tlsCertAndKeyFile string
-	tlsMode           TLSMode
+	tlsMode           SSLMode
 	// MongoDB installable versions
 	versions []MongoDbVersionConfig
 }
@@ -64,7 +64,7 @@ func (b *Builder) SetFCV(fcv string) *Builder {
 	return b
 }
 
-func (b *Builder) SetTLS(caFile, certAndKeyFile string, mode TLSMode) *Builder {
+func (b *Builder) SetTLS(caFile, certAndKeyFile string, mode SSLMode) *Builder {
 	b.tlsCAFile = caFile
 	b.tlsCertAndKeyFile = certAndKeyFile
 	b.tlsMode = mode
@@ -72,7 +72,7 @@ func (b *Builder) SetTLS(caFile, certAndKeyFile string, mode TLSMode) *Builder {
 }
 
 func (b *Builder) isTLSEnabled() bool {
-	return b.tlsCAFile != "" && b.tlsCertAndKeyFile != "" && b.tlsMode != TLSModeDisabled
+	return b.tlsCAFile != "" && b.tlsCertAndKeyFile != "" && b.tlsMode != SSLModeDisabled
 }
 
 func (b *Builder) AddVersion(version MongoDbVersionConfig) *Builder {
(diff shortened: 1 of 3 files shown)


pkg/automationconfig/automation_config_builder.go
```diff
@@ -61,6 +64,17 @@ func (b *Builder) SetFCV(fcv string) *Builder {
...
 
+func (b *Builder) SetTLS(caFile, certAndKeyFile string, mode SSLMode) *Builder {
+	b.tlsCAFile = caFile
+	b.tlsCertAndKeyFile = certAndKeyFile
+	b.tlsMode = mode
+	return b
+}
+
+func (b *Builder) isTLSEnabled() bool {
+	return b.tlsCAFile != "" && b.tlsCertAndKeyFile != ""
```
chatton, line 75: one thing to watch out for, it's a valid configuration for these fields to be non-empty, if mode is set to `disabled` , in this case this function would incorrectly return `true`

fabianlindfors:
So we'll only support MongoDB 4.2 for the community operator?

pkg/automationconfig/automation_config_builder.go
fabianlindfors: Good catch, fixed!

pkg/automationconfig/automation_config_test.go
```diff
@@ -174,3 +175,33 @@ func TestVersionManifest_BuildsForVersion(t *testing.T) {
...
+		assert.Equal(t, mode, p.Args26.Net.SSL.Mode)
+		assert.Equal(t, caPath, p.Args26.Net.SSL.CAFile)
+		assert.Equal(t, certAndKeyPath, p.Args26.Net.SSL.PEMKeyFile)
+		assert.True(t, p.Args26.Net.SSL.AllowConnectionsWithoutCertificate)
+	}
+
+	// Ensure the CA was configured as trusted for the agent
+	assert.Equal(t, caPath, ac.SSL.CAFilePath)
+	assert.Equal(t, ClientCertificateModeOptional, ac.SSL.ClientCertificateMode)
+}
```
chatton, line 207: This looks good, there's just one other case I'd like to see a test for what happens if `mode` is SSLModeDisabled, but the other values are non-empty. (we can just test `isTLSEnabled` directly I think)

fabianlindfors: Are you thinking a new test function or just adding a new test case to `TestTLS`?

rodrigovalin:
@fabianlindfors Older versions of MongoDB will be supported as well. The agent will make sure that the configuration files are written with adequate values for the target MongoDB version.

fabianlindfors:
@rodrigovalin I didn't know about that, thanks! Of course we should use that instead. I'll update this PR to use the new fields.

fabianlindfors:
I swiched to `tls` settings instead but it turns out that we need a newer version of the agent for the conversion feature. According to @rodrigovalin we can't update the agent right now because the newer versions require mongodb-tools to be installed. For now, I've reverted to using the `ssl` settings instead and once we have updated the agent we can update these as well.

The agent update is tracked as CLOUDP-65918 internally.

pkg/automationconfig/automation_config_test.go
fabianlindfors: Unit test added!

PR #69, opened 2020-06-23
CLOUDP-64907: black linting hook and evg task

bznein:
### All Submissions:

This PR adds a pre-commit hook to ensure that python files are formatted using [black](https://github.com/psf/black). It also adds an evergreen tasks to check that the files are correctly formatted (in case the pre-commit hook is not correctly set up by the author of the commit).

The Contributing document is updated with instructions on how to add hooks to the local copy of the repository.


scripts/dev/templates/Dockerfile.linting
```diff
@@ -0,0 +1,7 @@
+FROM {{base_image}}
+
+ADD . .
```
chatton, line 3: do we need to add everything? We can reduce the context size by just including the python files

scripts/git-hooks/pre-commit
```diff
@@ -7,9 +7,17 @@ if ! type goimports &> /dev/null; then
...
     goimports -w "${file}"
     git add "$file"
 done
+
+
+# Black linting of every python file that was changed
+for file in $(git diff --cached --name-only --diff-filter=ACM | grep '\.py$')
+do
+    black -q "$file"
+    git add "$file"
```
rodrigovalin, line 22: I think this git add is redundant, the files are already staged

contributing.md
```diff
@@ -137,6 +137,13 @@ python scripts/dev/e2e.py --test <new-test>
 
 # Before Committing your code
 
-Please make sure you sign our Contributor Agreement
-[here](https://www.mongodb.com/legal/contributor-agreement). This will be
+## Set up pre-commit hooks
+To set up the pre-commit hooks, please create symbolic links from the provided [hooks](https://github.com/mongodb/mongodb-kubernetes-operator/tree/master/scripts/git-hooks):
+
+1. Navigate to your `.git/hooks` directory: `cd .git/hooks`
+1. Create a symlink for every file in the `scripts/git-hooks` directory: `ln -s -f  ../../scripts/git-hooks/* .`
```
chatton, line 144: [nit] I think it would be nice to have each command on its own line, so the full line is copy paste-able. Not a huge deal of course!

.dockerignore
```diff
@@ -3,6 +3,5 @@
 agent
 build/_output
 zz_*
-scripts
```
chatton, line 6: this now means we're copying all of our dev scripts into all of the containers that don't need them in the case of `ADD . .` Maybe we don't need to address it right now, but could be worth configuring different `.dockerignore`s

scripts/dev/templates/Dockerfile.linting
bznein: > I think black is only a Python formatter, but not linter. For linting we should use pylint.

This is a good point. How would a linter fit in the process in our case? 
As it just outputs diagnostics we can't automate the process as with a formatter.

bznein:
> I think black is only a Python formatter, but not linter. For linting we should use pylint.

This is a good point. How would a linter fit in the process in our case?
As it just outputs diagnostics we can't automate the process as with a formatter.

scripts/dev/templates/Dockerfile.linting
bznein: Yes we could do that. At the moment we are actually testing three directories (`test`,`pkg` and `scripts`) but should we assume that python scripts will always be in the latter? in this case we could just add that.

scripts/git-hooks/pre-commit
bznein: I just checked (commit 27d8845) and without the `git add` the change made by `black` is not staged for commit

.dockerignore
bznein: Yes that is a valid concern.  Do we want to address it now? depending on the version of Docker available on evergreen there might be a nice way to address it: https://docs.docker.com/engine/release-notes/#19030

scripts/git-hooks/pre-commit
rodrigovalin: yes, you are right, when the files are changed, they need to be staged again

.dockerignore
bznein: The evg host has docker 19.03, so I'll have a go with this option

scripts/dev/templates/Dockerfile.linting
bznein: I decided to go with listing all directories that contain python scripts and add them one by one

bznein: (actually, I modified it to add all the files one by one, not the directories)

scripts/ci/run_black_formatting_check.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+pip3 install black
```
rodrigovalin, line 3: is this needed here, considering that it was added to the requirements file?

bznein: This scripts runs inside a docker container (as black can't run on Python <3.6, which is what we have on evg), so the requirements are not installed on it. This install just black and not the whole set of requirements as they are not needed here

scripts/dev/dockerfile_generator.py
```diff
@@ -2,40 +2,53 @@
 import argparse
 
 
-def operator_params():
+def operator_params(adds):
```
chatton, line 5: I think this name could be improved, I didn't know what this was until I saw the description + Dockerfile template, how about `files_to_add`? (also, type hinting!)

scripts/dev/templates/Dockerfile.python_formatting
```diff
@@ -0,0 +1,13 @@
+FROM {{base_image}}
+
+
+{% for add in adds %}
```
chatton, line 4:
```suggestion
{% for file in files_to_add %}
```
looks a bit nicer I think!

scripts/dev/dockerfile_generator.py
```diff
@@ -48,12 +61,23 @@ def render(image_name):
...
+        "--adds",
+        help='Paths to use in the ADD command (defaults to ".")',
+        type=str,
+        default=".",
+    )
+    parser.add_argument(
+        "--script_location",
+        help="Location of the python script to run. (Used only for python_formatting)",
+        default="./scripts/ci/run_black_formatting_check.sh",
+    )
```
chatton, line 74: Can we add a check which makes sure that this is only used with our formatting image? We should immediately fail if it's provided otherwise. There may be a built in with arg parse for this, if not we can just do a simple check I think!

scripts/dev/dockerfile_generator.py
```diff
@@ -48,12 +61,23 @@ def render(image_name):
...
     return parser.parse_args()
 
 
 def main():
     args = parse_args()
-    print(render(args.image))
+    print(render(args.image, args.adds.split("\n"), args.script_location))
```
chatton, line 80: [nit] we can replace "\n" with `os.linesep`

bznein:
The PR has been updated: we found a way to use Python 3.6+ on evergreen so I reverted the changes to the Python scripts (f-strings) and moved the black check outside of docker.

I made substantial changes to .evergreen.yaml to use shell.exec instead of subprocess.exec, useful when running several bash commands one after the other (there are issues when using && to concatenate them). This could help us in the future to get rid of even more bash files.

.evergreen.yml
```diff
@@ -82,39 +89,35 @@ functions:
...
       params:
         working_dir: mongodb-kubernetes-operator
-        command: "pip3 install mypy"
-
-    - command: subprocess.exec
-      type: test
-      params:
-        working_dir: mongodb-kubernetes-operator
-        command: "find . -type f -name '*.py' -exec python3 -m mypy --ignore-missing-imports {} +"
+        script: |
```
chatton, line 117: same as above

.evergreen.yml
```diff
@@ -82,39 +89,35 @@ functions:
...
       type: test
       params:
         working_dir: mongodb-kubernetes-operator
-        add_to_path:
-          - ${workdir}/bin
-        env:
-          KUBECONFIG: ${workdir}/kube_config
-        command: "python3 ./scripts/dev/e2e.py --test ${test} --tag ${version_id} --config_file ./scripts/ci/config.json"
+        script: |
+            . venv/bin/activate
```
chatton, line 108: I think generally we want to try and avoid using `script` as it can get quite out of hand in general.

We could turn it into a small bash script in `scripts/ci`

and use the `binary` option instead of `script`

.evergreen.yml
```diff
@@ -2,6 +2,13 @@ ignore:
   - "*.md"
 
 functions:
+  setup_virtualenv:
```
chatton, line 5: maybe we can install the requirements in this function?

.evergreen.yml
```diff
@@ -82,39 +89,35 @@ functions:
...
-        command: "pip3 install mypy"
-
-    - command: subprocess.exec
-      type: test
-      params:
-        working_dir: mongodb-kubernetes-operator
-        command: "find . -type f -name '*.py' -exec python3 -m mypy --ignore-missing-imports {} +"
+        script: |
+            . venv/bin/activate
+            pip3 install mypy
```
chatton, line 119: if the setup of venv already includes mypy, this can turn back into a `command` that just runs mypy right?

scripts/ci/build_and_push_image.sh
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env bash
 
+. venv/bin/activate
+pip3 install -r ./requirements.txt
```
LouisPlisso, line 4: why not installing the requirements while we setup the virtualenv

bznein:
> we try to avoid using scripts in evergreen file: it creates shell scripts that are very hard to debug

Thanks I was not aware of it, I am removing them :)
> 
> why do we need to activate a virtualenv instead of directly calling the python script as before?

Without virtualenv we have Python 3.5.3, which does not support f-strings and most importantly which can not run `black`.

scripts/dev/dockerfile_generator.py
```diff
@@ -1,59 +1,93 @@
 import jinja2
 import argparse
+import os
 
 
-def operator_params():
+def operator_params(files_to_add):
```
LouisPlisso, line 6: we need type hints on parameters
```suggestion
def operator_params(files_to_add: List[str]):
```

scripts/dev/dockerfile_generator.py
```diff
@@ -1,59 +1,93 @@
...
     }
 
     render_values = param_dict.get(image_name, dict())
 
     env = jinja2.Environment()
     env.loader = jinja2.FileSystemLoader(searchpath="scripts/dev/templates")
-    return env.get_template("Dockerfile.{}".format(image_name)).render(render_values)
+    return env.get_template(f"Dockerfile.{image_name}").render(render_values)
+
+    return env.get_template(f"Dockerfile.{image_name}").render(render_values)
```
LouisPlisso, line 61: unreachable code

scripts/dev/dockerfile_generator.py
bznein: Thanks! This was removed due to issue with the Python version on evg and I forgot to to put them back when switching to virtualenv!

scripts/dev/dockerfile_generator.py
```diff
@@ -1,59 +1,93 @@
...
+            'The --script_location argument can be used only when the image used is "python_formatting"'
+        )
+
+    return args
 
 
 def main():
     args = parse_args()
-    print(render(args.image))
+    print(render(args.image, args.files_to_add.split(os.linesep), args.script_location))
```
LouisPlisso, line 90: it's good practice to have the main function return an int (type hinted)

and in the `if __name__ == "__main__":` you can `sys.exit(main())`

scripts/dev/dockerfile_generator.py
```diff
@@ -1,9 +1,11 @@
 import jinja2
 import argparse
 import os
+import sys
+from typing import List, Dict, Union
 
 
-def operator_params(files_to_add):
+def operator_params(files_to_add: List[str]) -> Dict[str, Union[bool, str, List[str]]]:
```
LouisPlisso, line 8: if Union[bool, str, List[str]] is used a lot it may be worth defining it (the hardest part would be to find a relevant name ;) )

chatton: `Dict[str, Union[bool, str, List[str]]]` is even less readable than `Dict` I think! Was this a mypy error?

scripts/dev/dockerfile_generator.py
```diff
@@ -1,60 +1,99 @@
...
+    files_to_add, script
+) -> Dict[str, Union[bool, str, List[str]]]:
+    return {
+        "base_image": "python:slim",
+        "files_to_add": files_to_add,
+        "script_location": script,
+    }
+
+
+def render(image_name, files_to_add, script_location) -> str:
```
chatton, line 52: can we add type hinting to arguments too?

bznein: > Dict[str, Union[bool, str, List[str]]] is even less readable than Dict I think! Was this a mypy error?

This was suggested by @LouisPlisso. I don't have a strong preference but I will get his suggestion of defining that one as a type and using that one each time.

> can we add type hinting to arguments too?


Sure thing thanks!

.dockerignore
LouisPlisso: @chatton I think this line has just been moved down, because it added a `/` at the end, so no change in behaviour should happen AFAIU

PR #72, opened 2020-06-25
CLOUDP-59488: diagnostic operator logs to evergreen

bznein:
## This PR:
- Moves the logic of e2e test on evergreen from bash to python
- Adds diagnostics data and logs to evergreen for easier debugging of failing tests

### Notes:
- `f-strings` are removed from python scripts as they require Python 3.6+ which is not installed in Evg (in the future we might move the e2e tests inside a container to overcome this limitation)
- the `config.json` file now needs more keys compared to before, so e2e tests won't run unless you change it with an updated one such as this (also added to contributing.md):

```json
{
    "namespace": "default",
    "repo_url": "localhost:5000",
    "operator_image": "mongodb-kubernetes-operator",
    "e2e_image": "e2e",
    "prestop_hook_image": "prehook",
    "testrunner_image": "test-runner"
}
```

### All Submissions:


scripts/dev/dump_diagnostic.py
```diff
@@ -0,0 +1,120 @@
...
+        mdb = crdv1.list_custom_resource_definition(pretty="true")
+        crd_log.write(yaml.dump(clean_nones(mdb.to_dict())))
+    except ApiException as e:
+        print("Exception when calling list_custom_resource_definition: %s\n" % e)
+
+
+def dump_persistent_volume(diagnosticFile: typing.TextIO):
+    corev1 = client.CoreV1Api()
+    try:
+        diagnosticFile.write(header("Persistent Volume Claims"))
```
rodrigovalin, line 43: Printing persistent volumes, but the title says "Persistent Volume Claims"

scripts/dev/dump_diagnostic.py
```diff
@@ -0,0 +1,120 @@
...
+        diagnosticFile.write(yaml.dump(clean_nones(mdb.to_dict())))
+    except ApiException as e:
+        print("Exception when calling list_namespaced_stateful_set: %s\n" % e)
+
+
+def dump_pod_log_namespaced(namespace: str, name: str):
+    corev1 = client.CoreV1Api()
+    try:
+        if name.startswith("mdb0"):
+            pod_file_mongoDB_agent = open(
```
rodrigovalin, line 64: please do not use uppercase in this identifier

scripts/ci/run_test.sh
```diff
@@ -1,40 +1,5 @@
...
-
-# The test will have fully finished when tailing logs finishes
-kubectl logs -f -l run=test-runner
-
-result="$(kubectl get pod -l run=test-runner -o jsonpath='{ .items[0].status.phase }')"
-if [[ ${result} != "Succeeded" ]]; then
-  exit 1
-fi
+pip3 install -r ./requirements.txt
+SKIP_CLEANUP="1" python3 ./scripts/dev/e2e.py --skip-operator-install 1 --skip-image-build 1 --test ${test:?} --tag ${version_id:?} --config_file ./scripts/ci/config.json --dump_diagnostic 1
```
chatton, line 5: thinking now, is it the best idea to mix env variables as well as cli arguments? Should we just add a `--skip-cleanup` instead?

scripts/dev/dump_diagnostic.py
```diff
@@ -0,0 +1,120 @@
...
+        mdb = av1beta1.list_namespaced_stateful_set(namespace, pretty="true")
+        diagnosticFile.write(yaml.dump(clean_nones(mdb.to_dict())))
+    except ApiException as e:
+        print("Exception when calling list_namespaced_stateful_set: %s\n" % e)
+
+
+def dump_pod_log_namespaced(namespace: str, name: str):
+    corev1 = client.CoreV1Api()
+    try:
+        if name.startswith("mdb0"):
```
rodrigovalin, line 63: this needs to support more cases?

test/e2e/replica_set_change_version/replica_set_test.go
```diff
@@ -18,7 +19,15 @@ func TestMain(m *testing.M) {
 
 func TestReplicaSetUpgradeVersion(t *testing.T) {
 	ctx := f.NewContext(t)
-	defer ctx.Cleanup()
+
+	// When the tests are ran in the evg host we don't want to clean up
+	// the context, as we need it to dump diagnostics
+	// In all other cases (e.g. tests ran locally) we still clean up
+	// the context when we finish
+	if os.Getenv("SKIP_CLEANUP") != "1" {
```
chatton, line 27: a constant for this value would be nice, especially if it's used in other tests

scripts/dev/dump_diagnostic.py
```diff
@@ -0,0 +1,120 @@
...
+            pod_file_mongoDB_agent = open(
+                "logs/e2e/{}-mongodb-agent.log".format(name), "w"
+            )
+            pod_file_mongod = open("logs/e2e/{}-mongod.log".format(name), "w")
```
rodrigovalin, line 67: use a with context manager here instead of manually closing the file

test/e2e/replica_set_change_version/replica_set_test.go
```diff
@@ -18,7 +19,15 @@ func TestMain(m *testing.M) {
...
+		defer ctx.Cleanup()
+	}
+
 	if err := e2eutil.RegisterTypesWithFramework(&mdbv1.MongoDB{}); err != nil {
```
chatton, line 31: maybe we could incorporate the above check into this function, or create a separate `TestSetup` type function that handles this for us. It could be easy to forget to include this check.

scripts/dev/dump_diagnostic.py
```diff
@@ -0,0 +1,120 @@
...
+            pod_file.write(log)
+            pod_file.close()
+    except ApiException as e:
+        print("Exception when calling read_namespaced_pod_log: %s\n" % e)
+
+
+def dump_pods_and_logs_namespaced(diagnosticFile: typing.TextIO, namespace: str):
+    corev1 = client.CoreV1Api()
+    try:
+        diagnosticFile.write(header("Pods"))
```
rodrigovalin, line 92: snake_case please 🐍

scripts/dev/dump_diagnostic.py
```diff
@@ -0,0 +1,120 @@
...
+
+def dump_all(namespace: str):
+
+    if not os.path.exists("logs"):
+        os.makedirs("logs")
+
+    if not os.path.exists("logs/e2e"):
+        os.makedirs("logs/e2e")
+
+    diagnostic_file = open("logs/e2e/diagnostics.txt", "w")
```
rodrigovalin, line 111: use context managers

scripts/dev/e2e.py
```diff
@@ -207,32 +220,65 @@ def parse_args():
...
+        "--dump_diagnostic",
+        help="Dump diagnostic information into files",
+        type=bool,
+        default=False,
+    )
+    parser.add_argument("--config_file", help="Path to the config file")
     return parser.parse_args()
 
 
 def main():
```
chatton, line 242: our main function is getting a big large. Maybe we could split some of these things up into smaller functions. (e.g. building & pushing images)

.evergreen.yml
```diff
@@ -138,59 +183,52 @@ tasks:
 
   - name: e2e_test_feature_compatibility_version
     commands:
-      - func: clone
-      - func: setup_kubernetes_environment
       - func: run_e2e_test
         vars:
           test: feature_compatibility_version
+  #    - func: upload_e2e_logs
```
rodrigovalin, line 189: no uploading of logs for now?

scripts/dev/e2e.py
```diff
@@ -7,6 +7,7 @@
     load_yaml_from_file,
 )
 import k8s_conditions
+import dump_diagnostic
```
chatton, line 10: [nit] I think there are 2 different responsibilities being pulled into one. 

* Accessing/Fetching diagnostics about K8s resources
* Writing them to disk

I think it would be a nice to have a mechanism to access the resources separated from the act of writing them. What do you think?

scripts/dev/dump_diagnostic.py
chatton, line 63: ❗ 
```suggestion
        if name.startswith(name):
```

scripts/dev/dump_diagnostic.py
```diff
@@ -0,0 +1,120 @@
...
+    av1beta1 = client.AppsV1Api()
+    try:
+        diagnosticFile.write(header("Stateful Sets"))
+        mdb = av1beta1.list_namespaced_stateful_set(namespace, pretty="true")
+        diagnosticFile.write(yaml.dump(clean_nones(mdb.to_dict())))
+    except ApiException as e:
+        print("Exception when calling list_namespaced_stateful_set: %s\n" % e)
+
+
+def dump_pod_log_namespaced(namespace: str, name: str):
```
chatton, line 60: I think this function is a bit specific to our use case, I think it will always be safe to just grab all the pods in a given namespace, and if any pod has multiple containers, we can just grab the logs of those too. I don't think we need to explicitly look for agent/mongod/operator names.

scripts/dev/dump_diagnostic.py
```diff
@@ -0,0 +1,120 @@
...
+            pod_file_mongod.close()
+            pod_file_mongoDB_agent.close()
+        elif name.startswith("mongodb-kubernetes-operator"):
+            pod_file = open("logs/e2e/{}.log".format(name), "w")
+            log = corev1.read_namespaced_pod_log(
+                name=name, namespace=namespace, pretty="true"
+            )
+            pod_file.write(log)
+            pod_file.close()
+    except ApiException as e:
```
chatton, line 85: The scope of this exception is very broad, it's not clear which operation can throw an `ApiException`

scripts/dev/dump_diagnostic.py
```diff
@@ -0,0 +1,120 @@
...
+                name=name, namespace=namespace, pretty="true", container="mongodb-agent"
+            )
+            log_mongod = corev1.read_namespaced_pod_log(
+                name=name, namespace=namespace, pretty="true", container="mongod"
+            )
+            pod_file_mongoDB_agent.write(log_mongodb_agent)
+            pod_file_mongod.write(log_mongod)
+            pod_file_mongod.close()
+            pod_file_mongoDB_agent.close()
+        elif name.startswith("mongodb-kubernetes-operator"):
```
chatton, line 78: same point as above, we shouldn't care if it's the operator pod or a database pod or an agent

scripts/ci/run_test.sh
```diff
@@ -1,40 +1,5 @@
...
-kubectl wait --for=condition=Ready pod -l run=test-runner --timeout=600s
-
-# The test will have fully finished when tailing logs finishes
-kubectl logs -f -l run=test-runner
-
-result="$(kubectl get pod -l run=test-runner -o jsonpath='{ .items[0].status.phase }')"
-if [[ ${result} != "Succeeded" ]]; then
-  exit 1
-fi
+pip3 install -r ./requirements.txt
```
rodrigovalin, line 4: This is just a suggestion/idea, no need to do this now 

This script can be replaced with a direct call from Evergreen with the form:

```yaml
  run_e2e_test:
    - command: subprocess.exec
      type: test
      params:
        add_to_path:
          - ${workdir}/bin
        working_dir: mongodb-kubernetes-operator
        command: "python3 ./scripts/dev/e2e.py --skip-operator-install 1 --skip-image-build 1 --test ${test} --tag ${version_id} --config_file ./scripts/ci/config.json --dump_diagnostic 1"
        env:
          KUBECONFIG: ${workdir}/kube_config
```

And if we add defaults to `skip-operator-install`, `skip-image-build` and `dump_diagnostic` we could make it even nicer.

scripts/dev/dump_diagnostic.py
chatton, line 92: we should still to `snake_case` for python variable names

scripts/ci/run_test.sh
```diff
@@ -1,40 +1,5 @@
...
-
-echo "Test pod is ready to begin"
-kubectl wait --for=condition=Ready pod -l run=test-runner --timeout=600s
-
-# The test will have fully finished when tailing logs finishes
-kubectl logs -f -l run=test-runner
-
-result="$(kubectl get pod -l run=test-runner -o jsonpath='{ .items[0].status.phase }')"
-if [[ ${result} != "Succeeded" ]]; then
-  exit 1
```
chatton, line 39: 🥳

scripts/ci/run_test.sh
rodrigovalin, line 5: This SKIP_CLEANUP worries me a little bit. I don't see how this variable is used by the Python script but only by the Go tests. 

Can you explicitely pass this all the way to the test? This means basically to not use env variables

test/e2e/feature_compatibility_version_upgrade/feature_compatibility_version_upgrade_test.go
```diff
@@ -17,7 +18,14 @@ func TestMain(m *testing.M) {
 
 func TestFeatureCompatibilityVersionUpgrade(t *testing.T) {
 	ctx := f.NewContext(t)
-	defer ctx.Cleanup()
+
+	// When the tests are ran in the evg host we don't want to clean up
+	// the context, as we need it to dump diagnostics
+	// In all other cases (e.g. tests ran locally) we still clean up
+	// the context when we finish
+	if os.Getenv("SKIP_CLEANUP") != "1" {
```
chatton, line 26: Unless I'm missing something, we need to add this `SKIP_CLEANUP` env variable to the E2E container still right?>

test/e2e/replica_set_change_version/replica_set_test.go
rodrigovalin, line 27: I think I know we can't escape this env variable for now, but it worries me that we use a env variable to notify the need of a cleanup.

.evergreen.yml
bznein: This was a comment left by mistake, the upload has been moved into a teardown task

scripts/ci/run_test.sh
bznein: I like it! thanks!

bznein:
I updated the PR: with regards to the SKIP_CLEANUP env variable, I switched to a flag passed to the python scripts, which then passes it as an env variable to the testrunner pod. Apparently, this is the only way we have to pass a parameter like this to our tests.

.evergreen.yml
```diff
@@ -70,12 +89,21 @@ functions:
         add_to_path:
           - ${workdir}/bin
         working_dir: mongodb-kubernetes-operator
-        include_expansions_in_env:
-          - version_id
-          - test
-        binary: scripts/ci/run_test.sh
         env:
           KUBECONFIG: ${workdir}/kube_config
+        command: "pip3 install -r ./requirements.txt"
```
bznein, line 94: I had to have this separated from the other one instead of:

`pip3 install -r ./requirements.txt && python ...` as evergreen was doing something weird and complaining about passing `--test` to pip.

PR #73, opened 2020-06-26
CLOUDP-65917: Fix pre-stop hook status check

fabianlindfors:
### All Submissions:

Before this PR, our pre-stop hook couldn't correctly identify a version change and would restart the pod for all automation config changes. This was because the hook would check for a `WaitFeatureCompatibilityVersionCorrect` move in the automation plan. Unfortunately, this move is used for most automation config changes and not just version changes.

After talking with the agent team, we came to the conclusion that it should be possible to only look for the `ChangeVersion` move. The reason this hadn't worked until now was that the agent wouldn't immediately write the plan to file and the hook would hence read an outdated version of the health status.

With this PR, we change the hook to wait until the health status file has been updated and the latest plan is available. We do this by polling the file and checking if the process is in goal state. If it is not, that indicates that the agent is in the middle of executing a new plan and we are ready to inspect it. Based on my testing it takes roughly 5-10 seconds from the hook starting to the health status file being updated, the hook will wait up to 60 seconds.

### DIFF
--- a/cmd/prestop/main.go
+++ b/cmd/prestop/main.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/agenthealth"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


cmd/prestop/main.go
```diff
@@ -203,24 +187,72 @@ func isWaitingToBeDeleted(healthStatus agenthealth.MmsDirectorStatus) bool {
...
-			// agent on the first MongoD stopped.
-			for _, s := range m.Steps {
-				if s.Step == "WaitFeatureCompatibilityVersionCorrect" &&
-					s.Result == "" {
-					return true
-				}
-			}
 		case "ChangeVersion":
-			// This is the condition observed in the 2nd and 3rd Pods.
 			return true
```
chatton, line 193: with a switch statement with a single case, I feel like a simple 

`if m.Move == "ChangeVersion"` is more straightforward

fabianlindfors: Of course, that's much better! Fixed now, thanks!

PR #74, opened 2020-06-26
CLOUDP-64906: mypy hook and task

bznein:
This PR:

- Adds a `mypy` check to the pre-commit hook and aborts the commit if `mypy` finds errors. 
- Adds an evg task that runs `mypy` on the codebase and fails if `mypy` finds errors
- Adds dummy git hooks to our directory, so that an user that clones the repo and symlinks them will always have then up-to-date
- Changes some python code so that it doesn't file the evg task introduced in this PR

### All Submissions:


scripts/git-hooks/pre-commit
```diff
@@ -13,3 +13,32 @@ do
...
+# and OR together the error codes to see if at least one failed
+exit_status=0
+echo "Running mypy on staged python files"
+for file in $(git diff --cached --name-only --diff-filter=ACM | grep '\.py$')
+do
+    echo "Analyzing $file ..."
+    # We ignore missing import otherwise mypy will complain
+    # about 3rd party libraries not providing type hints
+    mypy --ignore-missing-imports "${file}"
+    ((exit_status=exit_status||$?))
```
chatton, line 28: the `OR`ing makes it more complicated than it needs to be I think, what about something more basic like

```bash
if ! mypy --ignore-missing-imports "${file}"; then
  exit_status=1
fi
```

scripts/git-hooks/pre-commit
```diff
@@ -13,3 +13,32 @@ do
     goimports -w "${file}"
     git add "$file"
 done
+
+# Mypy doesn't support recursive traversal of directories
+# So we manually call it on every staged python file
+# and OR together the error codes to see if at least one failed
+exit_status=0
```
chatton, line 20:
```suggestion
local exit_status=0
```

I think with our bash scripts we should be as vigilant as possible to reduce the scope of our variables (even if it's just a single pre-commit hook!)

scripts/dev/dev_config.py
```diff
@@ -43,15 +43,18 @@ def testrunner_image(self):
...
+            return DevConfig(json.loads(f.read()))
+    except FileNotFoundError:
+        print(
+            "No DevConfig found. Please ensure that the configuration file exists at '{}'".format(config_file_path)
         )
-    )
-    return None
+        raise
+    except Exception:
+        raise
```
chatton, line 60: just seeing now, we don't need to actually need this generic catch all if we just want to  re-throw this exception, this will happen anyway if it is not of type `FileNotFoundError`

scripts/git-hooks/pre-commit
```diff
@@ -2,14 +2,54 @@
...
+    fi
+
+    # Formats each file that was changed.
+    for file in $(git diff --cached --name-only --diff-filter=ACM | grep '\.go$')
+    do
+        goimports -w "${file}"
+        git add "$file"
+    done
+
+}
```
chatton, line 18: 🥳

PR #75, opened 2020-06-26
CLOUDP-58345: Add TLS support

fabianlindfors:
### All Submissions:

## Description

This PR implements TLS encryption for client-server and server-server communication. The implementation is based around a single certificate-key pair stored inside of a secret which is used by all servers. We have also discussed using a unique cert-key pair for each server but decided to start with a shared one for simplicity.

The certificate and key will have to be generated by the user and to aid in this we have chosen a format which is both compatible with Kubernetes and the popular [cert-manager](https://cert-manager.io/). The secret is expected to have two fields, `tls.crt` and `tls.key`, containing the PEM-encoded certificate and private key. Such a secret can be created with the following command `kubectl create secret tls NAME --cert=/path/to/certificate --key=/path/to/key`.

To verify these certificates, the servers will also need the CA certificate which the user should supply as a ConfigMap. This certificate should also be PEM-encoded and stored in a field named `ca.crt`.

The operator will mount the Secret and ConfigMap to each pod. MongoDB requires the server cert-key pair to be provided as a single, concatenated PEM file but the mounted secret will end up in two, separate files. To solve this, each pod will run an init container which concatenates the certificate and key and writes them to a volume shared with the agent and `mongod`.

## Example

```
apiVersion: mongodb.com/v1
kind: MongoDB
metadata:
  name: example-mongodb
spec:
  members: 3
  type: ReplicaSet
  version: "4.2.8"
  tls:
    enabled: true
    caConfigMapRef: "example-mongodb-ca"
    serverSecretRef: "example-mongodb-tls"
```

## Upgrading an existing cluster to use TLS

The upgrade from non-TLS to TLS is done in two steps:

1. The operator updates the pod template with the Secret and ConfigMap volume mounts. The statefulset will perform a rolling upgrade and which makes the certs and keys available on all pods. Once that is done, the operator will set an annotation on the MongoDB resource called `tlsRolledOut` which is used to determine that the operator should move on to step 2.
2. Once the rollout is complete, the operator will update the automation config ConfigMap and enable TLS. This change will be picked up by each agent and they'll proceed with the upgrade process outlined in: https://docs.mongodb.com/manual/tutorial/upgrade-cluster-to-ssl/.

Once TLS is enabled it will be required by default and clients must use it ([`requireSSL`](https://docs.mongodb.com/manual/reference/configuration-options/#net.tls.mode)). This could cause trouble during the upgrade procedure as clients haven't been updated to use TLS yet. To aid the upgrade, there is a setting for making TLS optional. When this is enabled, server-server communication will be encrypted but clients can connect both with and without TLS ([`preferTLS`](https://docs.mongodb.com/manual/reference/configuration-options/#net.tls.mode)). Once all clients are securely connected the setting can be disabled again.

```
tls:
  enabled: true
  optional: true
  caConfigMapRef: "example-mongodb-ca"
  serverSecretRef: "example-mongodb-tls"
```

## Questions

- Where are we going to document how to use TLS?

## Next steps

1. Add E2E tests for TLS and the upgrade procedure. This might also be a good time to update the CRD. Tracked internally by [CLOUDP-66230](https://jira.mongodb.org/browse/CLOUDP-66230).
2. ~~Implement support for disabling TLS on a running cluster. Tracked internally by [CLOUDP-66231](https://jira.mongodb.org/browse/CLOUDP-66231).~~ We have decided that this is not a priority at this time.
3. Use the new `tls` settings instead of the deprecated `ssl` settings for the automation config. The latest version of the agent will let us use `tls` and take care of backwards-compatibility. We are still working on upgrading the agent which is tracked internally by [CLOUDP-65918](https://jira.mongodb.org/browse/CLOUDP-65918).


pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -291,24 +336,75 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
...
+
+	// Ensure Secret exists
+	var secret corev1.Secret
+	secretName := types.NamespacedName{Name: mdb.Spec.TLS.ServerSecretRef, Namespace: mdb.Namespace}
+	if err := r.client.Get(context.TODO(), secretName, &secret); err != nil {
+		return errors.IsNotFound(err), err
+	}
+
+	// Ensure Secret has "tls.crt" and "tls.key" fields
+	if key, ok := secret.Data[tlsSecretKeyName]; !ok || len(key) == 0 {
```
chatton, line 367: again we have some functions in the `secret` package that will handle the common operations (or add some more if they don't meet your exact use case!)

a `secret.RequireKeys(secretGetter Getter, keys ...string) error` sounds like it could be a good fit

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -502,6 +603,40 @@ func buildStatefulSetModificationFunction(mdb mdbv1.MongoDB) statefulset.Modific
 
 	dataVolume := statefulset.CreateVolumeMount(dataVolumeName, "/data")
 
+	agentVolumeMounts := []corev1.VolumeMount{agentHealthStatusVolumeMount, automationConfigVolumeMount, dataVolume}
+	mongodVolumeMounts := []corev1.VolumeMount{mongodHealthStatusVolumeMount, hooksVolumeMount, dataVolume}
+
+	tlsPodSpec := podtemplatespec.Apply()
```
chatton, line 609: there should be a `podtemplatespec.NOOP()` that is semantically more correct for what you need here

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -502,6 +603,40 @@ func buildStatefulSetModificationFunction(mdb mdbv1.MongoDB) statefulset.Modific
...
+
+		// Configure a volume which mounts the secret holding the server key and certificate
+		// The same key-certificate pair is used for all servers
+		tlsSecretVolume := statefulset.CreateVolumeFromSecret("tls-secret", mdb.Spec.TLS.ServerSecretRef)
+		tlsSecretVolumeMount := statefulset.CreateVolumeMount(tlsSecretVolume.Name, tlsSecretMountPath, statefulset.WithReadOnly(true))
+
+		// MongoDB expects both key and certificate to be provided in a single PEM file
+		// We are using a secret format where they are stored in separate fields, tls.crt and tls.key
+		// Because of this we need to use an init container which reads the two files mounted from the secret and combines them into one
+		tlsPodSpec = podtemplatespec.Apply(
```
chatton, line 632: very good candidate for a separate function to build this function I think

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -291,24 +336,75 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
 	return nil
 }
 
-func buildAutomationConfig(mdb mdbv1.MongoDB, mdbVersionConfig automationconfig.MongoDbVersionConfig, client mdbClient.Client) (automationconfig.AutomationConfig, error) {
-	domain := getDomain(mdb.ServiceName(), mdb.Namespace, "")
+// checkTLSConfig will check that the configured ConfigMap and Secret exist and that they have the correct fields.
```
rodrigovalin, line 339: The docs are a bit misleading, the first return value is a "shouldRetry", but I had to go and look at the usages to understand it.

Returning a "true" value from a `checkTLSConfig` should indicate that *everything is ok*, but in this case, it means that the *process should retry*, which is kind of confusing.

At the same time, I think that both values can be wrapped into the same error type: if the function returns err == nil, we can proceed. Reconcile otherwise. <- this can be discussed

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -25,6 +25,9 @@ const (
 	// LastVersionAnnotationKey should indicate which version of MongoDB was last
 	// configured
 	LastVersionAnnotationKey = "lastVersion"
+
+	// TLSRolledOutKey indicates if TLS has been fully rolled out
+	TLSRolledOutKey = "tlsRolledOut"
```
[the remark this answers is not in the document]
fabianlindfors: Good one! I'm totally on board with doing this now. How would we go about migrating from the old `lastVersion` to `mongodb.com/v1.lastVersion` for existing clusters?

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -42,6 +45,29 @@ type MongoDBSpec struct {
...
+	Enabled bool `json:"enabled"`
+
+	// Optional configures if TLS should be required or optional for connections
+	// +optional
+	Optional bool `json:"optional"`
+
+	// ServerSecretRef is the name of a secret containing a private key and certificate to use for TLS
+	// The key and cert are expected to be PEM encoded and available at "tls.key" and "tls.crt"
+	// +optional
+	ServerSecretRef string `json:"serverSecretRef"`
```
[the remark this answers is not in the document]
fabianlindfors: Good point. As Secrets and ConfigMaps can only be referenced in the same namespace as the pods themselves we really only want the name. How about we change the naming to `serverSecretName` or just `serverSecret`?

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -181,6 +200,10 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	if err := r.completeTLSRollout(mdb); err != nil {
+		return reconcile.Result{}, err
```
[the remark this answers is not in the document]
fabianlindfors: Fixed!

pkg/apis/mongodb/v1/mongodb_types.go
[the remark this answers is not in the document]
chatton: I don't know if we need to worry about this. To be safe we could just do a `|| lastVersion`, but always set the new tag.

pkg/apis/mongodb/v1/mongodb_types.go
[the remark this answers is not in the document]
chatton: If we only ever need the name, and won't need to configure keys differently, then I'm okay with the name! (with the name change!)

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -252,6 +275,28 @@ func (r *ReplicaSetReconciler) createOrUpdateStatefulSet(mdb mdbv1.MongoDB) erro
 	return nil
 }
 
+// completeTLSRollout will update the automation config and set an annotation indicating that TLS has been rolled out.
+// At this stage, TLS hasn't yet been enabled but the keys and certs have all been mounted.
+// The automation config will be updated and the agents will continue work on gradually enabling TLS across the replica set.
+func (r *ReplicaSetReconciler) completeTLSRollout(mdb mdbv1.MongoDB) error {
+	if mdb.Spec.TLS.Enabled && !mdb.HasRolledOutTLS() {
```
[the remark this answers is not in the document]
fabianlindfors: Totally agree, I like the early-return pattern. Fixed!

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -291,24 +336,75 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
 	return nil
 }
 
-func buildAutomationConfig(mdb mdbv1.MongoDB, mdbVersionConfig automationconfig.MongoDbVersionConfig, client mdbClient.Client) (automationconfig.AutomationConfig, error) {
-	domain := getDomain(mdb.ServiceName(), mdb.Namespace, "")
+// checkTLSConfig will check that the configured ConfigMap and Secret exist and that they have the correct fields.
+func (r *ReplicaSetReconciler) checkTLSConfig(mdb mdbv1.MongoDB) (bool, error) {
+	if !mdb.Spec.TLS.Enabled {
+		return false, nil
+	}
```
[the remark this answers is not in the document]
fabianlindfors: I'm reliably inconsistent ;)

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -145,6 +153,17 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	if shouldRetry, err := r.checkTLSConfig(mdb); err != nil {
+		if shouldRetry {
```
LouisPlisso, line 157: I don't understand this: the pattern is if err is not nil, the value returned should not be used

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -291,24 +336,75 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
...
-	currentAc, err := getCurrentAutomationConfig(client, mdb)
-	if err != nil {
-		return automationconfig.AutomationConfig{}, err
+	r.log.Info("Ensuring TLS is correctly configured")
+
+	// Ensure CA ConfigMap exists
+	var caConfigMap corev1.ConfigMap
+	configMapName := types.NamespacedName{Name: mdb.Spec.TLS.CAConfigMapRef, Namespace: mdb.Namespace}
+	if err := r.client.Get(context.TODO(), configMapName, &caConfigMap); err != nil {
+		return errors.IsNotFound(err), err
```
LouisPlisso, line 351: if you want to use IsNotFound you have to check and then either return true, nil or false, err

you should not return a value and an error

pkg/controller/mongodb/mongodb_controller.go
LouisPlisso: I think this function is creating a lot of confusion and should be handled differently

maybe using a decorator pattern that handles the retry will make it clearer

pkg/controller/mongodb/mongodb_controller.go
LouisPlisso: yes the whole function is nearly 100 lines

pkg/controller/mongodb/mongodb_controller.go
fabianlindfors: Didn't know about that, thanks!

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -42,6 +45,29 @@ type MongoDBSpec struct {
...
+	TLS TLS `json:"tls"`
+}
+
+// TLS is the configuration used to set up TLS encryption
+type TLS struct {
+	Enabled bool `json:"enabled"`
+
+	// Optional configures if TLS should be required or optional for connections
+	// +optional
+	Optional bool `json:"optional"`
```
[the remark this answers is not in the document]
fabianlindfors: How would a user handle the case of upgrading an existing cluster to TLS and upgrading their clients to TLS as well? When TLS is enabled and becomes required, existing non-TLS clients would get disconnected. The documentation recommends doing this as part of the upgrade procedure in step 2: https://docs.mongodb.com/manual/tutorial/upgrade-cluster-to-ssl/.

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -42,6 +45,29 @@ type MongoDBSpec struct {
...
+
+	// ServerSecretRef is the name of a secret containing a private key and certificate to use for TLS
+	// The key and cert are expected to be PEM encoded and available at "tls.key" and "tls.crt"
+	// +optional
+	ServerSecretRef string `json:"serverSecretRef"`
+
+	// CAConfigMapRef is the name of a ConfigMap containing the certificate for the CA which signed the server certificates
+	// The certificate is expected to be available under the key "ca.crt"
+	// +optional
+	CAConfigMapRef string `json:"caConfigMapRef"`
```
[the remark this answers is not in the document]
fabianlindfors: My understanding is that not all cert-manager certs will have the CA field set. It seems to depend on the issuer. For example, the Let's Encrypt issuer will leave that field empty: https://github.com/jetstack/cert-manager/issues/1571.

pkg/controller/mongodb/mongodb_controller.go
fabianlindfors: I think you're right, the returned boolean is confusing. Positive bool should definitely indicate a proper TLS config.

Regarding just returning an error, I separated these because there are two error cases:
1. The user hasn't set up the required secret or configmap correctly. In this case we want to retry in a little while to give the user time to fix it.
2. Something fails when checking the resources, for example a intermittent error. In this case we want to retry immediately.

1. could of course be implemented by returning an error as well but that would cause an immediate requeue and the operator looping many times before the user has a chance to update the configuration.

fabianlindfors: @LouisPlisso your comments basically come down to this as well. Would love to hear your take on whether we should do an immediate requeue, and if not, how to best handle this three case situation.

pkg/apis/mongodb/v1/mongodb_types.go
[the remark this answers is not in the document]
fabianlindfors: I've changed the names to `serverSecretName` and `caConfigMapName`.

pkg/controller/mongodb/mongodb_controller.go
fabianlindfors: I've broken this out into a separate function now by introducing a new builder function: `podtemplatespec.WithVolumeMounts()`. The problem before was that the volume mounts needed to be applied to each container object which made it hard to refactor.

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -42,6 +45,29 @@ type MongoDBSpec struct {
 	// be set for the deployment
 	// +optional
 	FeatureCompatibilityVersion string `json:"featureCompatibilityVersion,omitempty"`
+
+	// TLS configuration for both client-server and server-server communication
+	// +optional
+	TLS TLS `json:"tls"`
```
chatton, line 51: ah, meant to say about this. In enterprise, the current structure is `spec.security.tls` as we have other `spec.security.authentication` fields as well. Do we want to do the same for community? (eg. SCRAM is coming soon)

fabianlindfors: @rodrigovalin what do you think?

pkg/apis/mongodb/v1/mongodb_types.go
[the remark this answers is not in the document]
rodrigovalin: In that case we can discard this comment.

pkg/apis/mongodb/v1/mongodb_types.go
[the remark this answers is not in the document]
rodrigovalin: I think you are right, although I think it is unfortunate the way we phrase this. Docs should be "make sure all of your clients can work with both TLS and non-TLS enabled clusters and then upgrade the cluster to TLS only.

This is exactly the same as having an HTTPS server that also handles HTTP in case the client does not support HTTPS. which I don't think anyone would agree that it is a good idea.

Anyway, I think I'm fine with having optional in here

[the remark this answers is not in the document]
fabianlindfors: The recommendation might stem from the fact that a lot of drivers don't seem to support automatic switching between TLS and non-TLS. Nevertheless, we are probably best of following the official recommendations.

pkg/apis/mongodb/v1/mongodb_types.go
rodrigovalin: I have a preference for `spec.tls` but not strong enough; however I think it is a good idea to maintain consistency with enterprise. Let's make this `spec.security.tls` instead.

fabianlindfors: I have no preference, I'll change it to `spec.security.tls`!

pkg/controller/mongodb/mongodb_controller.go
LouisPlisso: @fabianlindfors I think the confusion comes from this: don't spare an if and return an error if it's not an IsNotFound otherwise return a retry true without error

pkg/controller/mongodb/mongodb_controller.go
LouisPlisso: see comment below: a retryable error should not be surfaced as an error
if err is not nil, then you can use the retryable value

if I understand the code correctly, right now we have these possible outputs: (false, nil), (true, err), (false, err)

golang pattern would be to have: (false, nil), (true, nil), (false, err) [false with err because it's the default value of bool, but it should never be used if err is non-nil)

does that help?

fabianlindfors: Right, I agree! I'm not sure how to surface the feedback message for the user though in case they have something wrong in their configuration.

Basically right now I'm encoding both user errors and other errors in the return `error` which is dumb as it's not idiomatic Go. Using `(false, nil)`, `(true, nil)`, `(false, err)` wouldn't make it possible to return an error message for the user to see what is wrong with their config.

I have a few ideas but I'm not sure which one is the most idiomatic:
1. Return three values: `(valid bool, validationMessage string, err error)`. If `valid` is true then the reconcilation proceeds, if `err` is not nil then two others are ignored, and if `valid` is false then `validationMessage` contains some feedback for the user.
2. Encode the validation result in a new type `TLSValidationResult` and return `(TLSValidationResult, error)`. This struct could contain two values: `Valid bool` and `ValidationMessage string`. This is similar to solution 1.
3. Create a new error type for user config errors and then have a helper method that can check if an error is a user error (similarly to the Kubernetes error helpers). This way the function would only have to return a single `error`. This is my least favourite as the different cases get lost in the type system.

pkg/apis/mongodb/v1/mongodb_types.go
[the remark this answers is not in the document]
fabianlindfors: I think you're right, we don't have to worry about this. The operator will think the version hasn't changed as the new annotation is not set. It will then get updated at the end of the reconciliation loop.

Fixed :)

pkg/controller/mongodb/mongodb_controller.go
chatton: @fabianlindfors This is something has come up a lot, we want to represent more than two possible outcomes, but the restriction of `reconcilliation.Result, error` doesn't allow us to adequately represent the various states we encounter. I encourage you to take a look at the [workflow](https://github.com/10gen/ops-manager-kubernetes/blob/3877ddc263ec5576d8ceb98003ee35e819dbac7e/pkg/controller/operator/workflow/status.go#L13) package we have in enterprise (not to do a direct port, but to maybe get some ideas from it)

I think we will need to introduce something similar to these to account for all the states we will encounter.

chatton: I think the type of refactor I suggested is something for a future PR, of your suggestions, I like number 2 the most, with the intention of creating something that we can generally use to express different types of results.

fabianlindfors: Sounds like a good plan! I've created an internal ticket CLOUDP-66375 to create something similar to the workflow package for community.

I'll go with number 2 for now which should improve readability.

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -291,24 +336,75 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
...
+	}
 
-	currentAc, err := getCurrentAutomationConfig(client, mdb)
-	if err != nil {
-		return automationconfig.AutomationConfig{}, err
+	r.log.Info("Ensuring TLS is correctly configured")
+
+	// Ensure CA ConfigMap exists
+	var caConfigMap corev1.ConfigMap
+	configMapName := types.NamespacedName{Name: mdb.Spec.TLS.CAConfigMapRef, Namespace: mdb.Namespace}
```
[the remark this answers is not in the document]
fabianlindfors: Done!

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -291,24 +336,75 @@ func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
...
+	if err := r.client.Get(context.TODO(), configMapName, &caConfigMap); err != nil {
+		return errors.IsNotFound(err), err
 	}
-	newAc, err := automationconfig.NewBuilder().
+
+	// Ensure ConfigMap has a "ca.crt" field
+	if cert, ok := caConfigMap.Data[tlsCACertName]; !ok || cert == "" {
```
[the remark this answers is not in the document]
fabianlindfors: This is a bit tricky right now as we need to separate not found errors from the rest. I would suggest holding off on this until we implement a better way of handling reconciliation results. What do you think?

pkg/controller/mongodb/mongodb_controller.go
fabianlindfors: I've now implemented option 2 with a type called `tlsValidationResult`. This means that the function, now called `validateTLSConfig`, returns `(tlsValidationResult, error)`. `error` is only set if there was an actual error during validation and not for user errors.

CLOUDP-66375 will replace this with a more general solution for handling multi-case reconciliation results.

pkg/controller/mongodb/mongodb_controller.go
fabianlindfors: Check out the discussion we had in the other thread. I've implemented a solution which is more idiomatic.

pkg/controller/mongodb/mongodb_controller.go
fabianlindfors: @rodrigovalin do you like the new interface for `validateTLSConfig` more?

fabianlindfors:
Based on the feedback received I've changed the CRD format to be:

```
apiVersion: mongodb.com/v1
kind: MongoDB
metadata:
  name: example-mongodb
spec:
  members: 3
  type: ReplicaSet
  version: "4.2.8"
  security:
    tls:
      enabled: true
      caConfigMapName: "example-mongodb-ca"
      serverSecretName: "example-mongodb-tls"
```

I've also broken out the TLS functionality to a separate file, next to `mongodb_controller.go`. Me and @chatton have discussed breaking this out to a separate package as well and decided to do it at a later time. @chatton is currently experimenting with this for his SCRAM-SHA implementation and once he's found a nice design for that we might be able to give TLS the same treatment.

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -0,0 +1,144 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+
+	mdbv1 "github.com/mongodb/mongodb-kubernetes-operator/pkg/apis/mongodb/v1"
+)
+
+type tlsValidationResult struct {
+	Valid   bool
+	Message string
```
LouisPlisso, line 19: I don't see the message being used ever

LouisPlisso: this makes me think we only need a return type of bool, error for validateTLSConfig

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -472,6 +513,7 @@ func buildStatefulSetModificationFunction(mdb mdbv1.MongoDB) statefulset.Modific
 	labels := map[string]string{
 		"app": mdb.ServiceName(),
 	}
+	fmt.Printf("%+v\n", mdb)
```
chatton, line 516: ❗ make sure to remove this

pkg/controller/mongodb/mongodb_tls.go
fabianlindfors: `Message` is used on line 162 where it's logged for the user to see.

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -0,0 +1,144 @@
...
+}
```
chatton, line 20: I think this is a good middle ground for now 👍

pkg/controller/mongodb/mongodb_tls.go
chatton, line 19: can we make this unexported? I don't think we should have an unexported type with exported fields

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -0,0 +1,144 @@
...
+func (r *ReplicaSetReconciler) validateTLSConfig(mdb mdbv1.MongoDB) (tlsValidationResult, error) {
+	if !mdb.Spec.Security.TLS.Enabled {
+		return tlsValidationResult{Valid: true}, nil
+	}
+
+	r.log.Info("Ensuring TLS is correctly configured")
+
+	// Ensure CA ConfigMap exists
+	var caConfigMap corev1.ConfigMap
+	if err := r.client.Get(context.TODO(), mdb.TLSConfigMapNamespacedName(), &caConfigMap); err != nil {
```
chatton, line 34: maybe here we can use the `configmap` package

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -0,0 +1,144 @@
...
+
+	// Ensure ConfigMap has a "ca.crt" field
+	if cert, ok := caConfigMap.Data[tlsCACertName]; !ok || cert == "" {
+		message := fmt.Sprintf(`ConfigMap "%s" should have a CA certificate in field "%s"`, mdb.TLSConfigMapNamespacedName().String(), tlsCACertName)
+		return tlsValidationResult{Valid: false, Message: message}, nil
+	}
+
+	// Ensure Secret exists
+	var secret corev1.Secret
+	if err := r.client.Get(context.TODO(), mdb.TLSSecretNamespacedName(), &secret); err != nil {
```
chatton, line 53: same thing about `secret` package

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -0,0 +1,144 @@
...
+			message := fmt.Sprintf(`Secret "%s" not found`, mdb.TLSSecretNamespacedName().String())
+			return tlsValidationResult{Valid: false, Message: message}, nil
+		}
+
+		return tlsValidationResult{}, err
+	}
+
+	// Ensure Secret has "tls.crt" and "tls.key" fields
+	if key, ok := secret.Data[tlsSecretKeyName]; !ok || len(key) == 0 {
+		message := fmt.Sprintf(`Secret "%s" should have a key in field "%s"`, mdb.TLSSecretNamespacedName().String(), tlsSecretKeyName)
```
chatton, line 64: don't think we need the explicit `String()` call, this is handled by `Sprintf`

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -0,0 +1,144 @@
...
+
+	if err := r.setAnnotation(mdb.NamespacedName(), mdbv1.TLSRolledOutKey, "true"); err != nil {
+		return fmt.Errorf("error setting TLS annotation: %+v", err)
+	}
+
+	return nil
+}
+
+// buildTLSPodSpecModification will add the TLS init container and volumes to the pod template if TLS is enabled.
+func buildTLSPodSpecModification(mdb mdbv1.MongoDB) podtemplatespec.Modification {
```
chatton, line 98: very nice 👍

pkg/controller/mongodb/mongodb_controller.go
fabianlindfors: Whoops, removed!

pkg/controller/mongodb/mongodb_tls.go
fabianlindfors: Yes, should definitely be unexported

pkg/controller/mongodb/mongodb_tls.go
fabianlindfors: Didn't know about that, fixed!

pkg/controller/mongodb/mongodb_tls.go
fabianlindfors: Good points. Switched to configmap and secret now!

pkg/controller/mongodb/mongodb_tls.go
LouisPlisso: @fabianlindfors why don't we log it when the error is happening in validateTLSConfig, and have a simple bool, error return type

fabianlindfors: @LouisPlisso We definitely could but I mostly put it on that level to stay consistent with the other error logs, they all appear in the reconciliation loop. You think we should break that pattern for now? It will be fixed once we find a better way to handle the reconciliation result with CLOUDP-66375.

pkg/controller/mongodb/mongodb_controller.go
LouisPlisso: @fabianlindfors you mention you need a 3 case solution: bool, err allows for 3 cases as shown above
this is making it un-necessarily complicated
why bool, err is not working for you? you only need the struct to hold the log message: logging it right away would solve this and also acting closest to the error is a golang best practice (as here our action is only to log, why not doing it here)

fabianlindfors: The only reason is trying to stay consistent with our style of bubbling errors up to the `Reconcile()` and then logging there. Adding complexity just to stay in line with that though doesn't feel right, I'll change it!

pkg/controller/mongodb/mongodb_controller.go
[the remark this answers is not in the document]
fabianlindfors: I take that back. `ReadData` definitely works and has now been incorporated. I'll hold off on the other suggested helper function until we have proper handling of the reconciliation result in place.

pkg/controller/mongodb/mongodb_controller.go
fabianlindfors: @LouisPlisso This has now been changed. Please have another look!

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -17,17 +17,14 @@ import (
...
-
 // validateTLSConfig will check that the configured ConfigMap and Secret exist and that they have the correct fields.
-// If TLS is correctly configured, the function will return a result where result.valid = true.
-// If it's incorrectly configured the result will instead have result.valid = false and a message for the user in result.message.
-func (r *ReplicaSetReconciler) validateTLSConfig(mdb mdbv1.MongoDB) (tlsValidationResult, error) {
+// The possible return values are:
+// - (true, nil) if the config is valid
+// - (false, nil) if the config is not valid
+// - (_, err) if an error occured when validating the config
+func (r *ReplicaSetReconciler) validateTLSConfig(mdb mdbv1.MongoDB) (bool, error) {
```
LouisPlisso, line 25: the name and return value make it so clear now :)

fabianlindfors: Agreed :)

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -42,6 +45,35 @@ type MongoDBSpec struct {
...
+
+	// ServerSecretName is the name of a secret containing a private key and certificate to use for TLS
+	// The key and cert are expected to be PEM encoded and available at "tls.key" and "tls.crt"
+	// +optional
+	ServerSecretName string `json:"serverSecretName"`
+
+	// CAConfigMapName is the name of a ConfigMap containing the certificate for the CA which signed the server certificates
+	// The certificate is expected to be available under the key "ca.crt"
+	// +optional
+	CAConfigMapName string `json:"caConfigMapName"`
```
rodrigovalin, line 76: This one is better than the Server, because it implies something related to the CA.

At the same time we have `corev1.ConfigMapKeySelector` and `corev1.SecretKeySelector` that can very nicely support some of our use-cases, like, not making the users use a hard-coded ConfigMap entry name, but any name they like.

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -0,0 +1,142 @@
...
+	// The certificate is used by both mongod and the agent
+	caVolume := statefulset.CreateVolumeFromConfigMap("tls-ca", mdb.Spec.Security.TLS.CAConfigMapName)
+	caVolumeMount := statefulset.CreateVolumeMount(caVolume.Name, tlsCAMountPath, statefulset.WithReadOnly(true))
+
+	// Configure a volume which mounts the secret holding the server key and certificate
+	// The same key-certificate pair is used for all servers
+	tlsSecretVolume := statefulset.CreateVolumeFromSecret("tls-secret", mdb.Spec.Security.TLS.ServerSecretName)
+	tlsSecretVolumeMount := statefulset.CreateVolumeMount(tlsSecretVolume.Name, tlsSecretMountPath, statefulset.WithReadOnly(true))
+
+	// MongoDB expects both key and certificate to be provided in a single PEM file
```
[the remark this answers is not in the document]
fabianlindfors: Planning on implementing that later. Just created a ticket for this: CLOUDP-66615

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -42,6 +45,35 @@ type MongoDBSpec struct {
...
+	Enabled bool `json:"enabled"`
+
+	// Optional configures if TLS should be required or optional for connections
+	// +optional
+	Optional bool `json:"optional"`
+
+	// ServerSecretName is the name of a secret containing a private key and certificate to use for TLS
+	// The key and cert are expected to be PEM encoded and available at "tls.key" and "tls.crt"
+	// +optional
+	ServerSecretName string `json:"serverSecretName"`
```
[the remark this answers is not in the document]
fabianlindfors: I've also been bothered a bit by that. We can discuss this more for the E2E PR which will contain the CRD changes.

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -97,6 +136,14 @@ func (m MongoDB) ConfigMapName() string {
 	return m.Name + "-config"
 }
 
+func (m MongoDB) TLSConfigMapNamespacedName() types.NamespacedName {
```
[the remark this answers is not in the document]
fabianlindfors: True, this isn't entirely obvious. I've added comments for these two functions to clarify this. We should also ensure to make this clear once we document TLS usage properly.

pkg/apis/mongodb/v1/mongodb_types.go
fabianlindfors: Didn't know about those types, thanks! We should definitely look into using those (or something similar) to improve flexibility.

PR #76, opened 2020-06-29
Move ConfigMap/Secret/Service/StatefulSet packages from enterprise to community

chatton:
### All Submissions:

This PR brings over the various packages that act on native K8s types, and defines a small set of interface per type. The utility functions that act on those types now accept these interfaces.


chatton:
@rodrigovalin will double check and include unit tests 👍

pkg/kube/client/client.go
```diff
@@ -19,36 +25,18 @@ func NewClient(c k8sClient.Client) Client {
 
 type Client interface {
```
LouisPlisso, line 26: package client with a Client interface doesn't say anything
not sure it belongs to this PR, but this would need to be changed eventually

chatton: sure, we can decide on something better. Currently this is just an extension of the operator-sdk `client.Client` but we can for sure come up with a better name 👍

LouisPlisso: operatorsdkclient in any form and order would work ;)

PR #78, opened 2020-07-01
CLOUDP-59109: Configure SCRAM

chatton:
### All Submissions:

This PR introduces the capability to configure the MongoDB resources with SCRAM via `spec.security.authentication`

There is no user support, that will be added in a future PR, along with E2E tests.


pkg/util/generate/generate.go
```diff
@@ -0,0 +1,31 @@
...
+		return nil, err
+	}
+
+	return b, nil
+}
+
+func generateRandomString(numBytes int) (string, error) {
+	b, err := generateRandomBytes(numBytes)
+	return base64.StdEncoding.EncodeToString(b), err
+}
```
chatton, line 31: taken from enterprise (we can import this code from community now too)

deploy/crds/mongodb.com_v1_mongodb_scram_cr.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: example-scram-mongodb
+spec:
+  members: 3
+  type: ReplicaSet
+  version: "4.2.6"
+  security:
+    authentication:
+      enabled: true
+      modes: ["SCRAM"]
```
chatton, line 12: CRD changes will be included in E2E test, currently this file isn't actually being used, right now it's just a sample as to what it will look like

deploy/crds/mongodb.com_v1_mongodb_scram_cr.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: MongoDB
+metadata:
+  name: example-scram-mongodb
+spec:
+  members: 3
+  type: ReplicaSet
+  version: "4.2.6"
+  security:
+    authentication:
+      enabled: true
```
chatton, line 11: will be updated to remove this option - will always be true

pkg/controller/mongodb/mongodb_authentication.go
```diff
@@ -0,0 +1,60 @@
...
+const (
+	scramShaOption = "SCRAM"
+)
+
+// noOpAuthEnabler performs no changes, leaving authentication settings untouched
+type noOpAuthEnabler struct{}
+
+func (n noOpAuthEnabler) EnableAuth(auth automationconfig.Auth) automationconfig.Auth {
+	return auth
+}
```
chatton, line 22: this will be removed - we are always going to be enabling auth

pkg/controller/mongodb/mongodb_authentication.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	return noOpAuthEnabler{}, nil
+}
+
+// buildScramPodSpecModification will add the keyfile volume to the podTemplateSpec
+// the keyfile is owned by the agent, and is required to have 0600 permissions.
+func buildScramPodSpecModification(mdb mdbv1.MongoDB) podtemplatespec.Modification {
+	if !mdb.Spec.Security.Authentication.Enabled {
+		return podtemplatespec.NOOP()
+	}
```
chatton, line 47: again, removed as we will always enable auth

pkg/kube/statefulset/statefulset.go
```diff
@@ -92,15 +92,20 @@ func CreateVolumeFromConfigMap(name, sourceName string) corev1.Volume {
 	}
 }
 
-func CreateVolumeFromSecret(name, sourceName string) corev1.Volume {
-	return corev1.Volume{
+func CreateVolumeFromSecret(name, sourceName string, options ...func(v *corev1.Volume)) corev1.Volume {
```
rodrigovalin, line 95: 🥇

pkg/authentication/scram/scram_enabler.go
```diff
@@ -0,0 +1,56 @@
...
+	AgentKeyfileKey                       = "keyfile"
+)
+
+type authEnabler struct {
+	agentPassword string
+	agentKeyFile  string
+}
+
+func (s authEnabler) EnableAuth(auth automationconfig.Auth) automationconfig.Auth {
+	s.enableAgentAuthentication(&auth)
```
rodrigovalin, line 23: these 2 operations seem unbalanced, what do you think if you explicitely pass `s.agentPassword` and `s.agentKeyFile` to `enableAgentAuthentication` in order to make that function (`enableAgentAuthentication`) a regular function:

```go
enableAgentAuthentication(&auth, s.agentPassword, s.agentKeyFile)
enableDeploymentMechanisms(&auth)
return auth
```

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -151,7 +151,13 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	if err := r.ensureAutomationConfig(mdb); err != nil {
+	authEnabler, err := getAuthenticationEnabler(r.client, mdb)
```
rodrigovalin, line 154: to get the authEnabler we need the Kube client and the mdb obect, which is also used everytime we need to pass the authEnabler. Is it posible we make thr `mdb` obect the receiver of `getAuthenticationEnabler` so it can be used and obtained when it is needed and not have to be passed around?

rodrigovalin: You would call `mdb.getAuthenticationEnabler` in `mongodb_controller.go:362` and can avoid passing this object around.

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -333,26 +339,27 @@ func (r ReplicaSetReconciler) updateAndReturnStatusSuccess(mdb *mdbv1.MongoDB) (
...
 		SetTopology(automationconfig.ReplicaSetTopology).
 		SetName(mdb.Name).
 		SetDomain(domain).
 		SetMembers(mdb.Spec.Members).
-		SetPreviousAutomationConfig(currentAC).
+		SetPreviousAutomationConfig(currentAc).
 		SetMongoDBVersion(mdb.Spec.Version).
 		SetFCV(mdb.GetFCV()).
 		AddVersion(mdbVersionConfig).
+		SetAuthEnabler(enabler).
```
rodrigovalin, line 362: Related to previous comment, this could be changed for mdb.getAuthEnabler(), instead of passing `enabler` everywhere.

deploy/crds/mongodb.com_v1_mongodb_scram_cr.yaml
chatton: what we ended up deciding on was that it will not be possible to disable auth. So the enabled field will be removed alltogether. (currently Authentication is a struct so it can never be nill)

I can remove it in this PR to clear things up

pkg/controller/mongodb/mongodb_controller.go
chatton: yeah I was considering this before, I think it will end up being a lot cleaner, will play around some more with this idea!

chatton: @rodrigovalin one argument against making the mdb the receiver of this function is that it will no longer purely be responsible for the Kubernetes view of the resource. It will now (inderectly) know how to configure the AutomationConfig. I agree that we should move the getEnabler code into the function that builds the AutomationConfig, but I still think it would be better to have both the client and mdb as arguments to this function. As it is the operator's responsibility to bridge the 2 domains (AC + K8s). Let me know what you think!

rodrigovalin: I agree with that, and I think I don't have a better idea on how to fix the added `enabler` being passed everywhere :(

chatton: I think we can still create the enabler within the function it is used to minimise its scope. And not need to add to all of the method signatures

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -428,32 +429,38 @@ func buildService(mdb mdbv1.MongoDB) corev1.Service {
...
 	manifest, err := r.manifestProvider()
 	if err != nil {
 		return corev1.ConfigMap{}, fmt.Errorf("error reading version manifest from disk: %+v", err)
 	}
 
+	enabler, err := getAuthenticationEnabler(r.client, mdb)
+	if err != nil {
+		return corev1.ConfigMap{}, err
+	}
+
```
chatton, line 457: @rodrigovalin  we can construct the enabler inside this function, as we don't actually need it anywhere else.

deploy/crds/mongodb.com_v1_mongodb_scram_cr.yaml
rodrigovalin: no need to remove it. it makes sense

PR #80, opened 2020-07-01
Handle graceful shutdown properly

fabianlindfors:
### All Submissions:

When Kubernetes shuts down a pod it will send a SIGTERM signal to every container and give the process 30 seconds to finish up. The way our mongod container is currently set up, the signal is lost and the `mongod` process is not shutdown gracefully. If that pod is currently the primary this will cause a failure. Since 4.0.8, MongoDB is able to automatically perform a step-down when a primary receives SIGTERM: https://docs.mongodb.com/manual/tutorial/manage-mongodb-processes/#sigterm-and-replica-sets.

Our current setup could cause issues when performing a scale down, deleting a pod, or when performing a rolling update, for example to enable TLS. As we are using a shell script to run `mongod`  we need to trap the signal and forward it to the main process. This PR implements the SIGTERM trapping in bash.

Right now most of our E2E tests are running on version 4.0.6 which doesn't support graceful shutdowns. This problem could explain some flakiness we have experienced with "e2e_test_replica_set_readiness_probe" which deletes a random pod. Because of this, I've also updated all tests to run on 4.2.

### DIFF
--- a/pkg/controller/mongodb/mongodb_controller.go
+++ b/pkg/controller/mongodb/mongodb_controller.go
@@ -435,12 +435,26 @@ func preStopHookInit(volumeMount []corev1.VolumeMount) container.Modification {
 
 func mongodbContainer(version string, volumeMounts []corev1.VolumeMount) container.Modification {
 	mongoDbCommand := []string{
-		"/bin/sh",
+		"/bin/bash",
 		"-c",
 		// we execute the pre-stop hook once the mongod has been gracefully shut down by the agent.
 		`while [ ! -f /data/automation-mongod.conf ]; do sleep 3 ; done ; sleep 2 ;
+
+# terminate_mongod will send a SIGTERM signal to the mongod process
+terminate_mongod() {
+  kill -TERM "$mongo_pid" 2>/dev/null ;
+}
+
+# catch any SIGTERM signals and forward them to mongod
+# SIGTERM will be sent by Kubernetes during a graceful shutdown
+trap terminate_mongod SIGTERM ;
+
 # start mongod with this configuration
-mongod -f /data/automation-mongod.conf ;
+mongod -f /data/automation-mongod.conf &
+
+# capture the mongod PID for SIGTERM forwarding and wait for the process to finish
+mongo_pid=$! ;
+wait "$mongo_pid" ;
 
 # start the pre-stop-hook to restart the Pod when needed
 # If the Pod does not require to be restarted, the pre-stop-hook will
--- a/test/e2e/e2eutil.go
+++ b/test/e2e/e2eutil.go
@@ -128,7 +128,7 @@ func NewTestMongoDB(name string) mdbv1.MongoDB {
 		Spec: mdbv1.MongoDBSpec{
 			Members:                     3,
 			Type:                        "ReplicaSet",
-			Version:                     "4.0.6",
+			Version:                     "4.2.6",
 			FeatureCompatibilityVersion: "4.0",
 		},
 	}
--- a/test/e2e/feature_compatibility_version/feature_compatibility_version_test.go
+++ b/test/e2e/feature_compatibility_version/feature_compatibility_version_test.go
@@ -27,20 +27,20 @@ func TestFeatureCompatibilityVersion(t *testing.T) {
 	t.Run("Basic tests", mongodbtests.BasicFunctionality(&mdb))
 
 	t.Run("Test FeatureCompatibilityVersion is 4.0", mongodbtests.HasFeatureCompatibilityVersion(&mdb, "4.0", 3))
-	// Upgrade version to 4.2.6 while keeping the FCV set to 4.0
+	// Upgrade version to 4.2.8 while keeping the FCV set to 4.0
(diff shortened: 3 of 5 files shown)


pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -435,12 +435,26 @@ func preStopHookInit(volumeMount []corev1.VolumeMount) container.Modification {
 
 func mongodbContainer(version string, volumeMounts []corev1.VolumeMount) container.Modification {
 	mongoDbCommand := []string{
-		"/bin/sh",
+		"/bin/bash",
```
LouisPlisso, line 438: seems like just changing to bash may solve the issue as it seems that bash will forward the sigterm to its children: see option 2 here
https://pracucci.com/graceful-shutdown-of-kubernetes-pods.html#:~:text=When%20a%20pod%20should%20be,the%20running%20application%20and%20exit.

fabianlindfors: I actually read that article and tried using just bash before but unfortunately that didn't work. I've researched this a bit but haven't been able to come to a conclusion on how bash actually handles signals. Here's somebody claiming bash will ignore any signals when a child process is executing: https://unix.stackexchange.com/a/146770. This article suggests one needs to use `exec`: https://itnext.io/containers-terminating-with-grace-d19e0ce34290 which unfortunately we can't because of our pre-stop hook.

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -435,12 +435,26 @@ func preStopHookInit(volumeMount []corev1.VolumeMount) container.Modification {
...
 		"-c",
 		// we execute the pre-stop hook once the mongod has been gracefully shut down by the agent.
 		`while [ ! -f /data/automation-mongod.conf ]; do sleep 3 ; done ; sleep 2 ;
+
+# terminate_mongod will send a SIGTERM signal to the mongod process
+terminate_mongod() {
+  kill -TERM "$mongo_pid" 2>/dev/null ;
```
antonlisovenko, line 445: @fabianlindfors have you checked if the `kill` will wait for mongodb process to end? 
I don't remember the details and I could be wrong but while working on trap in mongodb enterprise I had to wait for the process to shutdown:
https://github.com/10gen/ops-manager-kubernetes/blob/6491bbfefc5e4f147198a7464bff76a207e389f2/docker/mongodb-enterprise-database/content/agent-launcher-lib.sh#L41

fabianlindfors: You're right! I thought it was working but the process wasn't waiting but was running the pre-stop hook immediately. The pre-stop hook stayed alive long enough for the server to shut down gracefully but that's a horrible way of doing it.

I've updated the script to use a `wait` inside of of `terminate_mongod` which seems to work in our case. I also added an `exit` after that to stop the pre-stop hook from running. It just slows down the shut down procedure without doing anything.

fabianlindfors:
To stay consistent with Kubernetes, the optimal way would be to use the actual lifecycle hooks instead of putting our pre-stop hook in a shell script. This would allow us to use `exec` in our command to give `mongod` PID 1 and avoid the trap.

@rodrigovalin What's the reason we opted to put the pre-stop hook in the script rather than as a lifecycle hook?

rodrigovalin:
I'm not really sure what was the reason we decided to use this bash script instead of Pod lifecycle, I think this is because the preStop hook is not called under every circumstance.

1. One thing we can try is to have a initContainer that just do `while [ ! -f /data/automation-mongod.conf ]; do sleep 3 ; done`
2. Have mongod running as its own container (PID1)
3. Find a proper way of calling the stop hook with: sidecard, pod lifecycle (needs more investigation)

fabianlindfors:
I created a ticket to track this investigation: https://jira.mongodb.org/browse/CLOUDP-66661

chatton:
@fabianlindfors @rodrigovalin  we initially attempted to use a native pre-stop hook, but this only gets executed on a non-graceful shutdown, and so when the agents step down the mongod it doesn't get called. As it currently stands, the name is misleading.

One other possible suggestion, if we do need to grow this bash script, maybe we could consider creating a go binary to handle this instead. I'm not sure how nice this would look, but I think it would be preferable to bash.

LouisPlisso:
@chatton @rodrigovalin maybe we had our prestop hook wrongly configured
seem like we have to pay attention to sleep intervals:
https://github.com/kubernetes/kubernetes/issues/39170#issuecomment-451404411

LouisPlisso:
@chatton also I agree that we should stay with bash: we are overriding the entrypoint so we want users to be able to see what exactly we are doing

chatton:
@LouisPlisso I'm happy to stay with bash, but if the main benefit is the visibility into what's happening, all the code will still be visible, we won't be hiding anything.

fabianlindfors:
We have decided to go with a sidecar solution instead. We'll remove our current pre-stop hook and make `mongod` the main process of the container. To handle version changes, we'll have a sidecar container which continuously watches the agent status and restarts the pod when necessary.

PR #81, opened 2020-07-02
CLOUDP-66230: Add E2E tests for TLS

fabianlindfors:
### All Submissions:

This PR adds two E2E tests for TLS, they are:

- `replica_set_tls`: A test for creating a new TLS-enabled cluster including a connectivity test over TLS.
- `replica_set_tls_upgrade`: A test which enables TLS on an existing cluster and ensures the cluster is reachable during the upgrade.

The test certificates and keys are pre-generated and stored in `testdata/tls`. This PR doesn't contain the CRD updates as it was already getting a bit big. Those changes will be a separate PR.


test/e2e/mongodbtests/mongodbtests_tls.go
```diff
@@ -0,0 +1,180 @@
...
+		Decode(&result)
+
+	value := result.Map()[key]
+	return value, nil
+}
+
+func getClientTLSConfig() *tls.Config {
+	// Read the CA certificate from test data
+	caPool := x509.NewCertPool()
+	caPEM, _ := ioutil.ReadFile("testdata/tls/ca.crt")
```
chatton, line 174: it's a test function where we know we created the file, but I think we should aim to always check the errors. If for some reason the file ended up being renamed, or moved, it might be tricky to find that this was the place where the problem was

test/e2e/replica_set_tls/replica_set_tls_test.go
```diff
@@ -0,0 +1,47 @@
...
+func TestReplicaSetTLS(t *testing.T) {
+	ctx, shouldCleanup := setup.InitTest(t)
+	if shouldCleanup {
+		defer ctx.Cleanup()
+	}
+
+	mdb := e2eutil.NewTestMongoDB("mdb0")
+	mdb.Spec.Security.TLS = e2eutil.NewTestTLSConfig(false)
+
+	t.Run("Create TLS Resources", mongodbtests.CreateTLSResources(&mdb, ctx))
```
chatton, line 28: I would vote for a `
```golang
if err := setup.CreateTlsResources(mdb.Namespace); err != nil {
    t.Fatal("Failed setup...")
}
```

test/e2e/replica_set_tls/replica_set_tls_test.go
```diff
@@ -0,0 +1,47 @@
...
+	t.Run("Create MongoDB Resource", mongodbtests.CreateMongoDBResource(&mdb, ctx))
+	t.Run("Config Map Was Correctly Created", mongodbtests.AutomationConfigConfigMapExists(&mdb))
+	t.Run("Stateful Set Reaches Ready State", mongodbtests.StatefulSetIsReady(&mdb))
+	t.Run("Stateful Set has OwnerReference", mongodbtests.StatefulSetHasOwnerReference(&mdb,
+		*metav1.NewControllerRef(&mdb, schema.GroupVersionKind{
+			Group:   mdbv1.SchemeGroupVersion.Group,
+			Version: mdbv1.SchemeGroupVersion.Version,
+			Kind:    mdb.Kind,
+		})))
```
chatton, line 37: maybe we should extract this into a `defaultOwnerReference` function or similar. I think we have this call to `metav1.NewControllerRef` in several places now.

test/e2e/mongodbtests/mongodbtests_tls.go
```diff
@@ -0,0 +1,180 @@
...
+
+		configMap := corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      tlsConfig.CAConfigMapName,
+				Namespace: mdb.Namespace,
+			},
+			Data: map[string]string{
+				"ca.crt": string(ca),
+			},
+		}
```
chatton, line 43: remember we have builders we can use for constructing literals

```golang
configMap := configmap.Builder().
    SetName(tlsConfig.CAConfigMapName).
    SetNamespace(mdb.Namespace).
    SetField("ca.crt", string(ca)).
    Build()
```

test/e2e/mongodbtests/mongodbtests_tls.go
```diff
@@ -0,0 +1,180 @@
...
+		secret := corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      tlsConfig.ServerSecretName,
+				Namespace: mdb.Namespace,
+			},
+			Data: map[string][]byte{
+				"tls.crt": cert,
+				"tls.key": key,
+			},
+		}
```
chatton, line 61: same comment as above, `secret.Builder`

test/e2e/replica_set_tls/replica_set_tls_test.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func TestReplicaSetTLS(t *testing.T) {
+	ctx, shouldCleanup := setup.InitTest(t)
+	if shouldCleanup {
+		defer ctx.Cleanup()
+	}
+
+	mdb := e2eutil.NewTestMongoDB("mdb0")
+	mdb.Spec.Security.TLS = e2eutil.NewTestTLSConfig(false)
```
chatton, line 26: I feel like we're going to need a better way of building MongoDB resources, perhaps a builder or even for now just a separate function to fully produce a TLS enabled MDB resource. I would also suggest maybe making the resource name a bit more meaningful, even just "mdb-tls" or "mdb-tls-upgrade"

test/e2e/mongodbtests/mongodbtests_tls.go
```diff
@@ -0,0 +1,180 @@
...
+		err = f.Global.Client.Create(context.TODO(), &secret, &f.CleanupOptions{TestContext: ctx})
+	}
+}
```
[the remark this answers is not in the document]
fabianlindfors: Agreed! Moved it now.

I also agree on your nit. Changed it to be a parameter and I think it looks pretty good:
```
mdb := e2eutil.NewTestMongoDB("mdb0")
if err := setup.CreateTLSResources(mdb.Namespace, ctx); err != nil {
	t.Fatalf("Failed to set up TLS resources: %+v", err)
}
```

test/e2e/mongodbtests/mongodbtests_tls.go
```diff
@@ -0,0 +1,180 @@
+package mongodbtests
```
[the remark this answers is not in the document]
fabianlindfors: I think that's a good idea, moved.

The packages are getting kind of hard to tell apart now though. As you mentioned in your comment, we should figure out a better way of structuring tests, including how we structure the packages.

test/e2e/mongodbtests/mongodbtests_tls.go
fabianlindfors: That's a good point. Added some error handling!

test/e2e/mongodbtests/mongodbtests_tls.go
fabianlindfors: Totally forgot about those, thank you!

test/e2e/mongodbtests/mongodbtests_tls.go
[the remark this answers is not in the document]
chatton: definitely! It was okay when there were only a few tests, but we want to make sure it's scalable as we add more and more

test/e2e/replica_set_tls/replica_set_tls_test.go
fabianlindfors: Fixed this by refactoring all tests to use the `BasicFunctionality` test. Now this owner reference only appears there.

test/e2e/replica_set_tls/replica_set_tls_test.go
fabianlindfors: Have updated the tests to use the name "mdb-tls" instead. As the name is embedded in the certificate I've opted to use the same for both, otherwise we would need one certificate for each test.

We also agreed that a builder would be a nice addition. That work is tracked by a different ticket.

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -260,19 +260,25 @@ func Connect(mdb *mdbv1.MongoDB) error {
...
 func IsReachableDuring(mdb *mdbv1.MongoDB, interval time.Duration, testFunc func()) func(*testing.T) {
+	return IsReachableDuringWithConnection(mdb, interval, testFunc, func() error {
+		return Connect(mdb, options.Client())
+	})
+}
+
+func IsReachableDuringWithConnection(mdb *mdbv1.MongoDB, interval time.Duration, testFunc func(), connectFunc func() error) func(*testing.T) {
 	return func(t *testing.T) {
 		ctx, cancel := context.WithCancel(context.Background()) // start a go routine which will periodically check basic MongoDB connectivity
+		defer cancel()
```
chatton, line 271: 👍

fabianlindfors:
@rodrigovalin I believe the test failure was caused by what I fixed in "Move context cancel outside goroutine". I'm going to run it a few times now to see if I can trigger another failure, otherwise I'll go ahead and merge.

PR #82, opened 2020-07-02
E2E test fixes

bznein:
## This PR

- Fixes the default names of arguments to the `e2e` script (they were the opposite of they intended use)
- Fixes a wrong indentation level that made one of the arguments dependent from another
- Fixes the check for readiness of the statefulset
- Fixes the choice of a random pod to get deleted in the test of the readiness probe: we were not seeding the `rand` and we were obtaining a random number in the range `[0,n-1)`, effectively never selecting the last pod as the one to get deleted

### All Submissions:

### DIFF
--- a/scripts/dev/e2e.py
+++ b/scripts/dev/e2e.py
@@ -318,14 +318,11 @@ def main():
     try:
         build_and_push_images(args, dev_config)
         prepare_and_run_testrunner(args, dev_config)
-        test_runner_pod=k8s_request_data.get_pod_namespaced(dev_config.namespace,TEST_RUNNER_NAME)
     finally:
         if not args.skip_dump_diagnostic:
             dump_diagnostic.dump_all(dev_config.namespace)
 
-    print(test_runner_pod.status.phase)
-    time.sleep(20)
-    print(test_runner_pod.status.phase)
+    test_runner_pod=k8s_request_data.get_pod_namespaced(dev_config.namespace,TEST_RUNNER_NAME)
     if test_runner_pod.status.phase != "Succeeded":
         sys.exit(1)
 
--- a/scripts/dev/e2e.py
+++ b/scripts/dev/e2e.py
@@ -228,12 +228,12 @@ def parse_args():
     parser = argparse.ArgumentParser()
     parser.add_argument("--test", help="Name of the test to run")
     parser.add_argument(
-        "--skip-operator-install",
+        "--install_operator",
         help="Do not install the Operator, assumes one is installed already",
         action="store_false",
     )
     parser.add_argument(
-        "--skip-image-build", help="Skip building images", action="store_false",
+        "--build_images", help="Skip building images", action="store_false",
     )
     parser.add_argument(
         "--tag",
@@ -242,12 +242,12 @@ def parse_args():
         default="latest",
     )
     parser.add_argument(
-        "--dump_diagnostic",
+        "--skip_dump_diagnostic",
         help="Dump diagnostic information into files",
         action="store_false",
     )
     parser.add_argument(
-        "--skip-cleanup",
+        "--perform-cleanup",
         help="skip the context cleanup when the test ends",
         action="store_false",
     )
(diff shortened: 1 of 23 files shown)


scripts/dev/e2e.py
```diff
@@ -257,31 +257,31 @@ def parse_args():
 
 def build_and_push_images(args, dev_config):
     test_runner_name = dev_config.testrunner_image
-    if not args.skip_operator_install:
+    if not args.install_operator:
```
chatton, line 260: yeah it's much more readable keeping it in the positive form

test/e2e/e2eutil.go
```diff
@@ -77,7 +78,7 @@ func WaitForStatefulSetToHaveUpdateStrategy(t *testing.T, mdb *mdbv1.MongoDB, st
 // have reached the ready status
 func WaitForStatefulSetToBeReady(t *testing.T, mdb *mdbv1.MongoDB, retryInterval, timeout time.Duration) error {
 	return waitForStatefulSetCondition(t, mdb, retryInterval, timeout, func(sts appsv1.StatefulSet) bool {
-		return sts.Status.ReadyReplicas == int32(mdb.Spec.Members)
+		return statefulset.IsReady(sts, mdb.Spec.Members)
```
chatton, line 81: 👍

test/e2e/replica_set_readiness_probe/replica_set_readiness_probe_test.go
```diff
@@ -18,6 +18,7 @@ func TestMain(m *testing.M) {
 
 func TestReplicaSetReadinessProbeScaling(t *testing.T) {
 
+	rand.Seed(time.Now().Unix())
```
chatton, line 21: nice, one small thing I would do is to put this in an `init` function

```golang
init(){
    rand.Seed(time.Now().Unix())
}
```

You don't need to explicitly call `init` functions

test/e2e/replica_set_readiness_probe/replica_set_readiness_probe_test.go
```diff
@@ -31,7 +32,7 @@ func TestReplicaSetReadinessProbeScaling(t *testing.T) {
 	t.Run("Stateful Set Reaches Ready State", mongodbtests.StatefulSetIsReady(&mdb))
 	t.Run("MongoDB is reachable", mongodbtests.IsReachableDuring(&mdb, time.Second*10,
 		func() {
-			t.Run("Delete Random Pod", mongodbtests.DeletePod(&mdb, rand.Intn(mdb.Spec.Members-1)))
+			t.Run("Delete Random Pod", mongodbtests.DeletePod(&mdb, rand.Intn(mdb.Spec.Members)))
```
chatton, line 35: from the docs

`Intn returns, as an int, a non-negative pseudo-random number in [0,n) from the default Source. It panics if n <= 0.`

That means we could access `pod-3` which doesn't exist!

bznein: Maybe I am missing something, but if `n=3`, `IntN` will return a number in the range `[0,3)`, so 3 will be excluded (notice the round parentheses as opposed to the square one).

Or am I missing something?

scripts/dev/e2e.py
```diff
@@ -228,12 +228,12 @@ def parse_args():
     parser = argparse.ArgumentParser()
     parser.add_argument("--test", help="Name of the test to run")
     parser.add_argument(
-        "--skip-operator-install",
+        "--install_operator",
```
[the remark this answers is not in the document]
bznein: you are right, I wanted to use a dash as well but missed that

scripts/dev/e2e.py
bznein: It was actually misleading as we were installing the operator if it was specified to NOT install it. I changed everything to be totally in the positive form now

test/e2e/replica_set_readiness_probe/replica_set_readiness_probe_test.go
chatton: nope you are absolutely right! I missed it 👍

test/e2e/replica_set_readiness_probe/replica_set_readiness_probe_test.go
LouisPlisso: I think init functions are making the code harder to follow as it's very hard to understand when they are executed

see side effect setup in https://medium.com/capital-one-tech/doing-well-by-doing-bad-writing-bad-code-with-go-part-1-2dbb96ce079a

test/e2e/replica_set_readiness_probe/replica_set_readiness_probe_test.go
```diff
@@ -16,6 +16,10 @@ func TestMain(m *testing.M) {
 	f.MainEntry(m)
 }
 
+func init() {
```
LouisPlisso, line 19: let's try to avoid init functions when possible

test/e2e/replica_set_readiness_probe/replica_set_readiness_probe_test.go
LouisPlisso: @bznein ^

test/e2e/replica_set_readiness_probe/replica_set_readiness_probe_test.go
bznein: Will do!

PR #83, opened 2020-07-02
Fixed OnDelete strategy

chatton:
### All Submissions:

Our strategy for changing the StatefulSet update strategy to OnDelete was not working as intended.

Our isReady check was not working, as once we set the strategy to OnDelete, this no longer triggers a rolling upgrade. We need a mechanism to track if the StatefulSet became un-ready while being set to OnDelete. I used annotations for this purpose.

### DIFF
--- a/pkg/controller/mongodb/mongodb_controller.go
+++ b/pkg/controller/mongodb/mongodb_controller.go
@@ -57,6 +57,8 @@ const (
 	clusterFilePath                = "/var/lib/automation/config/automation-config"
 	operatorServiceAccountName     = "mongodb-kubernetes-operator"
 	agentHealthStatusFilePathValue = "/var/log/mongodb-mms-automation/healthstatus/agent-health-status.json"
+
+	hasLeftReadyState = "mongodb.com/v1.hasLeftReadyState"
 )
 
 // Add creates a new MongoDB Controller and adds it to the Manager. The Manager will set fields on the Controller
@@ -150,6 +152,7 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		r.log.Warnf("Error creating/updating StatefulSet: %+v", err)
 		return reconcile.Result{}, err
 	}
+	//time.Sleep(10 * time.Second)
 
 	currentSts := appsv1.StatefulSet{}
 	if err := r.client.Get(context.TODO(), mdb.NamespacedName(), &currentSts); err != nil {
@@ -176,8 +179,13 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 	}
 
 	r.log.Debug("Setting MongoDB Annotations")
-	if err := r.setAnnotation(types.NamespacedName{Name: mdb.Name, Namespace: mdb.Namespace}, mdbv1.LastVersionAnnotationKey, mdb.Spec.Version); err != nil {
-		r.log.Warnf("Error setting annotation: %+v", err)
+
+	annotations := map[string]string{
+		mdbv1.LastVersionAnnotationKey: mdb.Spec.Version,
+		hasLeftReadyState:              "false",
+	}
+	if err := r.setAnnotations(mdb.NamespacedName(), annotations); err != nil {
+		r.log.Warnf("Error setting annotations: %+v", err)
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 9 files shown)


pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -150,6 +152,7 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		r.log.Warnf("Error creating/updating StatefulSet: %+v", err)
 		return reconcile.Result{}, err
 	}
+	//time.Sleep(10 * time.Second)
```
bznein, line 155: Leftover comment :)

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -220,10 +228,28 @@ func (r *ReplicaSetReconciler) isStatefulSetReady(mdb mdbv1.MongoDB, existingSta
 		return false, err
 	}
 
-	// comparison is done with bytes instead of reflect.DeepEqual as there are
-	// some issues with nil/empty maps not being compared correctly otherwise
+	//comparison is done with bytes instead of reflect.DeepEqual as there are
+	//some issues with nil/empty maps not being compared correctly otherwise
 	areEqual := bytes.Compare(stsCopyBytes, stsBytes) == 0
```
bznein, line 233: While setting up the linter, I see it complaining about this, we should use

`bytes.Equal(stsCopyBytes, stsBytes)`

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -251,15 +277,17 @@ func (r *ReplicaSetReconciler) createOrUpdateStatefulSet(mdb mdbv1.MongoDB) erro
 	return nil
 }
 
-// setAnnotation updates the monogdb resource with the given namespaced name and sets the annotation
+// setAnnotations updates the monogdb resource with the given namespaced name and sets the annotation
 // "key" with the provided value "val"
-func (r ReplicaSetReconciler) setAnnotation(nsName types.NamespacedName, key, val string) error {
+func (r ReplicaSetReconciler) setAnnotations(nsName types.NamespacedName, annotations map[string]string) error {
```
fabianlindfors, line 282: As we add more annotations I feel like this function is going to get a bit hard to use. To avoid overwriting existing annotations we'll probably have to always update `mdb.Annotations` directly and then pass it to the function. If we do this we'll also need a way to deal with the case when `Annotations` is nil.

One idea is to introduce a new function on the MongoDB type:
```
func (m *MongoDB) SetAnnotation(key, value, string) {
  if m.Annotations == nil {
    m.Annotations = map[string]string{}
  }

  m.Annotation[key] = value
}
```

And then change `setAnnotations` to `ReplicaSetReconciler.saveAnnotations(mdb mdbv1.MongoDB)` to persist the changes. This way we don't have to worry about overwriting. Thoughts?

pkg/controller/mongodb/mongodb_controller.go
fabianlindfors, line 155: Will this be needed at a later time?

test/e2e/e2eutil.go
```diff
@@ -77,15 +79,8 @@ func WaitForStatefulSetToHaveUpdateStrategy(t *testing.T, mdb *mdbv1.MongoDB, st
...
-	})
-}
-
-// waitForStatefulSetToBeUpdated waits until all replicas of the StatefulSet with the given name
-// are updated.
-func WaitForStatefulSetToBeUpdated(t *testing.T, mdb *mdbv1.MongoDB, retryInterval, timeout time.Duration) error {
-	return waitForStatefulSetCondition(t, mdb, retryInterval, timeout, func(sts appsv1.StatefulSet) bool {
-		return sts.Status.UpdatedReplicas == int32(mdb.Spec.Members)
+		return statefulset.IsReady(sts, mdb.Spec.Members)
+		//return sts.Status.ReadyReplicas == int32(mdb.Spec.Members)
```
fabianlindfors, line 83: Same here, do we need to keep this code?

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -251,15 +277,17 @@ func (r *ReplicaSetReconciler) createOrUpdateStatefulSet(mdb mdbv1.MongoDB) erro
 	return nil
 }
 
-// setAnnotation updates the monogdb resource with the given namespaced name and sets the annotation
+// setAnnotations updates the monogdb resource with the given namespaced name and sets the annotation
 // "key" with the provided value "val"
```
bznein, line 281: Maybe we can rewrite also the rest of the comment? "key" and "val" mean nothing now

pkg/controller/mongodb/replicaset_controller_test.go
```diff
@@ -127,6 +130,26 @@ func TestChangingVersion_ResultsInRollingUpdateStrategyType(t *testing.T) {
 
 	_ = mgrClient.Update(context.TODO(), &mdb)
 
+	// agents start the upgrade, they are not all ready
+	sts.Status.UpdatedReplicas = 1
+	sts.Status.ReadyReplicas = 2
+	err = mgrClient.Update(context.TODO(), &sts)
+
+	_ = mgrClient.Get(context.TODO(), types.NamespacedName{Name: mdb.Name, Namespace: mdb.Namespace}, &sts)
```
bznein, line 138: Can we use (here and in the rest of the file) `mdb.NamespacedName()?`

test/e2e/e2eutil.go
bznein, line 83: Leftover comment :)

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -57,6 +57,8 @@ const (
 	clusterFilePath                = "/var/lib/automation/config/automation-config"
 	operatorServiceAccountName     = "mongodb-kubernetes-operator"
 	agentHealthStatusFilePathValue = "/var/log/mongodb-mms-automation/healthstatus/agent-health-status.json"
+
+	hasLeftReadyState = "mongodb.com/v1.hasLeftReadyState"
```
fabianlindfors, line 61: We store the other annotation keys in `mongodb_types.go` and should stay consistent about this and their naming, e.g. `hasLeftReadyStateAnnotationKey`. In my eyes they do make more sense in the controller file though so perhaps we should move `LastVersionAnnotationKey` over!

pkg/controller/mongodb/mongodb_controller.go
bznein: (Of course it's not related to this PR so feel free to ignore and I'll do it on my linter PR!)

pkg/controller/mongodb/mongodb_controller.go
chatton: sure won't, removing 👍

pkg/controller/mongodb/mongodb_controller.go
chatton: I was thinking the same thing, I can move them over here.

pkg/controller/mongodb/mongodb_controller.go
chatton: I think this is a good point, maybe we can do a smaller follow up PR to avoid clutter in this one.

fabianlindfors: Good :)

PR #84, opened 2020-07-02
CLOUDP-65826: Add automatic linting to the codebase

bznein:
### All Submissions:


bznein:
This PR adds automatic linting to our codease through GitHub's super-linter, a GitHub action that runs on every pull request and reports the error found beside the evergreen tasks. It is at the moment configured only for YAML, Python and Bash. The support for GO is at the moment broken, so we will not enable it until the PR I opened on super-linter's repo gets merged. 

A couple of changes to the bash script have also been made as well to avoid bash linting failures.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,72 @@
...
+      #         sudo chmod 777 /usr/local/bin/operator-sdk
+      # - name: Generate DeepCopy
+      #   Run: operator-sdk generate k8s
+
+      - name: Lint Code Base
+        uses: docker://github/super-linter:v3
+        env:
+          VALIDATE_ALL_CODEBASE: true
+          # Now we set the PYTHONPATH to the path of the dependencies *inside* the container
+          PYTHONPATH: "/github/workspace/:\
```
rodrigovalin, line 66: this seems to be very convoluted for a github provided functionality :( is there a reason why the python linting requires so much setup work?

bznein: Unfortunately it seems that unless we want to ignore all the `missing-import` errors this is the only way at the moment:h ttps://github.com/github/super-linter/issues/157 

I guess this is the price we pay for adopting something that has been officially released just two weeks ago :(

.github/workflows/main.yaml
```diff
@@ -0,0 +1,72 @@
...
+# Documentation:
+# https://help.github.com/en/articles/workflow-syntax-for-github-actions
+#
+
+#############################
+# Start the job on all push #
+#############################
+on:
+  pull_request:
+    branches: [master]
```
LouisPlisso, line 19: @bznein can we rename our branch to main? 🙏

bznein: I think we need to wait for github to change the terminology to main? 

Or I think there's a way to change ourself the name of the branch to main, but in any case that should happen before we can change it here, otherwise it will never trigger as it won't see any branch named `main`

bznein: @LouisPlisso it seems rather straightforward to rename it to main: https://dev.to/rhymu8354/git-renaming-the-master-branch-137b

PR #94, opened 2020-07-08
E2e timeout and status extraction

bznein:
Since moving our e2e logic to python we had some intermittent failures with pods not reaching running states in time or updating their status fast enough for us to catch it. This PR aims at fixing these issues by increasing the timeout for these checks and by changing the way we check for the `Succeeded` state (polling it regularly until in times out or reaches the state)

We also have very seldom failures due to the API token for the service account not existing yet (https://evergreen.mongodb.com/task/mongodb_kubernetes_operator_e2e_tests_e2e_test_feature_compatibility_version_9fcae382ccc68af0b4e56c691bf061195c7821d9_20_07_07_15_24_19/0) which is usually fixed just with a little wait before retrying. 
I wasn't able to test this particular fix as it is really a rare occurrence, but the 10-seconds sleep should make it work.

On my sample size of roughly 20 patches ran, I haven't seen a single failure, while previously I believe we had at least one or two failing task every couple of patches, so I am hopeful this actually fixed (most of) the issues

### All Submissions:


scripts/dev/e2e.py
```diff
@@ -156,14 +156,28 @@ def create_test_runner_pod(
...
+                "API token was not found for service account, retrying in 10 seconds..."
+            )
+            time.sleep(10)
+            try:
+                corev1.create_namespaced_pod(dev_config.namespace, body=pod_body)
+            except ApiException as e:
+                print(
+                    "The API token was not found for two consecutive tries, aborting the test"
+                )
+                raise
```
chatton, line 180: do you think we need to special case this specific error? This is generally just a case of the cluster not being ready for requests, maybe it would be simpler to just retry any ApiException a few times with a pause in between. (It's also not that obvious why we want to retry on this specific error twice, and not others). What do you think?

bznein: Yeah I guess it makes sense. I didn't want to retry on exceptions that are caused by different issues but it's true that at the worst case they will happen again on the next try. I'll change it

scripts/dev/k8s_conditions.py
```diff
@@ -45,24 +47,51 @@ def wait(
...
-def _ignore_error_codes(fn, codes):
+def call_eventually_succeeds(
+    fn: Callable,
+    sleep_time: float = SLEEP_TIME,
+    timeout: int = INFINITY,
+    exceptions_to_ignore: Optional[Tuple[BaseException]] = None,
+) -> bool:
+    """
+    call_eventually_succeeds is similar to wait but for when we don't care about the returned value of fn
+    (it can even not return anything).
```
chatton, line 58: it looks like we never return anything, maybe we can update the comment to say the return value isn't examined at all

scripts/git-hooks/pre-commit
```diff
@@ -28,7 +28,7 @@ function mypy_check()
         echo "Analyzing $file ..."
         # We ignore missing import otherwise mypy will complain
         # about 3rd party libraries not providing type hints
-        if ! mypy --ignore-missing-imports "${file}"; then
+        if ! mypy --disallow-untyped-calls --disallow-untyped-defs --disallow-incomplete-defs --ignore-missing-imports "${file}"; then
```
chatton, line 31: 👍

scripts/dev/e2e.py
```diff
@@ -78,17 +78,19 @@ def _prepare_testrunner_environment(config_file: str):
...
     """Replicates the local kubeconfig file (pointed at by KUBECONFIG),
     as a ConfigMap."""
     corev1 = client.CoreV1Api()
     print("Creating kube-config ConfigMap")
 
     svc = corev1.read_namespaced_service("kubernetes", "default")
-    kube_config = os.getenv("KUBECONFIG")
-    with open(kube_config) as fd:
+    kube_config_path = os.getenv("KUBECONFIG")
+    assert kube_config_path is not None
```
chatton, line 89: rather than an assertion, I think it would be better to explicitly throw an exception (not `AssertionError`). We should reserve assert statements for test code. In this case there was an error setting up our environment, rather than a test failure.

maybe

```python
if kube_config_path is None:
   raise ValueError("kube_config_path must not be None")
```

scripts/dev/e2e.py
```diff
@@ -78,17 +78,19 @@ def _prepare_testrunner_environment(config_file: str):
...
+    with open(kube_config_path) as fd:
         kube_config = yaml.safe_load(fd.read())
 
+    assert kube_config is not None
```
chatton, line 93: ^^ same as above

scripts/dev/k8s_conditions.py
bznein: Yes poor phrasing in the comment!

PR #99, opened 2020-07-10
CLOUDP-66799: Configure SCRAM users

chatton:
### All Submissions:

This PR adds the following:

* Scram Credentials are generated and stored in a secret based on a password that is provided in the form of a secret
* Scram is now always on.
* Automation Config is now stored in a Secret instead of a ConfigMap as it now holds credentials (agent + user)
* E2E tests have been updated to accept user/password - this is temporary and an improved solution is being implemented as part of #104

TODO:
* Add E2E tests testing for correct credential handling, currently E2E tests verify scram authentication is working, but not much to do with k8s resources.
* CRD change (will come with E2E)

EDIT:

CRD generated with the command `operator-sdk generate crds --crd-version v1beta1`


deploy/crds/mongodb.com_v1_mongodb_scram_cr.yaml
```diff
@@ -8,7 +8,6 @@ spec:
   version: "4.2.6"
   security:
     authentication:
-      enabled: true
```
chatton, line 11: authentication is no longer optional

pkg/controller/mongodb/replica_set_controller.go
```diff
@@ -334,11 +345,11 @@ func (r ReplicaSetReconciler) updateAndReturnStatusSuccess(mdb *mdbv1.MongoDB) (
 }
 
 func (r ReplicaSetReconciler) ensureAutomationConfig(mdb mdbv1.MongoDB) error {
-	cm, err := r.buildAutomationConfigConfigMap(mdb)
+	s, err := r.buildAutomationConfigSecret(mdb)
```
fabianlindfors, line 348: :+1: Switching from a ConfigMap to a Secret for an existing cluster will reset the automation config version to 1. I checked with the agent team and they said this shouldn't cause any issues.

pkg/authentication/scram/scram.go
```diff
@@ -47,9 +81,238 @@ func EnsureAgentSecret(getUpdateCreator secret.GetUpdateCreator, secretNsName ty
...
+	if err != nil {
+		return false, err
+	}
+
+	existingSha1Creds, existingSha256Creds, err := readExistingCredentials(secretGetter, mdb.NamespacedName(), user.Name)
+	if err != nil {
+		return false, err
+	}
+
+	sha1CredsAreTheSame := reflect.DeepEqual(sha1Creds, existingSha1Creds)
```
fabianlindfors, line 172: Is `DeepEqual` necessary here? `ScramCreds` have no nested types and can be compared using `==` if I'm not mistaken.

chatton: you're dead right! 👍

pkg/authentication/scram/scram.go
```diff
@@ -47,9 +81,238 @@ func EnsureAgentSecret(getUpdateCreator secret.GetUpdateCreator, secretNsName ty
...
+		return scramcredentials.ScramCreds{}, scramcredentials.ScramCreds{}, err
+	}
+
+	// we should really never hit this situation. It would only be possible if the secret storing credentials is manually edited.
+	if !secret.HasAllKeys(credentialsSecret, sha1SaltKey, sha1ServerKeyKey, sha1ServerKeyKey, sha256SaltKey, sha256ServerKeyKey, sha256StoredKeyKey) {
+		return scramcredentials.ScramCreds{}, scramcredentials.ScramCreds{}, fmt.Errorf("credentials secret did not have all of the required keys")
+	}
+
+	scramSha1Creds := scramcredentials.ScramCreds{
+		IterationCount: 10000,
```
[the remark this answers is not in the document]
chatton: good idea 👍

pkg/authentication/scram/scram.go
```diff
@@ -47,9 +81,238 @@ func EnsureAgentSecret(getUpdateCreator secret.GetUpdateCreator, secretNsName ty
 		agentSecret.Data[AgentKeyfileKey] = []byte(generatedContents)
 	}
 
-	return automationConfigModification(
-		string(agentSecret.Data[AgentPasswordKey]),
-		string(agentSecret.Data[AgentKeyfileKey]),
-		[]automationconfig.MongoDBUser{},
-	), getUpdateCreator.UpdateSecret(agentSecret)
+	return automationConfigModification(string(agentSecret.Data[AgentPasswordKey]), string(agentSecret.Data[AgentKeyfileKey]), desiredUsers), secret.CreateOrUpdate(secretGetUpdateCreateDeleter, agentSecret)
```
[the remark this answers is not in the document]
chatton: yes I agree! What do you think about

```golang
	if err := secret.CreateOrUpdate(secretGetUpdateCreateDeleter, agentSecret); err != nil {
		return automationconfig.NOOP(), err
	}

	return automationConfigModification(
		string(agentSecret.Data[AgentPasswordKey]),
		string(agentSecret.Data[AgentKeyfileKey]), desiredUsers,
	), nil
```

I think this makes it easier to understand as we have an explicit `nil` returned .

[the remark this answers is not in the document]
fabianlindfors: Even better!

pkg/authentication/scram/scram.go
```diff
@@ -47,9 +81,238 @@ func EnsureAgentSecret(getUpdateCreator secret.GetUpdateCreator, secretNsName ty
...
+func ensureScramCredentials(getUpdateCreator secret.GetUpdateCreator, user mdbv1.MongoDBUser, mdb mdbv1.MongoDB) (scramcredentials.ScramCreds, scramcredentials.ScramCreds, error) {
+	passwordKey := user.PasswordSecretRef.Key
+	if passwordKey == "" {
+		passwordKey = defaultPasswordKey
+	}
+	password, err := secret.ReadKey(getUpdateCreator, passwordKey, types.NamespacedName{Name: user.PasswordSecretRef.Name, Namespace: mdb.Namespace})
+	if err != nil {
+		// if the password is deleted, that's fine we can read from the stored credentials that were previously generated
+		if errors.IsNotFound(err) {
+			zap.S().Debugf("password secret was not found, reading from credentials from secret/%s", scramCredentialsSecretName(mdb.Name, user.Name))
```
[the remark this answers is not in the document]
chatton: Yes it can definitely be cleaned up. I really don't like the idea of adding the logger to the method signature as it would make it less importable from other packages. 

However maybe this is the best thing to do as it allows us to easily distinguish between different resources using a `SugaredLogger`. I would say if we do vote for this, we should ensure the logging is consistent all throughout the application.

pkg/authentication/scram/scram.go
```diff
@@ -47,9 +78,250 @@ func EnsureAgentSecret(getUpdateCreator secret.GetUpdateCreator, secretNsName ty
...
+
+	sha1CredsAreDifferent := sha1Creds != existingSha1Creds
+	sha256CredsAreDifferent := sha256Creds != existingSha256Creds
+
+	return sha1CredsAreDifferent || sha256CredsAreDifferent, nil
+}
+
+// generateSalts generates 2 different salts. The first is for the sha1 algorithm
+// the second is for sha256
+func generateSalts() ([]byte, []byte, error) {
```
rodrigovalin, line 189: do these belong to the `generate` package?

pkg/authentication/scram/scram_test.go
```diff
@@ -0,0 +1,337 @@
...
+
+	assert.Equal(t, testSha256Salt, scram256Creds.Salt)
+	assert.Equal(t, testSha256StoredKey, scram256Creds.StoredKey)
+	assert.Equal(t, testSha256ServerKey, scram256Creds.ServerKey)
+	assert.Equal(t, 15000, scram256Creds.IterationCount)
+}
+
+func validScramCredentialsSecret(objectKey types.NamespacedName, username string) corev1.Secret {
+	return secret.Builder(). // valid secret
+					SetName(scramCredentialsSecretName(objectKey.Name, username)).
```
rodrigovalin, line 318: funny formatting, maybe move the comment about the return stmt to fix it?

pkg/authentication/scram/scram.go
```diff
@@ -47,9 +78,250 @@ func EnsureAgentSecret(getUpdateCreator secret.GetUpdateCreator, secretNsName ty
...
+	}
+
+	// the password has changed, or we are generating it for the first time
+	sha1Creds, sha256Creds, err := generateScramShaCredentials(user.GetUserName(), password)
+	if err != nil {
+		return scramcredentials.ScramCreds{}, scramcredentials.ScramCreds{}, err
+	}
+
+	// create or update our credentials secret for this user
+	zap.S().Debugf("Generating new credentials and storing in secret/%s", scramCredentialsSecretName(mdbNamespacedName.Name, user.GetUserName()))
```
rodrigovalin, line 134: the log here does not seems to be in-line with what's going on, the credentials have been generated already

pkg/authentication/scram/scram.go
```diff
@@ -47,9 +78,250 @@ func EnsureAgentSecret(getUpdateCreator secret.GetUpdateCreator, secretNsName ty
...
+			zap.S().Debugf("password secret was not found, reading from credentials from secret/%s", scramCredentialsSecretName(mdbNamespacedName.Name, user.GetUserName()))
+			return readExistingCredentials(getUpdateCreator, mdbNamespacedName, user.GetUserName())
+		}
+		return scramcredentials.ScramCreds{}, scramcredentials.ScramCreds{}, err
+	}
+
+	// we should only need to generate new credentials in two situations.
+	// 1. We are creating the credentials for the first time
+	// 2. We are changing the password
+	needToGenerateNewCredentials, err := needToGenerateNewCredentials(getUpdateCreator, user.GetUserName(), mdbNamespacedName, password)
```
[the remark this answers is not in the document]
chatton: yes! Will rename

pkg/authentication/scram/scram.go
chatton: yes I think you're right!

PR #101, opened 2020-07-13
CLOUDP-67051: Update CRD with TLS settings

fabianlindfors:
### All Submissions:

This PR adds TLS settings to the CRD. The design we settled upon looks like this:

```
apiVersion: mongodb.com/v1
kind: MongoDB
metadata:
  name: example-mongodb
spec:
  members: 3
  type: ReplicaSet
  version: "4.2.8"
  security:
    tls:
      enabled: true
      caConfigMapRef:
        name: "example-mongodb-ca"
      certificateKeySecretRef:
        name: "example-mongodb-tls"
```

### DIFF
--- a/deploy/crds/mongodb.com_mongodb_crd.yaml
+++ b/deploy/crds/mongodb.com_mongodb_crd.yaml
@@ -49,6 +49,46 @@ spec:
             members:
               description: Members is the number of members in the replica set
               type: integer
+            security:
+              description: Security configures security features, such as TLS, and
+                authentication settings for a deployment
+              properties:
+                tls:
+                  description: TLS configuration for both client-server and server-server
+                    communication
+                  properties:
+                    caConfigMapRef:
+                      description: CaConfigMap is a reference to a ConfigMap containing
+                        the certificate for the CA which signed the server certificates
+                        The certificate is expected to be available under the key
+                        "ca.crt"
+                      properties:
+                        name:
+                          type: string
+                      required:
+                      - name
+                      type: object
+                    certificateKeySecretRef:
+                      description: CertificateKeySecret is a reference to a Secret
+                        containing a private key and certificate to use for TLS The
+                        key and cert are expected to be PEM encoded and available
+                        at "tls.key" and "tls.crt"
+                      properties:
+                        name:
+                          type: string
+                      required:
+                      - name
+                      type: object
+                    enabled:
+                      type: boolean
+                    optional:
+                      description: Optional configures if TLS should be required or
+                        optional for connections
+                      type: boolean
+                  required:
+                  - enabled
+                  type: object
+              type: object
             type:
               description: Type defines which type of MongoDB deployment the resource
                 should create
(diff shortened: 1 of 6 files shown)


deploy/crds/mongodb.com_mongodb_crd.yaml
```diff
@@ -49,6 +49,46 @@ spec:
...
+                        The certificate is expected to be available under the key
+                        "ca.crt"
+                      properties:
+                        name:
+                          type: string
+                      required:
+                      - name
+                      type: object
+                    certificateKeySecretRef:
+                      description: CertificateKeySecret is a reference to a Secret
```
rodrigovalin, line 72: Maybe mention that this is a kubernetes.io/tls Secret type

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -58,15 +58,23 @@ type TLS struct {
...
-	// CAConfigMapName is the name of a ConfigMap containing the certificate for the CA which signed the server certificates
+	// CaConfigMap is a reference to a ConfigMap containing the certificate for the CA which signed the server certificates
 	// The certificate is expected to be available under the key "ca.crt"
 	// +optional
-	CAConfigMapName string `json:"caConfigMapName"`
+	CaConfigMap LocalObjectReference `json:"caConfigMapRef"`
+}
+
+// LocalObjectReference is a reference to another Kubernetes object by name.
+// TODO: Replace with a type from the K8s API. CoreV1 has an equivalent
```
rodrigovalin, line 73: I assume then that this TODO won't show in our CRD, is that right?

fabianlindfors: Exactly! There TODO is on the `Name` field which the Operator SDK uses when generating the CRD.

deploy/crds/mongodb.com_mongodb_crd.yaml
fabianlindfors: @rodrigovalin We don't require the secret to be of type `kubernetes.io/tls`, only that the right fields exist. Is it still worth mentioning?

rodrigovalin: it makes sense to mention that the secret is not expected to be "one type or another", that can't be enforced, as much as I understand, but it does not hurt to tell people that that's exactly the format we expect

deploy/crds/mongodb.com_mongodb_crd.yaml
```diff
@@ -49,6 +49,46 @@ spec:
...
+                        containing a private key and certificate to use for TLS The
```
bznein, line 73: [nit] is a period missing between TSL and The?

deploy/crds/mongodb.com_mongodb_crd.yaml
fabianlindfors: True! Updated the description now to mention this.

deploy/crds/mongodb.com_mongodb_crd.yaml
fabianlindfors: Sure is. Good catch!

PR #105, opened 2020-07-15
CLOUDP-66661: Enable graceful shutdowns

fabianlindfors:
### All Submissions:

When Kubernetes shuts down a pod it will send a SIGTERM signal to each container and give them 30s to gracefully shutdown. Currently, our `mongod` won't get the SIGTERM signal, causing the server to be forcefully shut down and possibly causing a primary loss. The larger underlying problem is that `mongod` is not the main process (PID 1) of its container but started from inside a bash script. The reason we've designed it this way is because of our pre-stop hook, used to handle version changes.

This PR switches from a pre-stop hook to a post-start hook, run before `mongod` is started in the container. The majority of the changes in this PR are to switch the name of the pre-stop hook and the ones that enable graceful shutdowns are quite simple. We now use  `exec` in the bash script, causing `mongod` to replace the bash process and get PID 1. This in turn enables graceful shutdowns as any SIGTERM signals will reach the server.

We've had some flaky tests which could be related to this issue. This PR probably won't fix those issues as many tests run on MongoDB 4.0.6, whilst support for SIGTERM didn't arrive until 4.0.8. We should create a new PR to upgrade our tests to a version where the graceful shutdowns can be handled properly.


pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -540,15 +540,17 @@ func mongodbContainer(version string, volumeMounts []corev1.VolumeMount) contain
...
+
+# wait for config to be created by the agent
+while [ ! -f /data/automation-mongod.conf ]; do sleep 3 ; done ; sleep 2 ;
 
-# start the pre-stop-hook to restart the Pod when needed
-# If the Pod does not require to be restarted, the pre-stop-hook will
-# exit(0) for Kubernetes to restart the container.
-/hooks/pre-stop-hook ;
+# start mongod with this configuration
+exec mongod -f /data/automation-mongod.conf ;
```
fabianlindfors, line 553: This `exec` causes `mongod` to become PID 1 and in turn enables graceful shutdowns.

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -540,15 +540,17 @@ func mongodbContainer(version string, volumeMounts []corev1.VolumeMount) contain
...
 		"-c",
-		// we execute the pre-stop hook once the mongod has been gracefully shut down by the agent.
-		`while [ ! -f /data/automation-mongod.conf ]; do sleep 3 ; done ; sleep 2 ;
-# start mongod with this configuration
-mongod -f /data/automation-mongod.conf ;
+		`
+# run post-start hook to handle version changes.
+# during a version change, mongod will be stopped and the
+# container will be restarted, at which points this hook runs.
+/hooks/version-upgrade-hook
```
fabianlindfors, line 547: We could have used the container lifecycle post-start hook instead of including the hook in our script. This would cause the hook to run concurrently with `mongod`, resulting in the server starting up one extra time before being killed again.

We opted to trigger the post-start hook from inside the bash script to avoid this extra restart.

cmd/versionhook/main.go
```diff
@@ -20,7 +20,7 @@ import (
 
 const (
 	agentStatusFilePathEnv = "AGENT_STATUS_FILEPATH"
-	logFilePathEnv         = "PRE_STOP_HOOK_LOG_PATH"
+	logFilePathEnv         = "VERSION_UPGRADE_HOOK_LOG_PATH"
```
fabianlindfors, line 23: Suggestion: instead of writing the log to a file we could simply write it to stdout and let Kubernetes pick up on the logs. This would make debugging easier as the log currently is unavailable when `mongod` is stopped and the log file will also be removed during a restart.

LouisPlisso: yes this was valid for enterprise but i thought all logs were in stdout

pkg/controller/mongodb/mongodb_controller.go
```diff
@@ -540,15 +540,17 @@ func mongodbContainer(version string, volumeMounts []corev1.VolumeMount) contain
...
 		"/bin/sh",
 		"-c",
-		// we execute the pre-stop hook once the mongod has been gracefully shut down by the agent.
-		`while [ ! -f /data/automation-mongod.conf ]; do sleep 3 ; done ; sleep 2 ;
-# start mongod with this configuration
-mongod -f /data/automation-mongod.conf ;
+		`
+# run post-start hook to handle version changes.
+# during a version change, mongod will be stopped and the
+# container will be restarted, at which points this hook runs.
```
LouisPlisso, line 546: I think we should stick to 1 line comment in the startup script and we have the details in the docs

pkg/controller/mongodb/mongodb_controller.go
LouisPlisso: indeed

LouisPlisso: we can probably just name it without hook since it's in the hooks directory

cmd/versionhook/main.go
fabianlindfors: Cool. Changed now to write directly to stdout

pkg/controller/mongodb/mongodb_controller.go
fabianlindfors: Agreed. Updated the comment and `architecture.md`.

pkg/controller/mongodb/mongodb_controller.go
fabianlindfors: That's true, renamed!

cmd/versionhook/main.go
```diff
@@ -29,19 +28,27 @@ const (
 )
 
 func main() {
-	fmt.Println("Calling pre-stop hook!")
```
chatton, line 32: thank you! Not sure how this one slipped in!

scripts/dev/e2e.py
```diff
@@ -123,11 +123,11 @@ def build_and_push_e2e(repo_url: str, tag: str, path: str) -> None:
     build_and_push_image(repo_url, tag, path, "e2e")
 
 
-def build_and_push_prehook(repo_url: str, tag: str, path: str) -> None:
+def build_and_push_version_upgrade_hook(repo_url: str, tag: str, path: str) -> None:
     """
-    build_and_push_prehook builds and pushes the pre-stop-hook image.
+    build_and_push_version_upgrade_hook builds and pushes the pod-deleter image.
```
chatton, line 128: we're no longer calling it the `pod-deleter` image right?

fabianlindfors: We sure aren't, thank you!

PR #110, opened 2020-07-16
Refactor authentication and TLS with unified automation config modification

fabianlindfors:
### All Submissions:

This PR refactors authentication (SCRAM-SHA) and TLS to use a unified way of setting up the automation config. We introduce a new `Modification` function for the automation config, similar to the functional builders used for Kubernetes objects. The builder is changed to accept these generic modification functions instead of having special fields for TLS and authentication. This way we reduce the operators knowledge about authentication and TLS and make the automation config simpler to extend.

With this PR we also re-arrange `automation_config.go` as the current order of types and functions is inconsistent and not top-down. There have been no other changes to that file.

### DIFF
--- a/pkg/automationconfig/automation_config_test.go
+++ b/pkg/automationconfig/automation_config_test.go
@@ -176,69 +176,14 @@ func TestVersionManifest_BuildsForVersion(t *testing.T) {
 	assert.Empty(t, version.Builds)
 }
 
-func TestTLS(t *testing.T) {
-	caPath := "/path/to/ca"
-	certAndKeyPath := "/path/to/cert"
-	mode := TLSModeRequired
-
-	ac, err := NewBuilder().
-		SetName("my-rs").
-		SetDomain("my-ns.svc.cluster.local").
-		SetMongoDBVersion("4.2.0").
-		SetMembers(3).
-		SetFCV("4.0").
-		SetTLS(caPath, certAndKeyPath, mode).
-		Build()
-
-	assert.NoError(t, err)
-	assert.Len(t, ac.Processes, 3)
-
-	// Ensure every process has TLS configured
-	for _, p := range ac.Processes {
-		assert.Equal(t, mode, p.Args26.Net.TLS.Mode)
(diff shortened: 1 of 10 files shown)


pkg/controller/mongodb/mongodb_authentication.go
```diff
@@ -14,29 +14,23 @@ const (
 	scramShaOption = "SCRAM"
 )
 
-// noOpAuthEnabler performs no changes, leaving authentication settings untouched
-type noOpAuthEnabler struct{}
-
-func (n noOpAuthEnabler) EnableAuth(auth automationconfig.Auth) automationconfig.Auth {
-	return auth
-}
-
```
chatton, line 23: 🥳

pkg/automationconfig/automation_config_test.go
```diff
@@ -176,55 +176,16 @@ func TestVersionManifest_BuildsForVersion(t *testing.T) {
...
-	for _, p := range ac.Processes {
-		assert.Equal(t, mode, p.Args26.Net.TLS.Mode)
-		assert.Equal(t, caPath, p.Args26.Net.TLS.CAFile)
-		assert.Equal(t, certAndKeyPath, p.Args26.Net.TLS.PEMKeyFile)
-		assert.True(t, p.Args26.Net.TLS.AllowConnectionsWithoutCertificate)
-	}
-
-	// Ensure the CA was configured as trusted for the agent
-	assert.Equal(t, caPath, ac.TLS.CAFilePath)
-	assert.Equal(t, ClientCertificateModeOptional, ac.TLS.ClientCertificateMode)
```
chatton, line 206: unless I'm missing something, I don't see the corresponding  tests added using the new functional approach. Do we have tests ensuring these fields are set correctly?

fabianlindfors: These were already tested since before in `replicaset_controller_test.go` with `TestAutomationConfig_IsCorrectlyConfiguredWithTLS`.

PR #115, opened 2020-07-20
Switch to operator-managed secret for TLS

fabianlindfors:
### All Submissions:

To simplify adding support for TLS rotations we have opted to replace the init container with an operator-managed secret. The init container was previously used to concatenate the certificate and key from the mounted TLS secret and write them to a new file that MongoDB can use. After some discussion we decided to instead have the operator create a new secret containing the concatenated cert and key which we'll mount to each pod. This PR also moves the TLS unit tests to a new file `mongodb_tls_test.go` as the main unit tests were getting quite big.

Me and @chatton have been discussing how to structure the files and packages now that TLS and SCRAM have added a bit of complexity to the `pkg/controller/mongodb` package. Breaking out all functionality is a bit tricky at the moment as there is a lot of coupling between the reconciler and these features. We haven't yet found a great solution but it will become easier once we refactor the reconciliation loop, which is planned after the TLS and SCRAM release.


pkg/controller/mongodb/mongodb_tls_test.go
```diff
@@ -0,0 +1,240 @@
+package mongodb
```
fabianlindfors, line 1: Moved from `replicaset_controller_test.go` and added `TestTLSOperatorSecret`.

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -73,16 +76,50 @@ func (r *ReplicaSetReconciler) validateTLSConfig(mdb mdbv1.MongoDB) (bool, error
...
+		return err
 	}
 
+	combined := fmt.Sprintf("%s%s", data[tlsSecretCertName], data[tlsSecretKeyName])
+
+	operatorSecret := secret.Builder().
+		SetName(mdb.TLSOperatorSecretNamespacedName().Name).
+		SetNamespace(mdb.TLSOperatorSecretNamespacedName().Namespace).
+		SetField(tlsOperatorSecretFileName, combined).
+		Build()
```
fabianlindfors, line 114: Question: should we set the owner reference of the secret to point to the MDB resource? I don't see a reason to keep the secret around.

rodrigovalin: it is a good idea to have the ownerreference set to something, so the resources belong to something, for maintenance or cleanup purposes.

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -73,16 +76,50 @@ func (r *ReplicaSetReconciler) validateTLSConfig(mdb mdbv1.MongoDB) (bool, error
...
+
+// ensureTLSSecret will create or update the operator-managed Secret containing
+// the concatenated certificate and key from the user-provided Secret.
+func ensureTLSSecret(getUpdateCreator secret.GetUpdateCreator, mdb mdbv1.MongoDB) error {
+	data, err := secret.ReadStringData(getUpdateCreator, mdb.TLSSecretNamespacedName())
+	if err != nil {
+		return err
 	}
 
+	combined := fmt.Sprintf("%s%s", data[tlsSecretCertName], data[tlsSecretKeyName])
```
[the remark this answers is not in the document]
fabianlindfors: This is checked earlier in the reconciliation by `validateTLSConfig` so they should always exist. It might be clearer to add that here as well. What do you say?

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -73,16 +76,50 @@ func (r *ReplicaSetReconciler) validateTLSConfig(mdb mdbv1.MongoDB) (bool, error
...
+	}
+
+	// The config is only updated after the certs and keys have been rolled out to all pods.
+	// The agent needs these to be in place before the config is updated.
+	// Once the config is updated, the agents will gradually enable TLS in accordance with: https://docs.mongodb.com/manual/tutorial/upgrade-cluster-to-ssl/
+	if hasRolledOutTLS(mdb) {
+		return tlsConfigModification(mdb), nil
+	} else {
+		return automationconfig.NOOP(), nil
+	}
```
chatton, line 99: [nit] redundant else clause

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -73,16 +76,50 @@ func (r *ReplicaSetReconciler) validateTLSConfig(mdb mdbv1.MongoDB) (bool, error
...
-	if !(mdb.Spec.Security.TLS.Enabled && hasRolledOutTLS(mdb)) {
-		return automationconfig.NOOP()
+// It will also ensure that the combined cert-key secret is created.
+func getTLSConfigModification(getUpdateCreator secret.GetUpdateCreator, mdb mdbv1.MongoDB) (automationconfig.Modification, error) {
+	if !mdb.Spec.Security.TLS.Enabled {
+		return automationconfig.NOOP(), nil
+	}
+
+	err := ensureTLSSecret(getUpdateCreator, mdb)
+	if err != nil {
```
chatton, line 88: [nit] as err isn't used elsewhere, we can narrow its scope with

```golang
if err := ensureTLSSecret(getUpdateCreator, mdb); err != nil {
    return automationconfig.NOOP(), err
}
```

pkg/controller/mongodb/mongodb_tls.go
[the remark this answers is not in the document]
chatton: I would vote for adding additional checks, if this function is called from a separate context, the validation would be nice to have.

pkg/controller/mongodb/mongodb_tls.go
fabianlindfors: Good! Added now.

pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -73,16 +78,51 @@ func (r *ReplicaSetReconciler) validateTLSConfig(mdb mdbv1.MongoDB) (bool, error
...
+	combined := fmt.Sprintf("%s%s", data[tlsSecretCertName], data[tlsSecretKeyName])
+
+	operatorSecret := secret.Builder().
+		SetName(mdb.TLSOperatorSecretNamespacedName().Name).
+		SetNamespace(mdb.TLSOperatorSecretNamespacedName().Namespace).
+		SetField(tlsOperatorSecretFileName, combined).
+		SetOwnerReferences([]metav1.OwnerReference{getOwnerReference(mdb)}).
+		Build()
+
+	return secret.CreateOrUpdate(getUpdateCreator, operatorSecret)
```
chatton, line 119: ❗  watch out there is a bug in this function (fixed in #99 ) if you want to grab it from there

pkg/controller/mongodb/mongodb_tls.go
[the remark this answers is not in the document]
rodrigovalin: maybe it is `ensureTLSSecret` the one that should call `validateTLSConfig`. For 1 reason in particular: The Secret validated by `validateTLSConfig` is not necessarilly the same we have here, because we are fetching it twice from the API.

We can do:

1. getTLSConfigModification (from reconcile)
2. ensureTLSSecret (from 1)
3. `validateTLSConfig` (from 2), and making this function to validate a Secret that is passed as an argument (not reading it again from inside the function). If that function succeeds, it means that the Secret that was validated, is the same one that will be used next

makes sense?

pkg/controller/mongodb/mongodb_tls.go
fabianlindfors: Already fixed that one but now it seems to have disappeared. Me and Git do not get along...

pkg/controller/mongodb/mongodb_tls.go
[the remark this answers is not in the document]
fabianlindfors: Me and Rodrigo discussed this and came to the conclusion that it's hard to fix with our current design. The problem is that `validateTLSConfig` returns a boolean which is used to control the requeue. If we move the function we would have to bubble this result all the way up to the reconciliation loop. We should keep this in mind when doing the reconciliation refactoring later.

For now, I have gone ahead and added extra checks by using the helper functions from the `secret` package. @chatton is that the kind of validation you had in mind?

PR #117, opened 2020-07-21
Add ability to watch secondary resources

fabianlindfors:
### All Submissions:

* [X] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change

This PR implements a custom `handler.EventHandler` called `ResourceWatcher`. This will allow us to watch for changes to specific secondary resources outside of the main MDB CRD. This PR is part of the work to implement support for TLS certificate rotations. The idea is to have one `ResourceWatcher` on the `Reconciler` for each resource type we want to watch. If we for example have a `secretWatcher` on our `Reconciler` we can register it like:

```
// Watch for changes to primary resource MongoDB
err = c.Watch(&source.Kind{Type: &mdbv1.MongoDB{}}, &handler.EnqueueRequestForObject{}, predicates.OnlyOnSpecChange())
if err != nil {
	return err
}

// Watch for changes to secondary Secrets
err = c.Watch(&source.Kind{Type: &corev1.Secret{}}, r.secretWatcher)
if err != nil {
	return err
}
```

Resource to watch can then be added from the reconciliation loop using: `r.secretWatcher.Watch(objectToWatch, dependentObject)`.

### DIFF
--- a/pkg/controller/watch/watch.go
+++ b/pkg/controller/watch/watch.go
@@ -23,8 +23,8 @@ func New() ResourceWatcher {
 	}
 }
 
-// Add will add a new object to watch.
-func (w ResourceWatcher) Add(watchedName, dependentName types.NamespacedName) {
+// Watch will add a new object to watch.
+func (w ResourceWatcher) Watch(watchedName, dependentName types.NamespacedName) {
 	existing, hasExisting := w.watched[watchedName]
 	if !hasExisting {
 		existing = []types.NamespacedName{}
--- a/pkg/controller/watch/watch.go
+++ b/pkg/controller/watch/watch.go
@@ -1,6 +1,7 @@
 package watch
 
 import (
+	"github.com/mongodb/mongodb-kubernetes-operator/pkg/util/contains"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
@@ -30,10 +31,8 @@ func (w ResourceWatcher) Add(watchedName, dependentName types.NamespacedName) {
 	}
 
 	// Check if resource is already being watched.
-	for _, existingName := range existing {
-		if dependentName == existingName {
-			return
-		}
+	if contains.NamespacedName(existing, dependentName) {
+		return
 	}
 
 	w.watched[watchedName] = append(existing, dependentName)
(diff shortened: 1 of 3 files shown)


pkg/controller/watch/watch.go
```diff
@@ -0,0 +1,73 @@
...
+
+// New will create a new ResourceWatcher with no watched objects.
+func New() ResourceWatcher {
+	return ResourceWatcher{
+		watched: make(map[types.NamespacedName][]types.NamespacedName),
+	}
+}
+
+// Add will add a new object to watch.
+func (w ResourceWatcher) Add(watchedName, dependentName types.NamespacedName) {
```
chatton, line 26: [nit] The name `Add` is a bit misleading, I think maybe something like `EnsureAdded` or `EnsureWatched` would be more accurate, what do you think?

pkg/controller/watch/watch.go
```diff
@@ -0,0 +1,73 @@
...
+	if !hasExisting {
+		existing = []types.NamespacedName{}
+	}
+
+	// Check if resource is already being watched.
+	for _, existingName := range existing {
+		if dependentName == existingName {
+			return
+		}
+	}
```
chatton, line 37: we cloud maybe add a  `contains.NamespacedName(nsNames, nsName)` for this.

fabianlindfors: True, added!

pkg/controller/watch/watch.go
fabianlindfors: I agree that `Add` is a bit misleading but I'm not sure including `Ensure` adds much. How about `Watch`? Feels natural that you would `Watch` with a `Watcher`!

chatton: I think I'm okay with Watch!

pkg/util/contains/contains.go
```diff
@@ -19,3 +22,12 @@ func AuthMode(slice []mdbv1.AuthMode, s mdbv1.AuthMode) bool {
...
 }
+
+func NamespacedName(nsNames []types.NamespacedName, nsName types.NamespacedName) bool {
+	for _, elem := range nsNames {
+		if elem == nsName {
+			return true
+		}
+	}
+	return false
+}
```
chatton, line 33: 👍

pkg/controller/watch/watch.go
fabianlindfors: Renamed!

PR #122, opened 2020-07-23
CLOUDP-66615: Add support for rotation of TLS certificates and keys

fabianlindfors:
### All Submissions:

This PR adds support for rotation of TLS certificates and private keys. When the Secret containing the cert and key is updated, the operator will automatically roll out the change to all servers. This is made possible by two changes:

1. The operator now watches for changes to the TLS secret provided by the user and triggers a reconciliation on change. During a reconciliation the operator will update the operator-managed Secret. This was made possible by #117.
2. The concatenated cert and key is now mounted to a file where the file name contains the hash of the certificate and key. For the agent to update the certificate used on the servers, some field needs to change in the automation config. Using a hash-based file name causes the automation config to change only when the certificate or key contents change, triggering a rolling update by the agents.

A new E2E test, `replica_set_tls_rotate`, has also been added. The existing test certificate has been updated to have serial number 1 instead of the previous randomly generated one. The new test certificate used for rotations has serial number 2, which the E2E test checks for when ensuring the certificate was rotated.


pkg/controller/mongodb/mongodb_tls.go
```diff
@@ -84,47 +87,62 @@ func getTLSConfigModification(getUpdateCreator secret.GetUpdateCreator, mdb mdbv
...
+		return tlsConfigModification(mdb, certificateKeyFileName), nil
 	}
 
 	return automationconfig.NOOP(), nil
 }
 
 // ensureTLSSecret will create or update the operator-managed Secret containing
 // the concatenated certificate and key from the user-provided Secret.
-func ensureTLSSecret(getUpdateCreator secret.GetUpdateCreator, mdb mdbv1.MongoDB) error {
+func ensureTLSSecret(getUpdateCreator secret.GetUpdateCreator, mdb mdbv1.MongoDB) (string, error) {
```
bznein, line 107: [nit] should we maybe rename the function to make clearer what it returns? It is not immediately obvious. However I am terrible at naming so I can't come up with anything better!

pkg/controller/mongodb/replica_set_controller.go
```diff
@@ -78,18 +80,21 @@ func Add(mgr manager.Manager) error {
 type ManifestProvider func() (automationconfig.VersionManifest, error)
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager, manifestProvider ManifestProvider) reconcile.Reconciler {
+func newReconciler(mgr manager.Manager, manifestProvider ManifestProvider) *ReplicaSetReconciler {
```
bznein, line 83: We should update this comment? (or maybe get entirely rid of it? it's quite obvious what the function does)

pkg/controller/mongodb/replica_set_controller.go
```diff
@@ -78,18 +80,21 @@ func Add(mgr manager.Manager) error {
...
 		scheme:           mgr.GetScheme(),
 		manifestProvider: manifestProvider,
 		log:              zap.S(),
+		secretWatcher:    &secretWatcher,
 	}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
-func add(mgr manager.Manager, r reconcile.Reconciler) error {
+func add(mgr manager.Manager, r *ReplicaSetReconciler) error {
```
bznein, line 97: Same as above

pkg/controller/mongodb/mongodb_tls.go
fabianlindfors: Agreed but I'm horribly at naming as well. If the function is hard to name it might be because it does to much though. How about we split out the cert and key fetching to another function, like `func getCertAndKey(getter secret.Getter, mdb mdbv1.MongoDB) (string, string, error)` and then have the other function be: `func ensureTLSSecret(getUpdateCreator secret.GetUpdateCreator, mdbv1.MongoDB, cert, key string) error`?

pkg/controller/mongodb/replica_set_controller.go
fabianlindfors: Definitely should! I got rid of it.

pkg/controller/mongodb/mongodb_tls.go
bznein: I think that would be a nice thing to do!

pkg/controller/mongodb/replica_set_controller.go
fabianlindfors: I updated this comment to give a bit more info on what it's suppose to do.

pkg/controller/mongodb/mongodb_tls.go
fabianlindfors: Done!

PR #123, opened 2020-07-23
Add Delete Function for ConfigMap

chatton:
### All Submissions:

This PR adds a simple function to delete ConfigMaps which is consistent with our current patterns.

### DIFF
--- a/pkg/kube/client/client_test.go
+++ b/pkg/kube/client/client_test.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"testing"
 
+	"k8s.io/apimachinery/pkg/types"
+
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/kube/configmap"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


pkg/kube/client/client_test.go
```diff
@@ -62,3 +64,21 @@ func TestAddingDataField_ModifiesExistingObject(t *testing.T) {
...
+	err := client.CreateConfigMap(cm)
+	assert.NoError(t, err)
+
+	err = client.DeleteConfigMap(types.NamespacedName{Name: "config-map", Namespace: "default"})
+	assert.NoError(t, err)
+
+	_, err = client.GetConfigMap(types.NamespacedName{Name: "config-map", Namespace: "default"})
+	assert.Error(t, err)
+	assert.Equal(t, err, notFoundError())
+}
```
bznein, line 84: Do we need both assert here?

I feel if the first one would fail, it will fail anyway in the second? If you think that's clearer this way is totally fine for me anyway!

chatton: yep makes sense, we can just leave the second one in and remove the first! 👍

PR #128, opened 2020-07-31
(DOCSP-11347): Document TLS configuration for Community Operator

jwilliams-mongo:
[JIRA](https://jira.mongodb.org/browse/DOCSP-11347)

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,9 +234,9 @@ To secure connections to MongoDB resources using TLS:
    kubectl apply -f <example>.yaml --namespace <my-namespace>
    ```
 1. From within the Kubernetes cluster, connect to the MongoDB resource.
-   - If `security.tls.optional` is omitted or `false`: clients must
+   - If `spec.security.tls.optional` is omitted or `false`: clients must
      establish TLS connections to the MongoDB servers in the replica set.
-   - If `security.tls.optional` is true, clients can establish TLS or
+   - If `spec.security.tls.optional` is true, clients can establish TLS or
      non-TLS connections to the MongoDB servers in the replica set.
 
    See the documentation for your connection method to learn how to establish a TLS connection to a MongoDB server.
--- a/README.md
+++ b/README.md
@@ -168,9 +168,14 @@ Before you secure MongoDB resource connections using TLS, you must:
      <metadata.name of the MongoDB resource>-2.<metadata.name of the MongoDB resource>-svc.<namespace>.svc.cluster.local
      ```
 
-1. Create a Kubernetes ConfigMap that contains the certificate for the CA that signed your server certificate. For a certificate file named `ca.crt`:
+1. Create a Kubernetes ConfigMap that contains the certificate for the CA that signed your server certificate. The key in the ConfigMap that references the certificate must be named `ca.crt`. Kubernetes configures this automatically if the certificate file is named `ca.crt`:
    ```
    kubectl create configmap <tls-ca-configmap-name> --from-file=ca.crt --namespace <namespace>
+   ```
+
+   For a certificate file with any other name, you must define the `ca.crt` key manually:
+   ```
+   kubectl create configmap <tls-ca-configmap-name> --from-file=ca.crt=<certificate-file-name>.crt --namespace <namespace>
    ```
 
 1. Create a Kubernetes secret that contains the server certificate and key for the members of your replica set. For a server certificate named `server.crt` and key named `server.key`:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -139,6 +142,89 @@ To upgrade this resource from `4.0.6` to `4.2.7`:
...
+   ```
+1. From within the Kubernetes cluster, connect to the MongoDB resource.
+   - If `security.tls.optional` is omitted or `false`: clients must
+     establish TLS connections to the MongoDB servers in the replica set.
+   - If `security.tls.optional` is true, clients can establish TLS or
+     non-TLS connections to the MongoDB servers in the replica set.
+
+   See the documentation for your connection method to learn how to establish a TLS connection to a MongoDB server.
+
 ## Supported Features
```
melissamahoney-mongodb, line 228: Q: Do we need to add anything here about TLS support between MDB resources? Might not be necessary

jwilliams-mongo: amended line below to be more explicit about this

README.md
```diff
@@ -139,6 +142,89 @@ To upgrade this resource from `4.0.6` to `4.2.7`:
...
+     ---
+     **NOTE**
+
+     When you enable TLS on an existing replica set deployment:
+
+     a. Set `security.tls.optional` to `true`.
+     b. Apply the configuration to Kubernetes.
+     c. Upgrade your existing clients to use TLS.
+     d. Set `security.tls.optional` to `false`.
+     e. Complete the remaining steps in the procedure.
```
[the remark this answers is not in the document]
jwilliams-mongo: interesting -- looks OK in the md preview extension in vscode. different md flavors, perhaps? Think I fixed it

README.md
```diff
@@ -139,6 +142,89 @@ To upgrade this resource from `4.0.6` to `4.2.7`:
...
+You can configure the MongoDB Community Kubernetes Operator to use TLS certificates to encrypt traffic between:
+
+- MongoDB hosts in a replica set, and
+- Client applications and MongoDB deployments.
+
+### Prerequisites
+
+Before you secure MongoDB resource connections using TLS, you must:
+
+1. Using your own Certificate Authority (CA), create a PEM-encoded TLS certificate for each member of the Application Database’s replica set. The certificate must have one of the following:
```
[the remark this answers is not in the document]
jwilliams-mongo: updated

fabianlindfors: There is only a single certificate and private key used for all members. The user only needs to create a single one and not one for each member.

README.md
```diff
@@ -139,6 +142,89 @@ To upgrade this resource from `4.0.6` to `4.2.7`:
...
+     ```
+   - The domain name for each of the replica set members as `Subject Alternative Names` (SAN):
+
+     ```
+     <metadata.name of the MongoDB resource>-0.<metadata.name of the MongoDB resource>-svc.<namespace>.svc.cluster.local
+     <metadata.name of the MongoDB resource>-1.<metadata.name of the MongoDB resource>-svc.<namespace>.svc.cluster.local
+     <metadata.name of the MongoDB resource>-2.<metadata.name of the MongoDB resource>-svc.<namespace>.svc.cluster.local
+     ```
+
+1. Create a Kubernetes ConfigMap that contains the certificate for the CA that signed your server certificate. For a certificate file named `ca.crt`:
```
fabianlindfors, line 171: I think it's worth clarifying here that the key within the ConfigMap needs to be called `ca.crt`. This will happen automatically when using `--from-file=ca.crt` but not if the user has their certificate stored with another file name.

This does not apply for the Secret though. `kubectl` will automatically make sure the certificate and private key end up under `tls.crt` and `tls.key` respectively, no matter the file names used.

README.md
```diff
@@ -139,6 +142,89 @@ To upgrade this resource from `4.0.6` to `4.2.7`:
...
+
+     ---
+     **NOTE**
+
+     When you enable TLS on an existing replica set deployment:
+
+     a. Set `security.tls.optional` to `true`.
+     b. Apply the configuration to Kubernetes.
+     c. Upgrade your existing clients to use TLS.
+     d. Set `security.tls.optional` to `false`.
```
fabianlindfors, line 201: Might be even better to recommend removing `security.tls.optional` entirely. Should make for a clearer resource file and it shouldn't ever have to be changed again.

README.md
```diff
@@ -139,6 +142,89 @@ To upgrade this resource from `4.0.6` to `4.2.7`:
...
+
+     ---
+   - `security.tls.certificateKeySecretRef.name`: Name of the Kubernetes secret that contains the server certificate and key that you created in the [prerequisites](#prerequisites-1).
+   - `security.tls.caConfigMapRef.name`: Name of the Kubernetes ConfigMap that contains the Certificate Authority certificate used to sign the server certificate that you created in the [prerequisites](#prerequisites-1).
+
+   ```yaml
+   security:
+     tls:
+       enabled: true
+       optional: false
```
fabianlindfors, line 213: I'd prefer if we omitted `optional` from this example. It should only be used during upgrades and we don't want to encourage people to enable it in other situations. Also makes for a simpler example!

On top of that, I think it would be clearer if we included a full example of the resource definition like we do under "Upgrade your MongoDB Resource Version and Feature Compatibility Version". Something like:

```
apiVersion: mongodb.com/v1
kind: MongoDB
metadata:
  name: example-mongodb
spec:
  members: 3
  type: ReplicaSet
  version: "4.2.7"
  security:
    tls:
      enabled: true
      certificateKeySecretRef:
        name: <tls-secret-name>
      caConfigMapRef:
        name: <tls-ca-configmap-name>
```

README.md
```diff
@@ -139,6 +142,89 @@ To upgrade this resource from `4.0.6` to `4.2.7`:
...
+     When you enable TLS on an existing replica set deployment:
+
+     a. Set `security.tls.optional` to `true`.
+     b. Apply the configuration to Kubernetes.
+     c. Upgrade your existing clients to use TLS.
+     d. Set `security.tls.optional` to `false`.
+     e. Complete the remaining steps in the procedure.
+
+     ---
+   - `security.tls.certificateKeySecretRef.name`: Name of the Kubernetes secret that contains the server certificate and key that you created in the [prerequisites](#prerequisites-1).
```
fabianlindfors, line 206: I think we should clarify that these settings go inside the top-level `spec` setting on the resource definition: `spec.security.tls.certificateKeySecretRef.name`.

README.md
fabianlindfors: I don't think so. Once TLS is enabled, server-server encryption should just work and the user doesn't have to worry about it.

README.md
jwilliams-mongo: I think I addressed this in https://github.com/mongodb/mongodb-kubernetes-operator/pull/128/files#diff-04c6e90faac2675aa89e2176d2eec7d8R156

README.md
fabianlindfors: Didn’t see your comment when writing mine but that’s a nice way to clarify!

README.md
jwilliams-mongo: updated

README.md
jwilliams-mongo: good idea. updated

README.md
jwilliams-mongo: another good idea. updated.

README.md
jwilliams-mongo: updated

README.md
jwilliams-mongo: see https://github.com/mongodb/mongodb-kubernetes-operator/pull/128/files#diff-04c6e90faac2675aa89e2176d2eec7d8R243. I'm happy to revert this change if you like.

fabianlindfors: I like it, no need to change!

PR #129, opened 2020-08-03
Use generic map for "args2_6" in automation config

fabianlindfors:
### All Submissions:

This PR is in preparation for CLOUDP-58544 which adds support for arbitrary MongoDB configuration in the CRD. To support this change, we need to treat "args2_6" as a generic map, instead of a hard-coded struct.

To simplify interactions with the map, this PR uses the objx library: https://github.com/stretchr/objx. The library helps us write better code with less typecasting without having to write our helper functions. On top of that, it also includes support for merging maps which will be helpful for the next PR. Based on the following criteria I feel that the library is good enough to be included in our code base:

- Maturity: the library has been around for a while and is very well-tested. Most issues have been closed and no major features seem to be missing.
- Pulse: the library isn't seeing many changes but because of it's limited scope and current state there aren't many changes to be made.
- License: MIT

### DIFF
--- a/pkg/automationconfig/automation_config_test.go
+++ b/pkg/automationconfig/automation_config_test.go
@@ -40,12 +40,8 @@ func TestBuildAutomationConfig(t *testing.T) {
 	for i, p := range ac.Processes {
 		assert.Equal(t, Mongod, p.ProcessType)
 		assert.Equal(t, fmt.Sprintf("my-rs-%d.my-ns.svc.cluster.local", i), p.HostName)
-		assert.Equal(t, map[string]interface{}{
-			"dbPath": DefaultMongoDBDataDir,
-		}, p.Args26["storage"])
-		assert.Equal(t, map[string]interface{}{
-			"replSetName": "my-rs",
-		}, p.Args26["replication"], "replication should be configured based on the replica set name provided")
+		assert.Equal(t, DefaultMongoDBDataDir, p.Args26.Get("storage.dbPath").MustStr())
+		assert.Equal(t, "my-rs", p.Args26.Get("replication.replSetName").MustStr())
 		assert.Equal(t, toHostName("my-rs", i), p.Name)
 		assert.Equal(t, "4.2.0", p.Version)
 		assert.Equal(t, "4.0", p.FeatureCompatibilityVersion)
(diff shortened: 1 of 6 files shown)


pkg/automationconfig/automation_config_test.go
```diff
@@ -145,9 +144,9 @@ func TestProcessHasPortSetToDefault(t *testing.T) {
 
 	assert.NoError(t, err)
 	assert.Len(t, ac.Processes, 3)
-	assert.Equal(t, ac.Processes[0].Args26.Net.Port, 27017)
-	assert.Equal(t, ac.Processes[1].Args26.Net.Port, 27017)
-	assert.Equal(t, ac.Processes[2].Args26.Net.Port, 27017)
+	for _, process := range ac.Processes {
+		assert.Equal(t, 27017, process.Args26.Get("net.port").Data())
```
rodrigovalin, line 148: Can we get this as an int value instead?

pkg/controller/mongodb/mongodb_tls_test.go
```diff
@@ -164,12 +158,10 @@ func TestAutomationConfig_IsCorrectlyConfiguredWithTLS(t *testing.T) {
 		for _, process := range ac.Processes {
 			operatorSecretFileName := tlsOperatorSecretFileName("CERT", "KEY")
 
-			assert.Equal(t, automationconfig.MongoDBTLS{
-				Mode:                               automationconfig.TLSModePreferred,
-				PEMKeyFile:                         tlsOperatorSecretMountPath + operatorSecretFileName,
-				CAFile:                             tlsCAMountPath + tlsCACertName,
-				AllowConnectionsWithoutCertificate: true,
-			}, process.Args26.Net.TLS)
+			assert.Equal(t, automationconfig.TLSModePreferred, process.Args26.Get("net.tls.mode").Data())
```
rodrigovalin, line 161: I think we should get the values with the `Must*` functions? this one would be MustStr()

fabianlindfors: I considered that but eventually went with `Data` as I thought it would yield better error messages. When using `Must` we can fail either through a panic or because of the assert. If using `Data` we can only fail through the assert and I `testify` will yield a nice error message showing what type we got compared to what type was expected.

I'd be happy to change it though. What do you say?

bznein:
> I really like the idioms introduced by the objx library, I vote yes to include it.
> 
> @chatton @bznein do you have any objections?

No objections! I think it makes the code much clearer so that's a yes for me as well!

pkg/controller/mongodb/mongodb_tls_test.go
rodrigovalin: I don't have a strong opinion, but if Data will produce better error messages, it is a good reason to use it.

fabianlindfors: Just did a test and the error messages are definitely better:

```
expected: string("100")
actual  : int(100)
```

vs.

```
panic: interface conversion: interface {} is int, not string [recovered] + stack trace
```

PR #134, opened 2020-08-07
CLOUDP-58544: Add support for arbitrary mongod config in CRD

fabianlindfors:
### All Submissions:

This PR adds support for overriding mongod configuration in the CRD. Any settings will override the default ones without exception and the same settings must be used for all members.

### DIFF
--- a/pkg/apis/mongodb/v1/mongodb_types.go
+++ b/pkg/apis/mongodb/v1/mongodb_types.go
@@ -1,10 +1,13 @@
 package v1
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
 	"k8s.io/apimachinery/pkg/types"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -48,12 +51,48 @@ type MongoDBSpec struct {
 
 	// +optional
 	StatefulSetConfiguration StatefulSetConfiguration `json:"statefulset,omitempty"`
+
+	// AdditionalMongodConfig is additional configuration that can be passed to
+	// each data-bearing mongod at runtime. Uses the same structure as the mongod
+	// configuration file: https://docs.mongodb.com/manual/reference/configuration-options/
+	// +kubebuilder:validation:Type=object
+	AdditionalMongodConfig MongodConfiguration `json:"additionalMongodConfig,omitempty"`
 }
 
 // StatefulSetConfiguration holds the optional custom StatefulSet
 // that should be merged into the operator created one.
 type StatefulSetConfiguration struct {
 	Spec appsv1.StatefulSetSpec `json:"spec"`
+}
+
+// MongodConfiguration holds the optional mongod configuration
+// that should be merged with the operator created one.
+//
+// The CRD generator does not support map[string]interface{}
+// on the top level and hence we need to work around this with
+// a wrapping struct.
+type MongodConfiguration struct {
+	Object map[string]interface{} `json:"-"`
+}
+
+// MarshalJSON defers JSON encoding to the wrapped map
+func (m *MongodConfiguration) MarshalJSON() ([]byte, error) {
+	return json.Marshal(m.Object)
+}
+
+// UnmarshalJSON will decode the data into the wrapped map
+func (m *MongodConfiguration) UnmarshalJSON(data []byte) error {
+	if m.Object == nil {
+		m.Object = map[string]interface{}{}
+	}
+
+	return json.Unmarshal(data, &m.Object)
+}
+
+func (m *MongodConfiguration) DeepCopy() *MongodConfiguration {
+	return &MongodConfiguration{
+		Object: runtime.DeepCopyJSON(m.Object),
+	}
 }
 
 type MongoDBUser struct {
--- a/pkg/controller/mongodb/replica_set_controller.go
+++ b/pkg/controller/mongodb/replica_set_controller.go
@@ -9,6 +9,9 @@ import (
 	"os"
 	"time"
 
+	"github.com/imdario/mergo"
+	"github.com/stretchr/objx"
+
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/controller/watch"
 
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/kube/persistentvolumeclaim"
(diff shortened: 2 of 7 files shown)


pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -56,6 +65,36 @@ type StatefulSetConfiguration struct {
...
 }
 
+// MongodConfiguration holds the optional mongod configuration
+// that should be merged with the operator created one.
+//
+// The CRD generator does not support map[string]interface{}
+// on the top level and hence we need to work around this with
+// a wrapping struct.
+type MongodConfiguration struct {
+	Object map[string]interface{} `json:"-"`
```
fabianlindfors, line 75: The Operator SDK does not support `type MongodConfiguration map[string]interface{}` as we use in enterprise. According to https://github.com/kubernetes-sigs/controller-tools/pull/126 and other threads this is a deliberate decision.

They instead encourage people to use other types, for example `json.RawMessage`. That would work for us but I opted to use a wrapping struct instead to simplify the programmatic creation of MongoDB resource definitions. If we were to use `json.RawMessage`, our tests would have to marshal the JSON twice and it would be painful for anybody who wants to programmatically interface with our operator.

pkg/controller/mongodb/replica_set_controller.go
```diff
@@ -468,6 +478,17 @@ func (r ReplicaSetReconciler) buildAutomationConfigConfigMap(mdb mdbv1.MongoDB)
 		Build(), nil
 }
 
+// getMongodConfigModification will merge the additional configuration in the CRD
+// into the configuration set up by the operator.
+func getMongodConfigModification(mdb mdbv1.MongoDB) automationconfig.Modification {
+	return func(ac *automationconfig.AutomationConfig) {
+		for i, _ := range ac.Processes {
+			// Mergo requires both objects to have the same type
+			mergo.Merge(&ac.Processes[i].Args26, objx.New(mdb.Spec.AdditionalMongodConfig.Object), mergo.WithOverride)
```
fabianlindfors, line 487: `objx` has a built-in merge function but unfortunately it does not support recursive merging. As we are already using `mergo` it seemed like a good fit here as well.

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -48,6 +51,12 @@ type MongoDBSpec struct {
 
 	// +optional
 	StatefulSetConfiguration StatefulSetConfiguration `json:"statefulset,omitempty"`
+
+	// AdditionalMongodConfig is additional configuration that can be passed to
+	// each data-bearing mongod at runtime. Uses the same structure as the mongod
+	// configuration file: https://docs.mongodb.com/manual/reference/configuration-options/
+	// +kubebuilder:validation:Type=object
+	AdditionalMongodConfig MongodConfiguration `json:"additionalMongodConfig,omitempty"`
```
fabianlindfors, line 59: Same naming as we use in enterprise

pkg/controller/mongodb/replica_set_controller.go
```diff
@@ -452,7 +455,14 @@ func (r ReplicaSetReconciler) buildAutomationConfigConfigMap(mdb mdbv1.MongoDB)
...
 	}
 
-	ac, err := buildAutomationConfig(mdb, manifest.BuildsForVersion(mdb.Spec.Version), currentAC, authModification, tlsModification)
+	ac, err := buildAutomationConfig(
+		mdb,
+		manifest.BuildsForVersion(mdb.Spec.Version),
+		currentAC,
+		authModification,
+		tlsModification,
+		getMongodConfigModification(mdb),
```
chatton, line 464: `buildAutomationConfig` already has `mdb` as an argument, do we need to pass `getMongodConfigModification(mdb)` as an additional one? Maybe we can call `getMongodConfigModification(mdb)` from inside `buildAutomationConfig` what do you think?

pkg/controller/mongodb/replica_set_controller.go
```diff
@@ -468,6 +478,17 @@ func (r ReplicaSetReconciler) buildAutomationConfigConfigMap(mdb mdbv1.MongoDB)
 		Build(), nil
 }
 
+// getMongodConfigModification will merge the additional configuration in the CRD
+// into the configuration set up by the operator.
+func getMongodConfigModification(mdb mdbv1.MongoDB) automationconfig.Modification {
+	return func(ac *automationconfig.AutomationConfig) {
+		for i, _ := range ac.Processes {
```
chatton, line 485:
```suggestion
		for i := range ac.Processes {
```

pkg/controller/mongodb/replica_set_controller.go
chatton: `mergo.WithOverride` will do a direct override. I think this means if a user provides any tls settings, we lose all of the operator configured ones. I would have to double check that though! (maybe we could add a unit test for this)

pkg/controller/mongodb/replicaset_controller_test.go
```diff
@@ -301,6 +303,35 @@ func TestAutomationConfig_versionIsNotBumpedWithNoChanges(t *testing.T) {
 	assert.Equal(t, currentAc.Version, 1)
 }
 
+func TestAutomationConfig_CustomMongodConfig(t *testing.T) {
+	mdb := newTestReplicaSet()
+
+	mongodConfig := objx.New(map[string]interface{}{})
+	mongodConfig.Set("net.port", 1000)
+	mongodConfig.Set("arbitrary.config.path", "value")
```
chatton, line 311: this is a very nice interface for accessing nested values!

pkg/controller/mongodb/replicaset_controller_test.go
```diff
@@ -301,6 +303,35 @@ func TestAutomationConfig_versionIsNotBumpedWithNoChanges(t *testing.T) {
...
+	for _, p := range currentAc.Processes {
+		// Ensure port was overridden
+		assert.Equal(t, float64(1000), p.Args26.Get("net.port").Data())
+
+		// Ensure custom value was added
+		assert.Equal(t, "value", p.Args26.Get("arbitrary.config.path").Data())
+
+		// Ensure default settings went unchanged
+		assert.Equal(t, automationconfig.DefaultMongoDBDataDir, p.Args26.Get("storage.dbPath").Data())
+		assert.Equal(t, mdb.Name, p.Args26.Get("replication.replSetName").Data())
```
chatton, line 331: can we maybe change a net.val1, and ensure that an operator configured net value was untouched? We may be in a situation where the user would need to fully configure what the operator configures if they touch a single field

fabianlindfors: Yes, good idea. I've added another test and it seems as if mergo works nicely without overriding the other.

pkg/controller/mongodb/replica_set_controller.go
fabianlindfors: Good idea! You think this should be a new field on the builder or should we keep up the use of `Modification`?

pkg/controller/mongodb/replicaset_controller_test.go
fabianlindfors: Agreed!

pkg/controller/mongodb/replica_set_controller.go
fabianlindfors: I checked and it seems as if mergo is smart enough to not do this. I also added another unit test to ensure that this is the case.

pkg/controller/mongodb/replica_set_controller.go
chatton: I think we can do a Modification!

PR #141, opened 2020-08-14
CLOUDP-68700: automatic release process

bznein:
# This PR
Introduces automatic pushing of the tagged release image of the `operator` and the `version upgrade hook` to `quay`.
After discussing it with @chatton we decided to stick with the `release_blocker` approach as the new evg features of `git tags` doesn't really suit our case (happy to elaborate on this, if needed). 

The PR introduces two new evergreen tasks, that are blocked by default and which release the images if the dependency is overrode. The retagging is done through python's `docker` client, which pulls the image from the development repos, rebuilds it with a new label that makes it not expire after 10 hours, then retags it and pushes it to the correct release repo.

The release tags are took from a new file in the root of the repo, called `release.json` which must be updated when doing a release.

### All Submissions:


bznein:
evergreen retry

bznein:
evergreen retry

.evergreen.yml
```diff
@@ -305,6 +327,32 @@ tasks:
...
+          image_type: operator
+
+  - name: release_version_upgrade_hook
+    commands:
+      - func: clone
+      - func: setup_virtualenv
+      - func: release_docker_image
+        vars:
+          old_image: quay.io/mongodb/community-operator-version-upgrade-post-start-hook
+          new_image: quay.io/mongodb/mongodb-kubernetes-operator-version-upgrade-post-start-hook
```
chatton, line 352: maybe we can take this opportunity to re-name the dev version, we could just add a `-dev` to the repo name maybe. What do you think?

scripts/dev/release_image.py
```diff
@@ -0,0 +1,65 @@
...
+        "--username", help="username for the registry", type=str,
+    )
+    parser.add_argument(
+        "--password", help="password for the registry", type=str,
+    )
+    parser.add_argument(
+        "--registry", help="The docker registry", type=str,
+    )
+    parser.add_argument(
+        "--labels", help="Labels for the new image", type=json.loads,
```
chatton, line 33: how will `type=json.loads` work here? won't the type of `json.loads` be a function? Do we want this to be a `Dict` instead?

scripts/dev/release_image.py
```diff
@@ -0,0 +1,65 @@
...
+
+def main() -> int:
+    args = parse_args()
+    with open(args.release_file) as f:
+        release = json.load(f)
+
+    if args.image_type == "operator":
+        new_tag = release["mongodb-kubernetes-operator"]
+    elif args.image_type == "versionhook":
+        new_tag = release["version-upgrade-hook"]
```
chatton, line 49: if we pass something that isn't `operator` or `versionhook` to this script we might get some unexpected results. I think we should fail loudly if the value is unknown.

scripts/dev/dockerutil.py
```diff
@@ -28,6 +28,42 @@ def push_image(tag: str) -> None:
...
+    password: Optional[str] = None,
+    registry: Optional[str] = None,
+) -> None:
+    with open(f"{path}/Dockerfile", "w") as f:
+        f.write(f"FROM {old_repo_url}:{old_tag}")
+    client = docker.from_env()
+    if username is not None and password is not None and registry is not None:
+        client.login(username=username, password=password, registry=registry)
+
+    i, _ = client.images.build(path=f"{path}", labels=labels, tag=new_tag)
```
chatton, line 48: [nit] we should give this a proper name like `image`
[q] what is the second value we're ignoring?

scripts/dev/dockerutil.py
```diff
@@ -28,6 +28,42 @@ def push_image(tag: str) -> None:
...
+    path: str,
+    labels: Optional[dict] = None,
+    username: Optional[str] = None,
+    password: Optional[str] = None,
+    registry: Optional[str] = None,
+) -> None:
+    with open(f"{path}/Dockerfile", "w") as f:
+        f.write(f"FROM {old_repo_url}:{old_tag}")
+    client = docker.from_env()
+    if username is not None and password is not None and registry is not None:
```
chatton, line 45: [nit] this sort of multiple None check where each condition is the same can be written in a way that is a little more pythonic if you prefer it!

```python
if all(value is not None for value in [username, password, registry]):
   # login
```

scripts/dev/release_image.py
```diff
@@ -0,0 +1,65 @@
...
+    parser.add_argument(
+        "--password", help="password for the registry", type=str,
+    )
+    parser.add_argument(
+        "--registry", help="The docker registry", type=str,
+    )
+    parser.add_argument(
+        "--labels", help="Labels for the new image", type=json.loads,
+    )
+    parser.add_argument("--image_type", help="Type of image to be released")
```
chatton, line 35: are we able to restrict this to a preset list of values?

.evergreen.yml
```diff
@@ -346,3 +394,25 @@ buildvariants:
       - name: build_e2e_image
       - name: build_testrunner_image
       - name: build_prehook_image
+
+  - name: release_blocker
+    display_name: release_blocker
+    run_on:
+      - ubuntu1604-packer  # Note: cheapest machine I found
```
chatton, line 401: [q] does this one have docker installed the same way? I would lean in favour of just using the same machines that we use for our tests just to keep the variety of environments as low as possible.

.evergreen.yml
bznein: I totally agree! It actually took me a while to figure out the right repo and I kept forgetting it!

scripts/dev/release_image.py
bznein: That doesn't seem possible directly. All the accepted solutions are workarounds. I used this one: https://stackoverflow.com/questions/18608812/accepting-a-dictionary-as-an-argument-with-argparse-and-python but open to suggestions!

scripts/dev/dockerutil.py
bznein: From the docs: ` The second item is a generator of the build logs as JSON-decoded objects.` Do we want to use them to output logs or are we ok with ignoring it?

.evergreen.yml
bznein: This runs only on the release blocker. The task that releases the images uses `ubuntu1604-test` ass all the other tasks. Here we just need a machine which can ran the automatically failing task

.evergreen.yml
bznein: (decided to keep it like this for now)

scripts/dev/release_image.py
bznein: Addendum from that topic:

```
The reason this works is actually quite interesting because internally argparse will just use the parameter value as a function to convert the argument. i.e. if type=int then it will use int(arg) or if type=json.loads then json.loads(arg)
```

chatton: ah very cool!

scripts/dev/dockerutil.py
chatton: seems okay to ignore it in this case!

scripts/dev/release_image.py
```diff
@@ -0,0 +1,72 @@
...
+    parser.add_argument(
+        "--registry", help="The docker registry", type=str,
+    )
+    parser.add_argument(
+        "--labels", help="Labels for the new image", type=json.loads,
+    )
+    parser.add_argument(
+        "--image_type",
+        help="Type of image to be released",
+        choices=["mongodb-kubernetes-operator", "version-upgrade-hook"],
```
chatton, line 38: 👍

PR #147, opened 2020-08-25
Ensure consistent errors

chatton:
### All Submissions:

This PR updates the codebase to use `pkg/errors.Errorf` in place of `fmt.Errorf`
the format strings have been updated to use `%s` (the intended mechanism for handling errors in format strings)

### DIFF
--- a/test/e2e/mongodbtests/mongodbtests.go
+++ b/test/e2e/mongodbtests/mongodbtests.go
@@ -8,6 +8,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/pkg/errors"
+
 	"k8s.io/apimachinery/pkg/types"
 
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -200,7 +202,7 @@ func Connectivity(mdb *mdbv1.MongoDB, username, password string) func(t *testing
 			Username:      username,
 			Password:      password,
 		})); err != nil {
-			t.Fatal(fmt.Sprintf("Error connecting to MongoDB deployment: %+v", err))
+			t.Fatal(fmt.Sprintf("Error connecting to MongoDB deployment: %s", err))
 		}
 	}
 }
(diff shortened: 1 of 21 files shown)


test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -200,7 +202,7 @@ func Connectivity(mdb *mdbv1.MongoDB, username, password string) func(t *testing
 			Username:      username,
 			Password:      password,
 		})); err != nil {
-			t.Fatal(fmt.Sprintf("Error connecting to MongoDB deployment: %+v", err))
+			t.Fatal(fmt.Sprintf("Error connecting to MongoDB deployment: %s", err))
```
bznein, line 205: [q] is there a specific reason here we use `fmt.Sprintf`? A few lines below we have the same structure (`t.Fatal`) with inside a `errors.Errorf`

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -288,7 +290,7 @@ func IsReachableDuringWithConnection(mdb *mdbv1.MongoDB, interval time.Duration,
 					return
 				case <-time.After(interval):
 					if err := connectFunc(); err != nil {
-						t.Fatal(fmt.Sprintf("error reaching MongoDB deployment: %+v", err))
+						t.Fatal(fmt.Sprintf("error reaching MongoDB deployment: %s", err))
```
bznein, line 293: Same as the previous comment

test/e2e/replica_set_tls/replica_set_tls_test.go
```diff
@@ -30,7 +30,7 @@ func TestReplicaSetTLS(t *testing.T) {
 	}
 
 	if err := setup.CreateTLSResources(mdb.Namespace, ctx); err != nil {
-		t.Fatalf("Failed to set up TLS resources: %+v", err)
+		t.Fatalf("Failed to set up TLS resources: %s", err)
```
bznein, line 33: Seeing this and referring to my other two comments, maybe using `t.Fatalf` every time would be better? I feel we have a mixture of three different approaches at the moment:

`t.Fatalf`
`t.Fatal(fmt.Sprintf(...))`
`t.Fatal(errors.Errorf(...))`

bznein (approved):
LGTM but unless I misunderstood something we can achieve even more consistency in our error handling

test/e2e/replica_set_tls/replica_set_tls_test.go
chatton: yes I agree! I will update to use one. I think `t.Fatalf` is the most straightfoward, what do you think?

bznein: I think so! No need to nest function calls!

PR #148, opened 2020-08-26
add WithAgentFlags function

bznein:
This PR adds a `WithAgentFlags` function that sets up the `AGENT_FLAGS` environment variable from a slice of env var representing single agent startup flags. 

### All Submissions:

### DIFF
--- a/pkg/kube/container/containers.go
+++ b/pkg/kube/container/containers.go
@@ -152,3 +152,16 @@ func WithSecurityContext(context corev1.SecurityContext) Modification {
 		container.SecurityContext = &context
 	}
 }
+
+// WithAgentFlags takes a slice of envVars corresponding to
+// pairs of key-value agent startup flags and concatenates them
+// into a single string that is then passed as env variable AGENT_FLAGS
+func WithAgentFlags(envs ...corev1.EnvVar) Modification {
+	return func(container *corev1.Container) {
+		agentParams := ""
+		for _, variable := range envs {
+			agentParams += " -" + variable.Name + " " + variable.Value
+		}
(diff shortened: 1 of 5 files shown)


pkg/kube/container/containers.go
```diff
@@ -152,3 +152,16 @@ func WithSecurityContext(context corev1.SecurityContext) Modification {
...
+// into a single string that is then passed as env variable AGENT_FLAGS
+func WithAgentFlags(envs ...corev1.EnvVar) Modification {
+	return func(container *corev1.Container) {
+		agentParams := ""
+		for _, variable := range envs {
+			agentParams += " -" + variable.Name + " " + variable.Value
+		}
+		container.Env = envvar.MergeWithOverride(container.Env, []corev1.EnvVar{{Name: "AGENT_FLAGS", Value: agentParams}})
+	}
+}
```
chatton, line 167: nice implementation! I think however the containers package is not the correct place for this functionality. This specific function is now only exists for use with the agent (especially with the hard coded `AGENT_FLAGS` env var name)

To me this is something that we could leverage both in the community and enterprise operator, but I think it belongs in an `agents` or `agent` package, rather than `container`. 

To me it makes more sense to have the following structure

 `[]customType` -> `[]corev1.EnvVar` -> `container.WithEnvs(agentEnvs)`

rather than including this logic in this package.

What do you think?

chatton (commented):
Nice it looks great! 

I think maybe we should add at least one or 2 basic unit tests to ensure we're getting the expected format, but after that I think we're good to merge!

chatton (approved):

PR #160, opened 2020-08-31
Add Mongo Tester

chatton:
### All Submissions:

This PR introduces a `MongoTester` type which provides a new mechanism for writing tests.

In order to keep this PR small, the TLS tests have not been incorporated into the tests yet, that will happen in a future PR. And only 2 tests have been refactored. The remainder will be refactored in a different PR.

### DIFF
--- a/test/e2e/util/mongotester/mongotester.go
+++ b/test/e2e/util/mongotester/mongotester.go
@@ -91,6 +91,7 @@ func (m *Tester) hasAdminParameter(key string, expectedValue interface{}, tries 
 			<-time.After(10 * time.Second)
 			actualValue, err := m.getAdminSetting(key)
 			if err != nil {
+				t.Logf("Unable to get admin setting %s with error : %s", key, err)
 				continue
 			}
 			found = reflect.DeepEqual(expectedValue, actualValue)
--- a/test/e2e/util/mongotester/mongotester.go
+++ b/test/e2e/util/mongotester/mongotester.go
@@ -11,7 +11,6 @@ import (
 	f "github.com/operator-framework/operator-sdk/pkg/test"
 	"github.com/stretchr/testify/assert"
 	"go.mongodb.org/mongo-driver/bson"
-	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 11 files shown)


test/e2e/util/mongotester/mongotester.go
```diff
@@ -0,0 +1,237 @@
...
+// NOTE: Tls is not configured as the mechanism that the ClientOptions are merged only merge on non-nil
+// values, meaning we need to remove option that configures TLS from the list if we want to not use tls.
+// For now we can just explicitly pass WithTls() or WithoutTls() to configure TLS.
+func FromResource(t *testing.T, mdb mdbv1.MongoDB, opts ...*options.ClientOptions) (*Tester, error) {
+	var clientOpts []*options.ClientOptions
+	clientOpts = append(clientOpts, WithHosts(mdb.Hosts()))
+	t.Logf("Configuring hosts: %s for MongoDB: %s", mdb.Hosts(), mdb.NamespacedName())
+
+	users := mdb.Spec.Users
+	if len(users) == 1 {
```
bznein, line 43: [q] it's not clear to me why we make this check. Is it fundamental to have only one user? What happens if we have multiple?

test/e2e/util/mongotester/mongotester.go
```diff
@@ -0,0 +1,237 @@
...
+			_, err = collection.InsertOne(ctx, bson.M{"name": "pi", "value": 3.14159})
+			if err != nil && shouldSucceed {
+				return false, nil
+			}
+			if err == nil && !shouldSucceed {
+				t.Logf("Was successfully able to connect, when we should not have been able to!")
+				return false, nil
+			}
+
+			successfulCheck := (err != nil && !shouldSucceed) || (err == nil && shouldSucceed)
```
bznein, line 143: [q] Can this boolean ever be false?

```golang
err == nil && shouldSucceed -> successfulCheck == true
err == nil && !shouldSucceed -> we return at line 139
err != nil && shouldSucceed -> we return at line 136
err != nil && !shouldSucceed -> successfulCheck == true
```

bznein (approved):
LGTM with a couple of questions

test/e2e/util/mongotester/mongotester.go
chatton: The idea here was that if we have a test which configures multiple users, we would need to be more explicit with which user we want to attempt to connect as (via `WithScram(username,password)`)

Since all of our tests right now just have a single user, it's just a short hand for that! If we start adding tests with many users, maybe we can make some improvements here

test/e2e/util/mongotester/mongotester.go
chatton: yes you're right, I think we can just remove this last check entirely

PR #179, opened 2020-09-07
Allow for arbitrary Status subresource updates

chatton:
### All Submissions:

This PR introduces a mechanism to allow finer grained status updates.

We will now be able to update arbitrary fields of the status sub resource by creating new status options.

### DIFF
--- a/pkg/util/status/status.go
+++ b/pkg/util/status/status.go
@@ -3,13 +3,14 @@ package status
 import (
 	"context"
 
-	"k8s.io/apimachinery/pkg/runtime"
+	mdbv1 "github.com/mongodb/mongodb-kubernetes-operator/pkg/apis/mongodb/v1"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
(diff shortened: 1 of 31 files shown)


pkg/util/status/status.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+
+	mdbv1 "github.com/mongodb/mongodb-kubernetes-operator/pkg/apis/mongodb/v1"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type Option interface {
+	ApplyOption(mdb *mdbv1.MongoDB)
```
chatton, line 13: @rodrigovalin this small signature change let's us not need to worry about tracking the state of a single mdb instance, as we can just modify the one that was passed to this function.

rodrigovalin (commented):
I think I love the idea you introduced with the different options for the status. I still think we can improve it a bit if we have another pair session. I'm mostly thinking about improving how we use `GetResult`

rodrigovalin (approved):
LGTM!

PR #186, opened 2020-09-09
CLOUDP-71334: Fix MongoDB 4.4 Scaling Up

chatton:
### All Submissions:

With MongoDB 4.4 it is no longer possible to scale up more than one member at a time (besides the first time creation of the deployment)

This PR brings the scale package from the enterprise code base and implements it for the community MongoDB resource

### DIFF
--- a/pkg/apis/mongodb/v1/mongodb_types.go
+++ b/pkg/apis/mongodb/v1/mongodb_types.go
@@ -220,15 +220,15 @@ type MongoDB struct {
 	Status MongoDBStatus `json:"status,omitempty"`
 }
 
-func (m MongoDB) DesiredReplicaSetMembers() int {
+func (m MongoDB) DesiredReplicas() int {
 	return m.Spec.Members
 }
 
-func (m MongoDB) CurrentReplicaSetMembers() int {
+func (m MongoDB) CurrentReplicas() int {
 	return m.Status.Members
 }
 
-func (m *MongoDB) MembersThisReconciliation() int {
+func (m *MongoDB) ReplicasThisReconciliation() int {
 	return scale.ReplicasThisReconciliation(m)
 }
 
--- a/pkg/apis/mongodb/v1/mongodb_types.go
+++ b/pkg/apis/mongodb/v1/mongodb_types.go
@@ -24,6 +24,8 @@ type Phase string
 
 const (
 	Running Phase = "Running"
+	Failed  Phase = "Failed"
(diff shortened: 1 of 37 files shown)


pkg/util/scale/scale.go
```diff
@@ -0,0 +1,43 @@
...
+
+func IsStillScaling(replicaSetScaler ReplicaSetScaler) bool {
+	return ReplicasThisReconciliation(replicaSetScaler) != replicaSetScaler.DesiredReplicaSetMembers()
+}
+
+func isScalingDown(replicaSetScaler ReplicaSetScaler) bool {
+	return replicaSetScaler.DesiredReplicaSetMembers() < replicaSetScaler.CurrentReplicaSetMembers()
+}
+
+// AnyAreStillScaling reports true if any of one the provided members is still scaling
```
rodrigovalin, line 35: Is this function used?

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -218,6 +220,18 @@ type MongoDB struct {
...
+func (m MongoDB) DesiredReplicaSetMembers() int {
+	return m.Spec.Members
+}
+
+func (m MongoDB) CurrentReplicaSetMembers() int {
+	return m.Status.Members
+}
+
+func (m *MongoDB) MembersThisReconciliation() int {
+	return scale.ReplicasThisReconciliation(m)
```
rodrigovalin, line 232: `Members` or `Replicas` ? What about agreeing on 1 term?

rodrigovalin (commented):
LGTM, straightforward implementation! ❤️ 

Can we agree on the term used to refer to the "elements of the set of mongodb on a replicaset" as "replicas" or "members"? Or do you think both terms should be used in different contexts?

pkg/util/scale/scale.go
chatton: good catch, this is actually used in the enterprise operator for the ShardedCluster implementation, this package will be imported there.

pkg/apis/mongodb/v1/mongodb_types.go
rodrigovalin: I vote for Replicas, replicas everywhere

chatton:
@rodrigovalin let me know if you're happy with the rename!

rodrigovalin (approved):
LGTM

PR #194, opened 2020-09-14
Make security a mandatory crd field

### LINKED ISSUE #178: CRD changes don't get noticed by operator
If you set resource limits as described in #176 and change the values operator doesn't notice any changes.
Expected behaviour would be for it to apply the new values.

I haven't tested in detail if that is true for everything.
Top level specs like version changes work, below spec.containers it doesn't get noticed any more apparently.

chatton:
### All Submissions:

closes #178

Users is a mandatory field under `spec.security.users` but `spec.security` itself was not mandatory, meaning if you remove security the CRD is successfully validated and there will be errors when operator attempts to update the resource during reconciliation.


rodrigovalin (approved):
LGTM

PR #196, opened 2020-09-14
(DOCSP-11424): Document SCRAM Auth

melissamahoney-mongodb:
This PR:
* Breaks the sections that were in the README into separate docs .md files
* Rearranges those docs .md files under one /docs directory
* Updates the README to point to the /docs directory
* Creates a /docs/README to act as a TOC
* Creates a new page for Creating Database Users

Staging:
* [Readme](https://github.com/melissamahoney-mongodb/mongodb-kubernetes-operator/blob/DOCSP-11424/README.md)
* [Install and Upgrade](https://github.com/melissamahoney-mongodb/mongodb-kubernetes-operator/blob/DOCSP-11424/docs/install-upgrade.md)
* [Deploy and Configure](https://github.com/melissamahoney-mongodb/mongodb-kubernetes-operator/blob/DOCSP-11424/docs/deploy-configure.md)
* [Create a DB User](https://github.com/melissamahoney-mongodb/mongodb-kubernetes-operator/blob/DOCSP-11424/docs/users.md)
* [Secure Connections w/ TLS](https://github.com/melissamahoney-mongodb/mongodb-kubernetes-operator/blob/DOCSP-11424/docs/secure.md)


docs/users.md
```diff
@@ -0,0 +1,86 @@
...
+     apiVersion: v1
+     kind: Secret
+     metadata:
+       name: <db-user-secret>  # corresponds to spec.users.passwordSecretRef.name in the MongoDB CRD
+     type: Opaque
+     stringData:
+       password: <my-plain-text-password> # corresponds to spec.users.passwordSecretRef.key in the MongoDB CRD
+     ...
+     ```
+1. Update `metadata.name` with the name of the secret and `stringData.password` with the user's password.
```
jwilliams-mongo, line 22: Offer clearer guidance here that you can provide anything you want for  <db-user-secret>. I think users might ask what the name of the secret is without realizing they can provide any value here.

docs/users.md
```diff
@@ -0,0 +1,86 @@
+# Create a Database User #
+
+You can create a MongoDB database user to authenticate to your MongoDB resource using SCRAM. First, [create a Kubernetes secret](#create-a-user-secret) for the new user's password. Then, [modify and apply the MongoDB CRD](#modify-the-mongodb-crd).
```
jwilliams-mongo, line 3: consider linking to the Manual page on SCRAM here.

docs/users.md
```diff
@@ -0,0 +1,86 @@
...
+
```
jwilliams-mongo, line 4: Question about behavior here: When you create one SCRAM user, you will no longer be able to perform any actions unauthenticated, correct? I think we can be a little clearer about this here.

README.md
```diff
@@ -265,17 +34,17 @@ The MongoDB Community Kubernetes Operator supports the following features:
 - Use of any of the available [Docker MongoDB images](https://hub.docker.com/_/mongo/)
 - Clients inside the Kubernetes cluster can connect to the replica set (no external connectivity)
 - TLS support for client-to-server and server-to-server communication
+- Creating users with SCRAM-SHA authentication
```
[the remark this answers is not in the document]
melissamahoney-mongodb: Ah good point, I just copy/pasted from below. Updated.

docs/deploy-configure.md
```diff
@@ -0,0 +1,85 @@
+# Deploy and Configure a MongoDB Resource #
```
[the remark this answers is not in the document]
melissamahoney-mongodb: DIdn't realize I could have multiple readmes, good idea. Added.

docs/users.md
melissamahoney-mongodb: There are a handful of commands you can run without being authenticated, like the native commands (like `listFiles()` ) and a few database commands. I tried `db.runCommand({connectionStatus : 1})` and `isMaster`, which both worked while unauthenticated. But I can't find a definitive list of them, and I don't think the ones that work do very much. Let me know if you want me to elaborate more in the topic. And maybe @chatton  has an opinion :)

docs/README.md
```diff
@@ -0,0 +1,10 @@
+# MongoDB Community Kubernetes Operator Documentation #
+
+# Table of Contents
```
jwilliams-mongo, line 3:
```suggestion
## Table of Contents
```
nit: this should be an H2.

jwilliams-mongo (approved):
LGTM % one nit

docs/users.md
```diff
@@ -0,0 +1,87 @@
...
+   kubectl delete secret <db-user-secret> --namespace <my-namespace>
+   ```
+
+## Next Steps
+
+- To authenticate to your MongoDB resource, run the following command:
+   ```
+   mongo "mongodb://<mongodb-resource-metadata.name>-svc.<my-namespace>.svc.cluster.local:27017/?replicaSet=<replica-set-name>" --username <username> --password <password> --authenticationDatabase <authentication-database>
+   ```
+- To change a user's password, create and apply a new secret with a `metadata.name` that is the same as the name specified in `passwordSecretRef.name` of the MongoDB CRD. The Operator automatically re-generates the SCRAM credentials.
```
chatton, line 87: [nit] `create a secret` already indicates that the kubernetes secret resource is being created, but here you are referring to creating a new resource definition that contains the yaml for the new password.

I think we should phrase this as `create and apply a new secret resource definition` or just `create a new secret as described above`

chatton: [tiny-nit] The line `The Operator automatically re-generates the SCRAM credentials.` sounds a little off to me, I think we can just say, `The Operator will automatically generate new credentials`

docs/users.md
chatton: I think it should be clear enough to link to the SCRAM docs like we do here, as everything written there will apply, I don't think we need to clarify any more about this, I don't feel too strongly about this though.

chatton (commented):
Great work @melissamahoney-mongodb , I love the structure you added to the docs here.

I had just had one or two small comments!

docs/users.md
```diff
@@ -0,0 +1,87 @@
...
+             db: <role-2-database>
+   ...
+   ```
+1. Save the file.
+1. Apply the updated MongoDB resource definition:
+
+   ```
+   kubectl apply -f <mongodb-crd>.yaml --namespace <my-namespace>
+   ```
+1. After the MongoDB resource is running, securely store the user's password and then delete the user secret:
```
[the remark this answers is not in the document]
melissamahoney-mongodb: Updated.

docs/users.md
melissamahoney-mongodb: Gotcha. Updated.

melissamahoney-mongodb:
Thanks for the review @chatton ! Ready for another look.

docs/users.md
```diff
@@ -0,0 +1,88 @@
...
+     ```
+1. Update the value of `metadata.name` with any name for this secret.
+1. Update the value of `stringData.password` with the user's password.
+1. Save the secret with a `.yaml` file extension.
+1. Apply the secret in Kubernetes:
+   ```
+   kubectl apply -f <db-user-secret>.yaml --namespace <my-namespace>
+   ```
+
+## Modify the MongoDB CRD
```
chatton, line 30: Ah! Sorry I think this may have been my fault, in this case CRD refers to a [kubernetes CRD](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) which is what we provide, the user will modify a MongoDB resource definition, but not the CRD itself. I think maybe most straight forward way to phrase this is `Modify the MongoDB Resource`

I think `MongoDB resource` is a good term to refer to individual instances of the `Custom Resource` which are described via the `Custom Resource Definition (CRD)`

melissamahoney-mongodb: Ah, I think I wasn't quite clear on that distinction myself. Thanks for the explanation! Updated.

chatton (approved):
LGTM thanks for the great work on this!

PR #201, opened 2020-09-18
Change default permissions for volumes created from secret

bznein:
This PR changes the default permissions on volumes created from a secret. This is needed for another work in the enterprise operator. It will still be possible to specify different modes through the yaml file (when the merging bug is fixed)

### All Submissions:


chatton (commented):
Looks good to me, as discussed in person we should add some unit tests to check the merging logic around this field.

bznein:
@chatton as discussed in person, putting a simple unit test for now and then I'll add a more proper tests around merging when fixing it. Ready for another review when you have time!

chatton (approved):
LGTM! 👍

PR #206, opened 2020-09-28
Disallow TLS disabling by checking last successful saved configuration

rodrigovalin:


pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -68,6 +68,15 @@ type MongoDBSpec struct {
 	AdditionalMongodConfig MongodConfiguration `json:"additionalMongodConfig,omitempty"`
 }
 
+func (m MongoDBSpec) ToJson() []byte {
+	b, err := json.Marshal(m)
+	if err != nil {
```
chatton, line 73: I think we should at least log this error, we won't have any insight into this otherwise

chatton (approved):
I just had one comment, no need to second pass LGTM

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -68,13 +68,24 @@ type MongoDBSpec struct {
...
+
+func FromJSON(jsonString string) (MongoDBSpec, error) {
+	m := MongoDBSpec{}
+	err := json.Unmarshal([]byte(jsonString), &m)
+	if err != nil {
+		return MongoDBSpec{}, err
+	}
+
+	return m, nil
+
```
chatton, line 88: [q] do these functions actually provide us any value? (applies to the ToJson function aswell)

We are just wrapping one function call, the calling. code will look almost the same anyway

```golang
spec, err := mdbv1.FromJson(jsonString)
if err != nil {
   // Handle error
}

// use spec
```

vs 

```golang
spec := MongoDBSpec{}
err := json.Unmarshal([]byte(jsonString), &spec)
if err != nil {
     // handle error
}
// use spec
```

pkg/controller/mongodb/replica_set_controller.go
```diff
@@ -331,13 +342,18 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
+		// If we could get all the way up to here, then save the current spec as the last
+		// successfully applied.
+		lastSuccessfulConfigurationSaved, err := mdb.Spec.ToJson()
+		if err != nil {
+			r.log.Error("Error converting Spec to Json: %s", err)
+		}
+		annotations = map[string]string{
+			lastSuccessfulConfiguration: string(lastSuccessfulConfigurationSaved),
+		}
+		r.setAnnotations(mdb.NamespacedName(), annotations)
```
chatton, line 355: this function can return an error! We should handle it

pkg/controller/mongodb/replica_set_controller.go
```diff
@@ -331,13 +342,18 @@ func (r *ReplicaSetReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
-	}
-	r.setAnnotations(mdb.NamespacedName(), annotations)
-
-	if err != nil {
+	if err == nil {
+		// If we could get all the way up to here, then save the current spec as the last
+		// successfully applied.
+		lastSuccessfulConfigurationSaved, err := mdb.Spec.ToJson()
+		if err != nil {
+			r.log.Error("Error converting Spec to Json: %s", err)
```
chatton, line 350: We should exit early here, otherwise we are going to attempt to add a junk value as an annotation!

pkg/controller/mongodb/replica_set_controller.go
```diff
@@ -541,6 +557,24 @@ func getCurrentAutomationConfig(getUpdater secret.GetUpdater, mdb mdbv1.MongoDB)
...
+	lastSuccessfulConfigurationSaved, ok := mdb.Annotations[lastSuccessfulConfiguration]
+	if !ok {
+		// First version of Spec, no need to validate
+		return nil
+	}
+
+	last, err := mdbv1.FromJSON(lastSuccessfulConfigurationSaved)
+	if err != nil {
+		r.log.Debugf("Could not Unmarshal annotation into MongoDBSpec: ", err)
+		return nil
```
chatton, line 573: why return nil here? If we have an annotation it should be able to be correctly read

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -209,6 +220,29 @@ func Scale(mdb *mdbv1.MongoDB, newMembers int) func(*testing.T) {
...
+	return TLS(mdb, false)
+}
+
+// EnableTLS changes the tls.enabled attribute to false
+func EnableTLS(mdb *mdbv1.MongoDB) func(*testing.T) {
+	return TLS(mdb, true)
+}
+
+// TLS changes the tls.enabled attribute to false
+func TLS(mdb *mdbv1.MongoDB, enabled bool) func(*testing.T) {
```
chatton, line 234: [nit] This function can be unexported

chatton (commented):
Looks great! I just had a few comments/suggestions

pkg/controller/mongodb/replica_set_controller.go
rodrigovalin: This function is supposed to validate the old and new specs; if for any reason the old spec can't be unmarshalled, then we assume that the actual JSON contents are still valid; for instance if the JSON document was saved with a previous version of the operator. In those cases, we must at least wait for the new spec to be saved at least once, before assuming its contents are invalid.

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -222,16 +222,16 @@ func Scale(mdb *mdbv1.MongoDB, newMembers int) func(*testing.T) {
...
-	return TLS(mdb, true)
+	return tls(mdb, true)
 }
 
-// TLS changes the tls.enabled attribute to false
-func TLS(mdb *mdbv1.MongoDB, enabled bool) func(*testing.T) {
+// tls changes the tls.enabled attribute to false
+func tls(mdb *mdbv1.MongoDB, enabled bool) func(*testing.T) {
 	return func(t *testing.T) {
 		t.Log("Setting TLS to disabled")
```
bznein: I guess the comment and the log call are outdated? This doesn't necessarily disable TLS, it depends on the `enabled` parameter

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -209,6 +220,29 @@ func Scale(mdb *mdbv1.MongoDB, newMembers int) func(*testing.T) {
 	}
 }
 
+// DisableTLS changes the tls.enabled attribute to false
+func DisableTLS(mdb *mdbv1.MongoDB) func(*testing.T) {
+	return tls(mdb, false)
+}
+
+// EnableTLS changes the tls.enabled attribute to false
```
bznein, line 228:
```suggestion
// EnableTLS changes the tls.enabled attribute to true
```

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -100,6 +100,17 @@ func MongoDBReachesRunningPhase(mdb *mdbv1.MongoDB) func(t *testing.T) {
 	}
 }
 
+// MongoDBReachesFailed ensure the MongoDB resource reaches the Running phase
```
bznein, line 103:
```suggestion
// MongoDBReachesFailed ensure the MongoDB resource reaches the Failed phase
```

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -209,6 +220,29 @@ func Scale(mdb *mdbv1.MongoDB, newMembers int) func(*testing.T) {
...
+
+// EnableTLS changes the tls.enabled attribute to true.
+func EnableTLS(mdb *mdbv1.MongoDB) func(*testing.T) {
+	return tls(mdb, true)
+}
+
+// tls function configures the security.tls.enabled attribute.
+func tls(mdb *mdbv1.MongoDB, enabled bool) func(*testing.T) {
+	return func(t *testing.T) {
+		t.Log("Setting TLS to disabled")
```
bznein, line 236: [nit] this log line is still outdated

bznein (approved):
LGTM

PR #209, opened 2020-09-29
Fix broken docs link

bznein:
Since moving some documentation inside the `docs` directory, some links became 404, so I fixed the path for them.

I also removed the link to the `operator-sdk` e2e-test docs, as they were moved to a `legacy` section of the website but now have been completely removed from it.

Unrelated but needed to submit the PR: a fix to the pre-commit hook (`mapfile` on an empty array returns an array with one empty element thus breaking the go linter)

Auto-formatter also removed a bunch of whitespace, which I don't think should be an issue but happy to revert if needed somewhere

### All Submissions:


rodrigovalin (approved):
LGTM

Not really sure why we need the change on the pre-commit hook?

PR #282, opened 2021-01-14
Remove testrunner

chatton:
### All Submissions:

This PR removes the Test Runner. Instead, now the operator deployment happens inside the test pod. e2e.py now passes the configuration as env variables to the test pod directly.


test/e2e/setup/setup.go
```diff
@@ -115,3 +128,153 @@ func GeneratePasswordForUser(mdbu mdbv1.MongoDBUser, ctx *f.Context, namespace s
 
 	return password, f.Global.Client.Create(context.TODO(), &passwordSecret, &f.CleanupOptions{TestContext: ctx})
 }
+
+func deployOperator(c f.FrameworkClient) error {
```
chatton, line 132: moved from testrunner/main.go

values now come from `testConfig` instead of `flags`. These are populated through env vars passed in `e2e.py`

scripts/ci/create_kind_cluster.sh
```diff
@@ -1,3 +1,6 @@
 #!/bin/sh
 
 kind create cluster --kubeconfig "${KUBECONFIG}"
+
+echo "Creating CRDs"
+kubectl apply -f deploy/crds/*crd.yaml
```
chatton, line 6: this can be moved into e2e.py in a future PR

scripts/dev/e2e.py
```diff
@@ -129,30 +108,89 @@ def build_and_push_version_upgrade_hook(repo_url: str, tag: str, path: str) -> N
...
-def create_test_runner_pod(args: argparse.Namespace) -> None:
-    """
-    create_test_runner_pod creates the pod which will run all of the tests.
-    """
-    dev_config = load_config(args.config_file)
+def create_test_pod(args: argparse.Namespace, dev_config: DevConfig) -> None:
     corev1 = client.CoreV1Api()
-    pod_body = _get_testrunner_pod_body(args)
-
+    test_pod = {
```
chatton, line 125: this was originally in operator-sdk-test.yaml

.evergreen.yml
```diff
@@ -220,18 +220,6 @@ tasks:
           image_type: versionhook
           expire_after: 48h
 
-  - name: build_testrunner_image
```
alyacb, line 223: I think this may decrease evergreen build times! 🎉

scripts/dev/e2e.py
```diff
@@ -17,26 +17,22 @@
...
+    return load_yaml_from_file("deploy/operator/role.yaml")
 
 
-def _load_testrunner_role_binding() -> Dict:
-    return load_yaml_from_file("deploy/testrunner/role_binding.yaml")
+def _load_test_role_binding() -> Dict:
+    return load_yaml_from_file("deploy/operator/role_binding.yaml")
 
 
-def _load_testrunner_cluster_role_binding() -> Dict:
```
alyacb, line 35: Do we no longer need to load the cluster role binding here? I see that the old changes loaded the cluster role binding but not the cluster role- is that why?

scripts/dev/e2e.py
```diff
@@ -179,47 +217,12 @@ def wait_for_pod_to_be_running(
         lambda: corev1.read_namespaced_pod(name, namespace),
         lambda pod: pod.status.phase == "Running",
         sleep_time=5,
-        timeout=90,
+        timeout=180,
```
alyacb, line 220: Why did you increase the timeout here?

scripts/dev/e2e.py
```diff
@@ -129,30 +108,89 @@ def build_and_push_version_upgrade_hook(repo_url: str, tag: str, path: str) -> N
...
+                            "value": f"{dev_config.repo_url}/{dev_config.operator_image}:{args.tag}",
+                        },
+                        {
+                            "name": "TEST_NAMESPACE",
+                            "value": dev_config.namespace,
+                        },
+                        {
+                            "name": "VERSION_UPGRADE_HOOK_IMAGE",
+                            # TODO: this needs to come from somewhere else
+                            "value": "quay.io/mongodb/mongodb-kubernetes-operator-version-upgrade-post-start-hook:1.0.2",
```
alyacb, line 158: If this doesn't need to be specified as an environment variable, maybe the VERSION_UPGRADE_HOOK_IMAGE could just live in some yaml/ as some constant. Or did you have something else in mind?

test/e2e/setup/test_config.go
```diff
@@ -0,0 +1,31 @@
...
+	versionUpgradeHookImage string
+	clusterWide             bool
+	performCleanup          bool
+}
+
+func loadTestConfigFromEnv() testConfig {
+	return testConfig{
+		namespace:               envvar.GetEnvOrDefault(testNamespaceEnvName, "default"),
+		operatorImage:           envvar.GetEnvOrDefault(operatorImageEnvName, "quay.io/mongodb/community-operator-dev:latest"),
+		versionUpgradeHookImage: envvar.GetEnvOrDefault(versionUpgradeHookEnvName, "quay.io/mongodb/mongodb-kubernetes-operator-version-upgrade-post-start-hook:1.0.2"),
```
alyacb, line 27: I see that there is a default value here for the `versionUpgradeHookImage`- why is it hard-coded in e2e.py:157?

alyacb (commented):
Looks good! I just have a few questions, but I'm really excited to see the test-runner go away!  I think this will speed up and simplify e2e testing quite a bit.

scripts/dev/e2e.py
chatton: The e2e-test pod takes longer to get started than the testrunner did!

scripts/dev/e2e.py
chatton: when we need to make changes, it won't always be `1.0.2` it will be the `version_id` of an individual evergreen patch, or whatever tag we want if building locally. For now it isn't undergoing any changes, but this will need to be configurable.

test/e2e/setup/test_config.go
chatton: I will add a separate smaller PR to configure this (reason mentioned in the other comment!)

scripts/dev/e2e.py
alyacb: That makes sense, thank you for explaining!

alyacb (approved):
LGTM!

PR #286, opened 2021-01-19
CLOUDP-79918: Support custom MongoDB roles

alyacb:
### All Submissions:

* [X] Have you opened an Issue before filing this PR?

JIRA issue: https://jira.mongodb.org/browse/CLOUDP-79918

Adds CRD support for custom user-defined roles.


deploy/crds/mongodb.com_mongodb_crd.yaml
```diff
@@ -50,14 +50,95 @@ spec:
               description: Members is the number of members in the replica set
               type: integer
             replicaSetHorizons:
-              description: Add this parameter and values if you need your database
```
alyacb, line 53: I'm not sure why this description changed when it was auto-generated. Should I change it back?

deploy/crds/mongodb.com_mongodb_crd.yaml
```diff
@@ -50,14 +50,95 @@ spec:
...
+                      - db
+                      - name
+                      type: object
+                    type: array
+                required:
+                - db
+                - privileges
+                - role
                 type: object
               type: array
```
alyacb, line 143: The `type:array` field for roles here is initially set to object by the operator-sdk generate command. I'm not sure why, since roles is defined as an array field, but I had to manually change this to array to get it to work. Ideally I'd prefer not to introduce new peculiarities to the crd auto-generation, but I'm not sure why this happened.

deploy/crds/mongodb.com_mongodb_crd.yaml
```diff
@@ -169,10 +250,10 @@ spec:
                       type: object
                     type: array
                   scramCredentialsSecretName:
-                    description: ScramCredentialsSecretName appended by string "scram-credentials" is the name of the secret object
-                      created by the mongoDB operator for storing SCRAM credentials
+                    description: ScramCredentialsSecretName appended by string "scram-credentials"
+                      is the name of the secret object created by the mongoDB operator
+                      for storing SCRAM credentials
                     type: string
-                    pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
```
alyacb, line 175: Should I add this back? The regex was not accepted by the auto-generation tool, so I commented it out.

deploy/crds/mongodb.com_v1_mongodb_custom_role.yaml
```diff
@@ -0,0 +1,46 @@
+---
```
alyacb, line 1: I added a file to illustrate how one could potentially define a custom role via the CRD- not sure if we need one, or if having this explained in the docs would be sufficient. Thoughts?

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -72,12 +72,58 @@ type MongoDBSpec struct {
 	// configuration file: https://docs.mongodb.com/manual/reference/configuration-options/
 	// +kubebuilder:validation:Type=object
 	AdditionalMongodConfig MongodConfiguration `json:"additionalMongodConfig,omitempty"`
+
+	// User-specified custom MongoDB roles that should be configured in the deployment.
+	// +optional
+	Roles []AutomationConfigRole `json:"roles,omitempty"`
```
alyacb, line 78: Unfortunately, the `AutomationConfigRole` struct has a counterpart in `automationconfig` called `CustomRole`, which has the same structure. The reason this was necessary is because `operator-sdk generate k8s` only generates structs that are defined in this file, but the same struct is needed in the automation config. Since `mongodb_types` imports `automationconfig`, I can't import `mongodb_types` in `automationconfig` (lest I introduce a circular dependency).

Is there a way to avoid this duplication?

test/e2e/e2eutil.go
```diff
@@ -181,6 +181,10 @@ func NewTestMongoDB(name string, namespace string) (mdbv1.MongoDB, mdbv1.MongoDB
 							DB:   "admin",
 							Name: "clusterAdmin",
 						},
+						{
+							DB:   "admin",
+							Name: "userAdmin",
```
alyacb, line 186: This was necessary in order to allow the new e2e test to see roles (none of the other roles for this user have the `rolesInfo` privilege). It seemed easier than adding a flag/new parameter to this function. It might make more sense to try to use a custom role to do this. Thoughts?

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -276,6 +290,56 @@ func Connect(mdb *mdbv1.MongoDB, opts *options.ClientOptions) error {
...
+				{"rolesInfo", 1},            // nolint
+				{"showPrivileges", true},    // nolint
+				{"showBuiltinRoles", false}, // nolint
+			}).Decode(&result)
+		if err != nil {
+			t.Fatal(err)
+			return false, nil
+		}
+
+		jsonRoles, err := json.Marshal(result["roles"])
```
alyacb, line 326: There should be a better way to do this than convert to json and back (since we have the info in bson). I'll try to use something like a `map[string][]CustomRole` in the `Decode()` call, since the result is a map with a roles field that contains all the result roles. Other ideas are welcome.

pkg/automationconfig/automation_config.go
```diff
@@ -140,6 +141,31 @@ type Auth struct {
...
+	Roles                      []Role                      `json:"roles"`
+	AuthenticationRestrictions []AuthenticationRestriction `json:"authenticationRestrictions,omitempty"`
+}
+
+type Privilege struct {
+	Resource Resource `json:"resource"`
+	Actions  []string `json:"actions"`
+}
+
+type Resource struct {
```
alyacb, line 157: The definition for a resource can be found here: https://docs.mongodb.com/manual/reference/resource-document/
It takes one of the following forms:
```
{db: string, collection: string}
{cluster: true}
{anyResource: true}
```

However, if either db or collection are set to "", that has a special meaning that indicates "all" databases/collections. Unfortunately in this case, the `omitempty` struct tag omits empty strings. If I mark db/collection as 'omitempty', the case where db and collection are set to "" gets sent to the agent as `{}`. Otherwise, if a user just tries to specify `{cluster: true}`, the agent will get `{db: "", collection: "", cluster: true}`, which is considered invalid.

Is there some way to omit these fields only when they are set to nil, for instance? That way we could preserve both the case where `{db: "", collection: ""}` and `{cluster: true}`.

test/e2e/mongodbtests/mongodbtests.go
alyacb: Fixed!

pkg/automationconfig/automation_config.go
chatton: yes, we can achieve this using a string reference

```golang
Collection  *string `json:"collection,omitempty"`
```

omitempty will leave out the field if the underlying type has the zero value, and `""` is not the zero value for a pointer.

This is a pretty idiomatic way of dealing with optional fields when the normal zero value has a special meaning

pkg/automationconfig/automation_config_builder.go
```diff
@@ -191,3 +191,90 @@ func withFCV(fcv string) func(*Process) {
...
+	cluster bool,
+	anyResource bool,
+	actions []string) *CustomRoleBuilder {
+
+	res := Resource{
+		DB:         db,
+		Collection: collection,
+	}
+	if cluster {
+		res = Resource{
```
chatton, line 237: we lose the values for `DB` and `Collection` with this re-assignment right?

Can we just do 

```golang
res := Resource{
		DB:         db,
		Collection: collection,
                Cluster: cluster,
                AnyResource: anyResource,
	}
```

pkg/apis/mongodb/v1/zz_generated.deepcopy.go
```diff
@@ -5,6 +5,7 @@
 package v1
```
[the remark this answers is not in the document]
bznein: It's not accidental, can you remind me tomorrow so I can add context? Thanks!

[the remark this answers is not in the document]
bznein: https://jira.mongodb.org/browse/CLOUDP-70457

deploy/crds/mongodb.com_v1_mongodb_custom_role.yaml
bznein: I am absolutely in favour of having examples for any new feature we implement

chatton: the more examples the better!

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -276,6 +290,49 @@ func Connect(mdb *mdbv1.MongoDB, opts *options.ClientOptions) error {
 	})
 }
 
+func ConnectAndVerifyExpectedCustomRoles(mdb *mdbv1.MongoDB, database, username, password string, expectedRoles []automationconfig.CustomRole) func(t *testing.T) {
```
chatton, line 293: it looks like a lot of this functionality we actually already have! If you look at the `tester, err := FromResource(t, mdb)` call, that function call configures the underlying golang client with the correct SCRAM credentials for that resource, and there is a method [hasAdminParameter](https://github.com/mongodb/mongodb-kubernetes-operator/blob/5fa5815cd5e458121c4c1e99361fb45cea5c703d/test/e2e/util/mongotester/mongotester.go#L138) which can be used to check for a specific field in the `admin` db. 

The end result would look something like 

```golang
t.Run("Custom Role was created ", tester.VerifyRoles(&mdb, roles))
```

since the client configuration should already be done.

test/e2e/replica_set_custom_role/replica_set_custom_role_test.go
```diff
@@ -0,0 +1,69 @@
...
+	t.Run("AutomationConfig has the correct version", mongodbtests.AutomationConfigVersionHasTheExpectedVersion(&mdb, 1))
+
+	roles := []automationconfig.CustomRole{
+		automationconfig.NewCustomRoleBuilder().
+			WithRole("testRole").
+			WithDB("admin").
+			AddPrivilege("test", "", false, false, []string{
+				"collStats", "createCollection", "dbStats", "find", "viewRole",
+			}).
+			Build(),
```
chatton, line 64: it might be a bit more convinent to have a function which accepts a MongoDB.Spec.Roles and returns the AutomationConfig version (will leave up to you!)

chatton (commented):
This is looking great! I think we can actually re-use a lot of existing code for the tests you added

deploy/crds/mongodb.com_mongodb_crd.yaml
alyacb: Done! Note that formatting still changed a bit because of the way the comment was formatted.

pkg/automationconfig/automation_config.go
alyacb: Thank you! Using a string reference everywhere instead worked like a charm. This is done now.

alyacb: Thank you! Using a string reference everywhere instead worked like a charm. This is done now.

pkg/apis/mongodb/v1/mongodb_types.go
```diff
@@ -72,12 +72,58 @@ type MongoDBSpec struct {
...
 }
 
 // ReplicaSetHorizonConfiguration holds the split horizon DNS settings for
 // replica set members.
 type ReplicaSetHorizonConfiguration []automationconfig.ReplicaSetHorizons
 
+// AutomationConfigRole defines a custom MongoDB role.
+type AutomationConfigRole struct {
```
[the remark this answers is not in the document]
alyacb: I renamed this to CustomRole.

pkg/automationconfig/automation_config_builder.go
alyacb: My intention when writing that was to omit the fields (since the accepted format does not allow db/collection to coexist with either of the boolean flags). However, with the string references and the refactoring to use a converter this works. Done!

deploy/crds/mongodb.com_v1_mongodb_custom_role.yaml
```diff
@@ -0,0 +1,46 @@
...
+  security:
+    authentication:
+      modes: ["SCRAM"]
+  roles:
+    - role: testRole
+      db: admin
+      privileges:
+        - resource:
+            db: "test"
+            collection: ""
```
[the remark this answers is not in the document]
alyacb: I added a few comments to this file to explain fields.

deploy/crds/mongodb.com_v1_mongodb_custom_role.yaml
```diff
@@ -0,0 +1,46 @@
+---
+apiVersion: mongodb.com/v1
+kind: MongoDB
+metadata:
+  name: example-mongodb
```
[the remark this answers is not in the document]
alyacb: Done! It is now custom-role-mongodb

pkg/automationconfig/automation_config_builder.go
```diff
@@ -191,3 +191,90 @@ func withFCV(fcv string) func(*Process) {
...
+
+	b.privileges = append(b.privileges, Privilege{
+		Resource: res,
+		Actions:  actions,
+	})
+	return b
+}
+
+// AddRole adds an inherited role to a custom role.
+func (b *CustomRoleBuilder) AddRole(role string, db string) *CustomRoleBuilder {
```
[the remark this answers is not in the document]
alyacb: I removed the builder in the latest change, so this is no longer applicable.

pkg/automationconfig/automation_config_builder.go
```diff
@@ -191,3 +191,90 @@ func withFCV(fcv string) func(*Process) {
...
+func (b *CustomRoleBuilder) AddAuthenticationRestriction(clientSource []string, serverAddress []string) *CustomRoleBuilder {
+	restriction := AuthenticationRestriction{
+		ClientSource:  clientSource,
+		ServerAddress: serverAddress,
+	}
+	if b.authenticationRestrictions == nil {
+		b.authenticationRestrictions = []AuthenticationRestriction{restriction}
+	} else {
+		b.authenticationRestrictions = append(b.authenticationRestrictions, restriction)
+	}
```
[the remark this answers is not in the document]
alyacb: That's cool! Nil check has been removed (in the converter, as the builder is no more).

pkg/controller/mongodb/replica_set_controller.go
```diff
@@ -551,6 +551,45 @@ func (r ReplicaSetReconciler) validateUpdate(mdb mdbv1.MongoDB) error {
...
+
+	return func(config *automationconfig.AutomationConfig) {
+		config.Roles = []automationconfig.CustomRole{}
+
+		for _, role := range mdb.Spec.Roles {
+			b := automationconfig.NewCustomRoleBuilder().WithRole(role.Role).WithDB(role.DB)
+
+			// Add privileges.
+			for _, privilege := range role.Privileges {
+				b = b.AddPrivilege(
```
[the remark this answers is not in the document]
alyacb: You are right! Builder is gone now, but I'll keep it in mind.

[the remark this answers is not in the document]
alyacb: You are right! Builder is gone now, but I'll keep it in mind.

pkg/controller/mongodb/replica_set_controller.go
```diff
@@ -551,6 +551,45 @@ func (r ReplicaSetReconciler) validateUpdate(mdb mdbv1.MongoDB) error {
...
+				)
+			}
+
+			// Add roles.
+			for _, dbRole := range role.Roles {
+				b = b.AddRole(dbRole.Name, dbRole.DB)
+			}
+
+			// Add authentication restrictions (if any).
+			if role.AuthenticationRestrictions != nil {
```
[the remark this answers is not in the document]
alyacb: Good to know! This is gone now with the refactoring to use a converter.

test/e2e/replica_set_custom_role/replica_set_custom_role_test.go
alyacb: I made the change! Converter has replaced the builder (and it made testing changes a little easier).

test/e2e/mongodbtests/mongodbtests.go
alyacb: The functionality is similar; however, there are a few key differences. The main one is that I am running a different command than `getAdminSetting` in order to filter out builtin roles:
```golang
RunCommand(ctx, bson.D{
  {"rolesInfo", 1},            // nolint
  {"showPrivileges", true},    // nolint
  {"showBuiltinRoles", false}, // nolint
}).Decode(&result)
```
instead of
```golang
RunCommand(context.TODO(), bson.D{{Key: "getParameter", Value: 1}, {Key: key, Value: 1}})
```
Another difference is that because it is not possible to predict the order in which roles will be returned from mongo, and because it may not match the expected order, I am now using `assert.ElementsMatch(t, result.Roles, expectedRoles)` to compare the array result, whereas `hasAdminSetting` uses `reflect.DeepEqual(expectedValue, actualValue)`. Note that I added additional roles to the e2e test to ensure that all valid formats of resources are accepted.

I also prefer using the strong typing of decoding the result into a struct instead of a map using:
```golang
// CustomRolesResult is a type to decode the result of getting rolesInfo.
type CustomRolesResult struct {
	Roles []automationconfig.CustomRole
}
```

That being said, I will spend some time to see if I can refactor the tester nicely to reduce code duplication while keeping at least the first two requirements for this function intact.

alyacb:
I pushed some changes to address PR comments and expand the e2e test (and fix the case where `db: ""` or `collection: ""` in a resource).

test/e2e/mongodbtests/mongodbtests.go
alyacb: This is done! I refactored to reuse the old code. I think it looks much cleaner now, so thank you for pointing this out!

deploy/crds/mongodb.com_mongodb_crd.yaml
chatton: What is the error that we get from it? 

If it's an easy fix I think we can just do it here, if not maybe we can temporarily remove it and create a ticket to fix. Maybe @irajdeep knows a bit more about this?

test/e2e/replica_set_custom_role/replica_set_custom_role_test.go
```diff
@@ -0,0 +1,94 @@
...
+	t.Run("Create MongoDB Resource", mongodbtests.CreateMongoDBResource(&mdb, ctx))
+	t.Run("Basic tests", mongodbtests.BasicFunctionality(&mdb))
+	t.Run("Keyfile authentication is configured", tester.HasKeyfileAuth(3))
+	t.Run("Test Basic Connectivity", tester.ConnectivitySucceeds())
+	t.Run("Ensure Authentication", tester.EnsureAuthenticationIsConfigured(3))
+	t.Run("AutomationConfig has the correct version", mongodbtests.AutomationConfigVersionHasTheExpectedVersion(&mdb, 1))
+
+	// Verify automation config roles and roles created in admin database.
+	roles := mdbv1.ConvertCustomRolesToAutomationConfigCustomRole(mdb.Spec.Roles)
+	t.Run("AutomationConfig has the correct custom role", mongodbtests.AutomationConfigHasTheExpectedCustomRoles(&mdb, roles))
```
chatton, line 91: [q] is there a benefit of passing the roles here? We already have access to the mdb, we can get them from the spec there right?

test/e2e/util/mongotester/mongotester.go
```diff
@@ -135,7 +136,33 @@ func (m *Tester) HasTlsMode(tlsMode string, tries int, opts ...OptionApplier) fu
...
 }
 
-func (m *Tester) hasAdminParameter(key string, expectedValue interface{}, tries int, opts ...OptionApplier) func(t *testing.T) {
+// CustomRolesResult is a type to decode the result of getting rolesInfo.
+type CustomRolesResult struct {
+	Roles []automationconfig.CustomRole
+}
+
+func (m *Tester) VerifyRoles(expectedRoles []automationconfig.CustomRole, tries int, opts ...OptionApplier) func(t *testing.T) {
+	return m.hasAdminCommandResult(func(m *Tester, t *testing.T) bool {
```
chatton, line 145: [q] what is the reason for the `m` in the function signature `func(m *Tester, t *testing.T)` ?

It looks like the functions that use this are already on the `m *Tester` type and already have access to `m`.

chatton (commented):
This is looking great! Nice work so far, I just had a few more comments!

test/e2e/replica_set_custom_role/replica_set_custom_role_test.go
alyacb: I don't think its a major benefit, but this way we only do the conversion once- otherwise we would have to call the conversion function both in the automation config check and in the check against roles in the admin database.

test/e2e/util/mongotester/mongotester.go
alyacb: You're right, its superfluous. I have removed it. Good catch!

pkg/controller/mongodb/replica_set_controller.go
```diff
@@ -551,6 +551,16 @@ func (r ReplicaSetReconciler) validateUpdate(mdb mdbv1.MongoDB) error {
 	return validation.Validate(prevSpec, mdb.Spec)
 }
 
+func getCustomRolesModification(mdb mdbv1.MongoDB) (automationconfig.Modification, error) {
+	if mdb.Spec.Roles == nil {
```
chatton, line 555: what do you think about moving `mdb.Spec.Roles` to `mdb.Spec.Security.Authentication.Roles`? I think it makes more sense to group these roles with the other auth settings.

deploy/crds/mongodb.com_mongodb_crd.yaml
chatton: let's leave it for now and we can address this in a future PR.

deploy/crds/mongodb.com_mongodb_crd.yaml
alyacb: Fixed itself when I moved roles field to Spec.Security.Authentication.

deploy/crds/mongodb.com_mongodb_crd.yaml
alyacb: Done!

pkg/controller/mongodb/replica_set_controller.go
alyacb: Done!

chatton (approved):
LGTM, 👍  Great work on this one!

PR #300, opened 2021-01-27
Ensured volume mounts are sorted consistently

chatton:
### All Submissions:

This PR ensures when sorting volume mounts all of the fields are taken into consideration.


pkg/util/merge/merge_test.go
```diff
@@ -588,20 +585,6 @@ func TestMergeVolumeAddVolume(t *testing.T) {
 	assert.Equal(t, corev1.EmptyDirVolumeSource{}, *volume1.EmptyDir)
 }
 
-func TestMergeVolumeMounts(t *testing.T) {
```
chatton, line 591: this is actually the incorrect merging behaviour - uniqueness should be determined by name, path and subpath

pkg/util/merge/merge_test.go
```diff
@@ -239,7 +239,7 @@ func TestMergeContainer(t *testing.T) {
 
 		t.Run("Volume Mounts are overridden", func(t *testing.T) {
 			volumeMounts := mergedContainer.VolumeMounts
-			assert.Len(t, volumeMounts, 2)
+			assert.Len(t, volumeMounts, 3, "volume mounts can have the same name, the uniqness is the combination of name, path and subpath")
```
bznein, line 242:
```suggestion
			assert.Len(t, volumeMounts, 3, "volume mounts can have the same name, the uniqueness is the combination of name, path and subpath")
```

bznein (approved):

pkg/util/merge/merge.go
```diff
@@ -430,16 +431,21 @@ func VolumeMounts(original, override []corev1.VolumeMount) []corev1.VolumeMount
...
-		return mergedMounts[i].Name < mergedMounts[j].Name
+		return volumeMountToString(mergedMounts[i]) < volumeMountToString(mergedMounts[j])
 	})
 
 	return mergedMounts
 }
 
+// volumeMountToString returns a string consisting of all components of the given VolumeMount.
+func volumeMountToString(v corev1.VolumeMount) string {
+	return strings.Join([]string{v.Name, v.MountPath, v.SubPath}, "-")
```
alyacb, line 442: [q] Is it guaranteed that none of `Name`, `MountPath`, and `SubPath` will contain the "-" character? The reason I ask is so we don't get a case where we have two different (valid) volume mounts with the same `volumeMountToString()` value, e.g.:
```
Name: a-b
MountPath: c
SubPath: d
```
and 
```
Name: a
MountPath: b-c
SubPath: d
```
Will both have volumeMountToString: `a-b-c-d`

pkg/util/merge/merge_test.go
alyacb: [nit] Should it be "uniqueness", not "uniqness"?

alyacb (commented):
Just a question and a nit, otherwise looks good!

pkg/util/merge/merge_test.go
alyacb: Resolving! You fixed it before I could submit the comment 🎉

pkg/util/merge/merge.go
chatton: good point! Maybe we can change it to include a character that will not be able to appear in the names/paths

pkg/util/merge/merge_test.go
chatton: yes! Addressed in this commit https://github.com/mongodb/mongodb-kubernetes-operator/pull/300/commits/b2896a20a2281c621b1927209bfc73ec36d2ea6a

alyacb (approved):
LGTM! 🎉

PR #309, opened 2021-02-01
Added IsTransient function to check if an error is transient or not

chatton:
### All Submissions:

This PR lowers the severity of the transient `the object has been modified; please apply your changes to the latest version and try again` error

### DIFF
--- a/pkg/controller/mongodb/mongodb_status_options.go
+++ b/pkg/controller/mongodb/mongodb_status_options.go
@@ -1,7 +1,10 @@
 package mongodb
 
 import (
+	"errors"
+
 	mdbv1 "github.com/mongodb/mongodb-kubernetes-operator/pkg/apis/mongodb/v1"
+	"github.com/mongodb/mongodb-kubernetes-operator/pkg/util/apierrors"
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/util/result"
 	"go.uber.org/zap"
 
@@ -112,6 +115,10 @@ func (o *optionBuilder) withStatefulSetReplicas(members int) *optionBuilder {
 }
 
 func (o *optionBuilder) withMessage(severityLevel severity, msg string) *optionBuilder {
+	if apierrors.IsTransient(errors.New(msg)) {
+		severityLevel = Debug
+		msg = ""
+	}
 	o.options = append(o.options, messageOption{
 		message: message{
 			messageString: msg,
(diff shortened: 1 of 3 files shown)


pkg/controller/mongodb/mongodb_status_options.go
```diff
@@ -112,6 +115,10 @@ func (o *optionBuilder) withStatefulSetReplicas(members int) *optionBuilder {
 }
 
 func (o *optionBuilder) withMessage(severityLevel severity, msg string) *optionBuilder {
+	if apierrors.IsTransient(errors.New(msg)) {
```
bznein, line 118: I don't know if I'm totally convinced of this flow:

`withMessage` receives a string, which then is used to create an `error`, from which then a `string` is extracted.

But I understand if this is intended in order to have `IsTransient` adaptable to other scenarios in which it might receive an actual `error`

pkg/util/apierrors/apierrors.go
```diff
@@ -0,0 +1,12 @@
+package apierrors
+
+import "strings"
+
+// objectModifiedText is an error indicating that we are trying to update a resource that has since been updated.
+// in this we just want to retry but not log it as an error.
```
bznein, line 6:
```suggestion
// in this case we just want to retry but not log it as an error.
```

pkg/util/apierrors/apierrors_test.go
```diff
@@ -0,0 +1,37 @@
...
+			fmt.Errorf("Error updating the status of the MongoDB resource: Operation cannot be fulfilled on mongodbcommunity.mongodb.com \"mdb0\": The object has been modified; please apply your changes to the latest version and try again"),
+			true,
+		},
+		{
+			"Test Transient lower case error",
+			fmt.Errorf("error updating the status of the MongoDB resource: Operation cannot be fulfilled on mongodbcommunity.mongodb.com \"mdb0\": the object has been modified; please apply your changes to the latest version and try again"),
+			true,
+		},
+		{
+			"Test Not Transient Errir",
```
bznein, line 25:
```suggestion
			"Test Not Transient Error",
```

bznein (approved):
LGTM with a consideration

pkg/controller/mongodb/mongodb_status_options.go
chatton: @bznein I agree it's not a very nice usage. We could do something like

`IsTransientError` and `IsTransientMessage` or similar.

What do you think?

bznein: Yes I think it would be better :) Thanks!

chatton:
evergreen retry

PR #315, opened 2021-02-03
CLOUDP-81150: Replace operator-sdk test with envtest

alyacb:
### All Submissions:

* [X] Have you opened an Issue before filing this PR?

This PR removes `operator-sdk test` from our e2e tests. It switches to using `go test` to run the test and `envtest` to configure the environment. Most changes to individual e2e tests are just to refactor them to rely on globals/ a new main entry function refactored to test/e2e/client.go.

The downside is that test cleanup no longer happens after tests, which will not affect evergreen runs but will make post-test cleanup a manual process when developing locally.


deploy/e2e/role.yaml
```diff
@@ -0,0 +1,98 @@
...
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - rbac.authorization.k8s.io
```
alyacb, line 85: Needed a new role to add these additional permissions so the test could create the operator role. Furthermore, this had to be a cluster role for the clusterwide test.

go.mod
```diff
@@ -4,21 +4,31 @@ go 1.14
...
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
+	github.com/go-logr/zapr v0.1.1 // indirect
+	github.com/golang/groupcache v0.0.0-20191027212112-611e8accdfc9 // indirect
+	github.com/golang/protobuf v1.4.2 // indirect
 	github.com/hashicorp/go-multierror v1.0.0
 	github.com/hashicorp/golang-lru v0.5.4 // indirect
 	github.com/imdario/mergo v0.3.9
 	github.com/klauspost/compress v1.9.8 // indirect
-	github.com/konsorten/go-windows-terminal-sequences v1.0.3 // indirect
-	github.com/operator-framework/operator-sdk v0.19.4
```
alyacb, line 12: It looks like the operator-sdk test code was the only dependency on operator-sdk.

pkg/controller/mongodb/replicaset_controller_test.go
```diff
@@ -656,7 +657,7 @@ func setStatefulSetReadyReplicas(t *testing.T, c k8sClient.Client, mdb mdbv1.Mon
 
 // loadTestFixture will create a MongoDB resource from a given fixture
 func loadTestFixture(yamlFileName string) (mdbv1.MongoDBCommunity, error) {
-	testPath := fmt.Sprintf("testdata/%s", yamlFileName)
+	testPath := path.Join(e2eutil.TestdataDir, yamlFileName)
```
alyacb, line 660: The testdata dir was referenced in many places, and it changed since `go test` uses the directory of the test file as the current, not the root directory of the project. I refactored it along with the deploy directory into constants in e2eutil.

scripts/dev/e2e.py
```diff
@@ -163,17 +161,10 @@ def create_test_pod(args: argparse.Namespace, dev_config: DevConfig) -> None:
                         },
                     ],
                     "command": [
-                        "/bin/operator-sdk",
+                        "go",
```
alyacb, line 164: Key change here: `go test` instead of `operator-sdk test`

scripts/dev/e2e.py
```diff
@@ -45,14 +45,12 @@ def _prepare_test_environment(config_file: str) -> None:
 
     print("Creating Role")
     k8s_conditions.ignore_if_already_exists(
-        lambda: rbacv1.create_namespaced_role(dev_config.namespace, _load_test_role())
+        lambda: rbacv1.create_cluster_role(_load_test_role())
```
alyacb, line 48: Needed to use a cluster role to be able to create a cluster role for the clusterwide test.

test/e2e/client.go
```diff
@@ -0,0 +1,74 @@
+package e2eutil
+
+import (
```
alyacb, line 3: This file essentially replaces `f.MainEntry`, `f.Context`, and `f.Global`.

test/e2e/replica_set_cross_namespace_deploy/replica_set_cross_namespace_deploy_test.go
```diff
@@ -28,18 +27,18 @@ func TestCrossNamespaceDeploy(t *testing.T) {
 
 	postfix, err := generate.RandomValidDNS1123Label(5)
 	if err != nil {
-		t.Fatal()
+		t.Fatal(err)
```
alyacb, line 30: Flyby fix: I neglected to report errors in this file when I first committed it.

bznein:
Can you also make sure that the parts about e2e tests here: https://github.com/mongodb/mongodb-kubernetes-operator/blob/master/docs/contributing.md are up-to-date? Maybe they already are but I remember writing them some time ago with references to `operator-sdk test` so maybe they need some tweaking

alyacb:
> Can you also make sure that the parts about e2e tests here: https://github.com/mongodb/mongodb-kubernetes-operator/blob/master/docs/contributing.md are up-to-date? Maybe they already are but I remember writing them some time ago with references to `operator-sdk test` so maybe they need some tweaking

Done in previous commit! I removed mention of `operator-sdk` and added `envtest`.

test/e2e/client.go
```diff
@@ -0,0 +1,74 @@
...
+
+var TestClient client.Client
+var OperatorNamespace string
+var WatchNamespace string
+
+type CleanupOptions struct {
+	TestContext *Context
+}
+
+func (*CleanupOptions) ApplyToCreate(*client.CreateOptions) {}
```
chatton, line 24: [q] what is this for?

test/e2e/client.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	fmt.Println("Starting test")
+	code := m.Run()
+
+	err = testEnv.Stop()
+	if err != nil {
```
chatton, line 63: [nit] same comment as above

test/e2e/client.go
```diff
@@ -0,0 +1,74 @@
...
+		fmt.Println(err)
+	}
+
+	os.Exit(code)
```
chatton, line 67: with the os.Exit at this level, this isn't immediately obvious when looking at any individual test

e.g.

```golang
e2eutil.TestMainEntry(m)
```

Maybe we could make it a little more explict in the tests. Maybe if this function was instead called something like `RunTest(m *testing.M) int`

We get something like

```golang
os.Exit(e2eutil.RunTest(m))
```

What do you think?

test/e2e/client.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func NewContext(t *testing.T) *Context {
+	return &Context{}
+}
+
+func (ctx *Context) Cleanup() {}
```
chatton, line 74: [q] The plan is to implement this in a future PR right?

test/e2e/e2eutil.go
```diff
@@ -24,24 +23,29 @@ import (
 	k8sClient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	DeployDir   = "/go/deploy"
```
chatton, line 27: it looks like this is only being used in the `setup` package, can we unexport it and move it there?

test/e2e/setup/setup.go
```diff
@@ -126,37 +108,41 @@ func GeneratePasswordForUser(mdbu mdbv1.MongoDBUser, ctx *f.Context, namespace s
...
 
-func deployOperator(c f.FrameworkClient) error {
+func deployOperator(c client.Client) error {
 	testConfig := loadTestConfigFromEnv()
 
-	if err := buildKubernetesResourceFromYamlFile(c, path.Join(deployDir, "role.yaml"), &rbacv1.Role{}, withNamespace(testConfig.namespace)); err != nil {
+	e2eutil.OperatorNamespace = testConfig.namespace
+	fmt.Printf("Setting operator namespace to %s\n", e2eutil.OperatorNamespace)
+	if testConfig.clusterWide {
+		e2eutil.WatchNamespace = "*"
```
chatton, line 120: [q] do we need an `e2eutil.WatchNamespace` package level field? It looks like it could be replaced with a function level scoped variable (I only see it being used in `deployOperator`)

chatton (commented):
This is looking great, massive work so far! Just had a few questions.

deploy/e2e/operator.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: apps/v1
```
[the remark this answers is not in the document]
alyacb: Makes sense! The e2e role.yaml has additional permissions so it can create the test environment. I needed a new operator.yaml and serviceaccount.yaml because I changed the name of the role from mongodb-kubernetes-operator, however I'll look into whether we can just have role and role binding in this directory while reusing deploy/operator.yaml.

deploy/e2e/operator.yaml
```diff
@@ -0,0 +1,40 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      name: mongodb-kubernetes-operator
+  template:
+    metadata:
+      labels:
+        name: mongodb-kubernetes-operator
+    spec:
+      serviceAccountName: e2e-test
```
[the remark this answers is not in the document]
alyacb: I believe so

test/e2e/client.go
alyacb: Ah this is a leftover from when I was passing a `CleanupOptions` object to `client.Create(...)` calls. I'll remove it.

test/e2e/client.go
alyacb: Yes. I can also remove this for now and reintroduce it if/once I have a solution to clean up the test. As it turns out if I want to keep this, I also need to keep the empty `ApplyToCreate` function so I can pass CleanupOptions to Create calls. What do you think would be best?

deploy/e2e/operator.yaml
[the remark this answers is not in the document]
alyacb: Update: I removed operator.yaml from the new e2e directory, since we weren't using it anyway (the test deploys using deploy/operator). We do, however, need a new cluster role, cluster rolebinding, and serviceaccount to create the operator role and role binding in the test itself for the deploy. We also need these to be able to create a new namespace in the cross-namespace test.

test/e2e/client.go
alyacb: Sorry, I'm not sure which comment this refers to. If it's the exit code one, I also refactored this function to return an error code for the test to handle instead.

test/e2e/client.go
alyacb: You're right! I refactored this function to return an error and an exit code, to be handled in the caller. Each test now calls os.exit explicitly.

test/e2e/e2eutil.go
alyacb: Done!

test/e2e/setup/setup.go
alyacb: Done!

test/e2e/client.go
alyacb: Resolving as I implemented this function.

test/e2e/client.go
alyacb: Resolving since this is necessary to implement the cleanup logic, even though we don't use it. CleanupOptions needs this method so that it matches the interface type passed to client.Create()

alyacb:
I implemented cleanup logic so that the --preform-cleanup flag works with the new changes. My last commit also added comments/ reorganised setup.go slightly.

test/e2e/client.go
chatton: ah I think this comment was regarding preferring 

```golang
if err := fn(); err != nil {
   return err
}
```

over 

```golang
err := fn()
if err != nil {
  return err
}
```

Definitely not a blocker.

chatton (approved):
LGTM! Thanks for all the great work with this refactor 🥇

PR #321, opened 2021-02-04
Fixed merging of host aliases

bznein:
We weren't sorting hostaliases when merging, so the order was not predictable as ranging over a map has no guarantees on the order. This caused a flaky unit test.

I've ran this test locally around 10 times with no failures.

### All Submissions:

### DIFF
--- a/pkg/util/merge/merge_podtemplate_spec.go
+++ b/pkg/util/merge/merge_podtemplate_spec.go
@@ -1,6 +1,8 @@
 package merge
 
 import (
+	"sort"
+
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/util/contains"
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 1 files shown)


chatton (approved):
Ah thank you for fixing this! LGTM

alyacb (approved):
💯 This is great! LGTM! 

Just to double check that it works: since go test caches test results, did you clear the cache every time you ran the test or run it with a different flag, e.g.
```bash
for i in $(seq 1 100); go test -v -count=$i ./pkg/...; done
```
(with a different $i every time)?

bznein:
> 💯 This is great! LGTM!
> 
> Just to double check that it works: since go test caches test results, did you clear the cache every time you ran the test or run it with a different flag, e.g.
> 
> ```shell
> for i in $(seq 1 100); go test -v -count=$i ./pkg/...; done
> ```
> 
> (with a different $i every time)?

I cleaned the cache every time! But that would work as well!

This is what I did:

`go clean -testcache && go test ./pkg/...` (in a for loop of course)

PR #323, opened 2021-02-04
move Mypy to GitHub action

bznein:
This moves also mypy to github actions, with a caveat :(

### All Submissions:


.github/workflows/code_health.yml
```diff
@@ -11,3 +11,13 @@ jobs:
...
+
+  Mypy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Mypy linting
+        uses: jpetrucciani/mypy-check@master
+        with:
+          path: scripts/*/*.py
```
bznein, line 23: This isn't 100% the same as before.
Before we were using some ugly bash magic to get all python files, we can't do that with this (or other github actions) and we need to either specify the folders containing python files or some magic globbing like this.

A glob that would work would be `**/*.py`  but that syntax breaks github actions :(

Three possibilities here:
1) Trust that all python files will always be in `scripts/*/` or that we will remember to change this 
2) Close this PR and leave mypy checks in evg
3) Create a docker image that works for us

I'm open to both 1 and 2, I won't have time to do 3 for sure :)

rodrigovalin (approved):
LGTM

I agree with option 1). There should not be Python files scattered around the codebase. If they are all kept in `scripts` then I think that's fine.

chatton (approved):
I agree with @rodrigovalin , everything being under scripts makes a lot of sense - this is great that we can move away from our own bash scripts and use github actions, nice job!

PR #326, opened 2021-02-10
CLOUDP-82250: add required fields to automation config

chatton:
### All Submissions:

This PR adds all the fields that we use in the Enterprise operator to the AutomationConfig type. They are not yet all used in the Community operator.


pkg/util/versions/versions.go
```diff
@@ -0,0 +1,31 @@
...
+func CalculateFeatureCompatibilityVersion(versionStr string) string {
+	v1, err := semver.Make(versionStr)
+	if err != nil {
+		return ""
+	}
+
+	baseVersion, _ := semver.Make("3.4.0")
+	if v1.GTE(baseVersion) {
+		ans, _ := MajorMinorVersion(versionStr)
+		return ans
```
bznein, line 19: I think we can make it much easier. All `MajorMinorVersion` is doing is re-making the `semver` version we already have here as `v1`.

We can simply replace all of this with: `return fmt.Sprintf("%d.%d", v1.Major, v1.Minor)
`

pkg/util/versions/versions_test.go
```diff
@@ -0,0 +1,36 @@
+package versions
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCalculateFCV(t *testing.T) {
```
bznein, line 9: Can we add a test for the path that doesn't get a valid string? for example 3.0.0 -> ""

And one for an invalid semver like `3.4.5.2`?

pkg/automationconfig/automation_config_secret.go
```diff
@@ -11,6 +11,22 @@ import (
...
+func ReadFromSecret(secretGetter secret.Getter, secretNsName types.NamespacedName) (AutomationConfig, error) {
+	acSecret, err := secretGetter.GetSecret(secretNsName)
+	if err != nil && !apiErrors.IsNotFound(err) {
+		return AutomationConfig{}, err
+	}
+
+	// the secret not existing is perfectly fine, this will happen during the first reconciliation.
+	if apiErrors.IsNotFound(err) {
+		return AutomationConfig{}, nil
+	}
```
bznein, line 25: [q] what do you think, for readability, to write something like:


```go
if err != nil 
{
		return AutomationConfig{}, nil
	} else {
		return AutomationConfig{}, err
	}
```

We could also add a `ignoreNotFound` to `apiErrors` method similarly to: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/interfaces.go#L131

bznein, line 25: [q] what do you think of writing this as:

```go
	if err != nil {
		// the secret not existing is perfectly fine, this will happen during the first reconciliation.
			return AutomationConfig{}, nil
		} else  {
			return AutomationConfig{}, err
		}
}
```

Or maybe using [IgnoreNotFound](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/interfaces.go#L131) (although this would require changing `secret.Getter` to a client, or cast it - or write that method in our own error package)

pkg/automationconfig/automation_config_test.go
```diff
@@ -233,3 +235,50 @@ func TestMongoDBVersionsConfig(t *testing.T) {
...
+			createAutomationConfig("name0", "differentVersion", "domain0", Options{DownloadBase: "downloadBase1"}, Auth{Disabled: false}, 2, 2),
+			createAutomationConfig("name0", "version0", "domain0", Options{DownloadBase: "downloadBase0"}, Auth{Disabled: true}, 5, 2),
+		)
+
+		assert.NoError(t, err)
+		assert.False(t, areEqual)
+	})
+}
+
+func createAutomationConfig(name, version, domain string, opts Options, auth Auth, members, acVersion int) AutomationConfig {
```
bznein, line 274: [micro-nit] to make it faster to understand, can you maybe change `version` here to `mdbVersion` and also in the tests `version0` to `mdbVersion0` or something that makes it clear that is `mdb` (such as `4.4.0`).

pkg/util/versions/versions_test.go
```diff
@@ -0,0 +1,36 @@
...
+	assert.Equal(t, "4.2", CalculateFeatureCompatibilityVersion("4.2.0"))
```
bznein, line 10: can we increase coverage by adding a test for a version < 3.4.0 and one that is not a valid semver version?

pkg/util/versions/versions.go
bznein, line 19: [q] is this `MajorMinorVersion` function used anywhere else (since it's exported)?

if not, I think we can get rid of it. In this case all it does is recreating the exact same semver type that here is `v1`, we can just replace these two lines with:

`	 return fmt.Sprintf("%d.%d", v1.Major, v1.Minor)
`

bznein (commented):
Looking good! Just a couple of question and a nit!

pkg/automationconfig/automation_config_secret.go
chatton: ah I think we can use `IgnoreNotFound` to make it quite concise! I generally prefer to avoid using the `else` keyword if possible to make the main logic as flat as possible. We could do something like 

```golang

func ReadFromSecret(secretGetter secret.Getter, secretNsName types.NamespacedName) (AutomationConfig, error) {
	acSecret, err := secretGetter.GetSecret(secretNsName)
	if err != nil {
		return AutomationConfig{}, client.IgnoreNotFound(err)
	}
	return FromBytes(acSecret.Data[ConfigKey])
}
```

pkg/util/versions/versions.go
chatton: ah you're right, this was taken from enterprise, but for now we don't have a different use for it 👍

pkg/util/versions/versions_test.go
chatton: will add some more tests to cover this!

bznein (approved):
LGTM great stuff!

PR #339, opened 2021-02-18
CLOUDP-82915: Add Readinessprobe Dockerfile

chatton:
### All Submissions:

This PR adds a simple Dockerfile that builds the readiness probe. The enterprise codebase will use this image.

Note: This is still not being used in the community code base, refactoring the code to use the new readiness probe will come in a future PR.


scripts/dev/dockerfiles/Dockerfile.readiness
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.13-alpine as builder
```
irajdeep, line 1: should we bump the go version here as well?

scripts/dev/dockerfiles/Dockerfile.readiness
```diff
@@ -0,0 +1,13 @@
...
+
+COPY ./cmd/readiness /build/
+COPY ./pkg /build/pkg
+COPY ./api /build/api
+COPY ./go.mod /build/
```
irajdeep, line 6: So we have different go.mod file for the probe, can/should we use the project's go.mod?

scripts/dev/dockerfiles/Dockerfile.readiness
chatton: yes I think there's no harm doing that

scripts/dev/dockerfiles/Dockerfile.readiness
chatton: this is a single go.mod file, we just have a different main (in this case `cmd/readiness/main.go`)

irajdeep (approved):
LGTM 👍

PR #347, opened 2021-02-19
CLOUDP-79275: Add functions to check for agent readiness via annotations

chatton:
### All Submissions:

* [X] Have you signed our [CLA](https://www.mongodb.com/legal/contributor-agreement)?
* [X] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
* [X] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).

This PR introduces functions to check for agent readiness by reading annotations from the associated pods.

It is not used yet in the community code base. It will be introduced in a future PR along with the new readiness probe.

For context, this code is adapted from: [this method](https://github.com/10gen/ops-manager-kubernetes/blob/89b2faf71a313e267afe38521a81af1cfc7ae9f5/controllers/operator/appdbreplicaset_controller.go#L600)

PR #349, opened 2021-02-19
CLOUDP-78662: Use new readiness probe

chatton:
### All Submissions:

~~This should be reviewed after: #347~~

This PR updates the operator to use the new readiness probe (reading agent status from the pods themselves)

There were a few additional steps that needed to be taken to get everything working.

* The probe is now provided via an init container and the StatefulSet spec is updated accordingly.
* Now that we have deterministic checking of the agent status, we no longer need to spread partial TLS upgrade logic in different parts of the reconcile loop, we can now do this all in one place, with the only thing that needs to change being the order of the AutomationConfig update vs the StatefulSet update. These two updates have been moved to one place, with the order of operations depending on the state of the resource.
* Removed code around scale down detection, as the new readiness probe handles this for us.


controllers/replica_set_controller.go
```diff
@@ -182,22 +178,6 @@ func (r ReplicaSetReconciler) Reconcile(ctx context.Context, request reconcile.R
 		)
 	}
 
-	// if we're scaling down, we need to wait until the StatefulSet is at the
-	// desired number of replicas. Scaling down has to happen one member at a time
-	if scale.IsScalingDown(mdb) {
```
chatton, line 187: the check for scaling down no longer needs to happen as the readiness check later on accepts the expected number of replicas that should be ready.

controllers/replica_set_controller.go
```diff
@@ -331,60 +289,103 @@ func (r ReplicaSetReconciler) Reconcile(ctx context.Context, request reconcile.R
...
+
+	currentSts, err := r.client.GetStatefulSet(mdb.NamespacedName())
+	if err != nil {
+		return false, fmt.Errorf("error getting StatefulSet: %s", err)
+	}
+
+	r.log.Debugf("Ensuring StatefulSet is ready, with type: %s", getUpdateStrategyType(mdb))
+
+	isReady := statefulset.IsReady(currentSts, mdb.StatefulSetReplicasThisReconciliation())
+	return isReady || currentSts.Spec.UpdateStrategy.Type == appsv1.OnDeleteStatefulSetStrategyType, nil
```
chatton, line 308: we check for the OnDelete strategy type here as we want to be considered ready in the case of a version change even if the sts itself is not ready.

controllers/replicaset_scaledown.go
```diff
@@ -21,38 +16,3 @@ func updateScalingStatus(statusWriter k8sClient.StatusWriter, mdb mdbv1.MongoDBC
 	)
 	return err
 }
-
-// checkIfStatefulSetMembersHaveBeenRemovedFromTheAutomationConfig ensures that the expected number of StatefulSet
-// replicas are ready. When a member has its process removed from the Automation Config, the pod will eventually
-// become unready. We use this information to determine if we are safe to continue the reconciliation process.
-func checkIfStatefulSetMembersHaveBeenRemovedFromTheAutomationConfig(stsGetter statefulset.Getter, statusWriter k8sClient.StatusWriter, mdb mdbv1.MongoDBCommunity) (reconcile.Result, error) {
```
chatton, line 28: existed as part of the scale down check - can be removed completely

scripts/dev/e2e.py
```diff
@@ -164,7 +164,8 @@ def create_test_pod(args: argparse.Namespace, dev_config: DevConfig) -> None:
                         "go",
                         "test",
                         "-v",
-                        "-timeout=60m",
+                        "-timeout=30m",
+                        "-failfast",
```
chatton, line 168: this was added to prevent the tests from continuing when they had already failed.

scripts/dev/e2e.py
```diff
@@ -164,7 +164,8 @@ def create_test_pod(args: argparse.Namespace, dev_config: DevConfig) -> None:
                         "go",
                         "test",
                         "-v",
-                        "-timeout=60m",
+                        "-timeout=30m",
```
chatton, line 167: the new readiness probe prevents the need to wait for so long.

controllers/replica_set_controller.go
```diff
@@ -77,11 +82,6 @@ const (
 	// configured
 	lastVersionAnnotationKey    = "mongodb.com/v1.lastVersion"
 	lastSuccessfulConfiguration = "mongodb.com/v1.lastSuccessfulConfiguration"
-	// tlsRolledOutAnnotationKey indicates if TLS has been fully rolled out
-	tlsRolledOutAnnotationKey      = "mongodb.com/v1.tlsRolledOut"
-	hasLeftReadyStateAnnotationKey = "mongodb.com/v1.hasLeftReadyStateAnnotationKey"
-
-	trueAnnotation = "true"
```
chatton, line 84: tls annotation checking is no longer required, this is now accounted for when determining the order of the sts/automation config updates.

controllers/replica_set_controller.go
```diff
@@ -519,37 +502,37 @@ func getCustomRolesModification(mdb mdbv1.MongoDBCommunity) (automationconfig.Mo
 	}, nil
 }
 
-func (r ReplicaSetReconciler) buildAutomationConfigSecret(mdb mdbv1.MongoDBCommunity) (corev1.Secret, error) {
+func (r ReplicaSetReconciler) buildAutomationConfig(mdb mdbv1.MongoDBCommunity) (automationconfig.AutomationConfig, error) {
```
chatton, line 505: extracted this function into two separate ones, one that builds the Automation Config and one that builds the secret.

pkg/util/functions/functions.go
```diff
@@ -0,0 +1,38 @@
...
+		return runInOrder(funcs...)
+	}
+	return runReversed(funcs...)
+}
+
+func runInOrder(funcs ...func() (bool, error)) (bool, error) {
+	for _, fn := range funcs {
+		ready, err := fn()
+		if err != nil {
+			return ready, err
```
bznein, line 18: [micro-nit] Can we find a better name than `ready`? This package is quite generic and `ready` seems to be more suited for a specific case.

pkg/util/functions/functions.go
```diff
@@ -0,0 +1,38 @@
...
+
+// RunSequentially executes a series of functions sequentially. Each function returns a boolean
+// indicating if the function was successfully, and an error indicating if something went wrong.
+// if any function returns an error, an early exit happens. The first parameter indicates if the functions
+// should run in the order provided. A value of true indicates they should run in reverse.
+func RunSequentially(reversed bool, funcs ...func() (bool, error)) (bool, error) {
+	if !reversed {
+		return runInOrder(funcs...)
+	}
+	return runReversed(funcs...)
```
bznein, line 11: [micro-nit] I find it more logic to swap the two, but it's 100% up to you :D

pkg/util/functions/functions.go
```diff
@@ -0,0 +1,38 @@
+package functions
+
+// RunSequentially executes a series of functions sequentially. Each function returns a boolean
+// indicating if the function was successfully, and an error indicating if something went wrong.
+// if any function returns an error, an early exit happens. The first parameter indicates if the functions
+// should run in the order provided. A value of true indicates they should run in reverse.
+func RunSequentially(reversed bool, funcs ...func() (bool, error)) (bool, error) {
```
bznein, line 7: [q] why are we using a boolean that tells us if we want to flip the order of execution?

I'm thinking of how a dev would interpret seeing this function called:

`RunSequentially(true, functions...)`

would make me thing that these will be ran in order, not in the opposite one.

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -160,7 +160,7 @@ func CreateMongoDBResource(mdb *mdbv1.MongoDBCommunity, ctx *e2eutil.Context) fu
 
 func BasicFunctionality(mdb *mdbv1.MongoDBCommunity) func(*testing.T) {
 	return func(t *testing.T) {
-		t.Run("Config Map Was Correctly Created", AutomationConfigSecretExists(mdb))
+		t.Run("Secret Was Correctly Created", AutomationConfigSecretExists(mdb))
```
bznein, line 163: 👍🏻

bznein (commented):
Looking really good! 
I think functionally this is ready to be merged, I just have a couple of comments about clarity of the code in one particular file

pkg/util/functions/functions.go
chatton: ah, yes that probably makes more sense! I'll do that refactor 👍

chatton:
@bznein ready for another look when you have time!

bznein (approved):
LGTM!

PR #360, opened 2021-02-24
CLOUDP-83427: Add function to ensure creation of the Automation Config Secret

chatton:
### All Submissions:

* [N/A] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).

This PR introduces a new function which can be used in community + enterprise projects that ensures the creation of the Automation Config secret. The most recent Automation Config is returned, this is important to ensure we have the correct target version for the agent readiness checks.


pkg/automationconfig/automation_config_secret.go
```diff
@@ -0,0 +1,66 @@
...
+	if existingAcBytes, ok := existingSecret.Data[ConfigKey]; !ok {
+		// the secret exists but the key is not present. We can update the secret
+		existingSecret.Data[ConfigKey] = acBytes
+	} else {
+		// the secret already exists, we should check to see if we're making any changes.
+		existingAutomationConfig, err := FromBytes(existingAcBytes)
+		if err != nil {
+			return AutomationConfig{}, err
+		}
+		// we are attempting to update with the same version, no change is required.
```
bznein, line 40: [q] what's `version` here? Not the one you set [here](https://github.com/mongodb/mongodb-kubernetes-operator/pull/360/files#diff-d990f1832c6e156b4f6f747a6c3fc767f546cb53fb9a76684550a7d51b67cd71R247)?

Do you just mean "with the same content"?

bznein (approved):
LGTM

pkg/automationconfig/automation_config_secret.go
chatton: yes the same content, the agent will re-attempt to configure the state of the cluster if the version is different (event if contents are the same) so we just need to check for equal content. (while excluding the version field)

PR #366, opened 2021-03-02
CLOUDP-83838: Fix logging configuration and add volume for logs

chatton:
### All Submissions:

We previously had not had a log volume set, so logs were being lost with pod restarts. The logging settings were also being misconfigured.

they were set at process[.].systemLog rather than the correct process[n].args2_6.systemLog, the previous settings were being ignored by the agent as the field was not being recognized.

### DIFF
--- a/controllers/build_statefulset_test.go
+++ b/controllers/build_statefulset_test.go
@@ -58,13 +58,13 @@ func assertStatefulSetIsBuiltCorrectly(t *testing.T, mdb mdbv1.MongoDBCommunity,
 	assert.True(t, reflect.DeepEqual(probes.New(defaultReadiness()), *probe))
 	assert.Equal(t, probes.New(defaultReadiness()).FailureThreshold, probe.FailureThreshold)
 	assert.Equal(t, int32(5), probe.InitialDelaySeconds)
-	assert.Len(t, agentContainer.VolumeMounts, 5)
+	assert.Len(t, agentContainer.VolumeMounts, 6)
 
 	assertContainsVolumeMountWithName(t, agentContainer.VolumeMounts, "agent-scripts")
 	assertContainsVolumeMountWithName(t, agentContainer.VolumeMounts, "automation-config")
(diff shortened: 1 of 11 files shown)


controllers/build_statefulset_test.go
```diff
@@ -58,13 +60,36 @@ func assertStatefulSetIsBuiltCorrectly(t *testing.T, mdb mdbv1.MongoDBCommunity,
...
+	assertContainsVolumeMountWithName(t, mongodContainer.VolumeMounts, "logs-volume")
+	assertContainsVolumeMountWithName(t, mongodContainer.VolumeMounts, "my-rs-agent-scram-credentials")
 
 	initContainer := sts.Spec.Template.Spec.InitContainers[0]
 	assert.Equal(t, versionUpgradeHookName, initContainer.Name)
 	assert.Equal(t, "version-upgrade-hook-image", initContainer.Image)
 	assert.Len(t, initContainer.VolumeMounts, 1)
 }
+
+func assertContainsVolumeMountWithName(t *testing.T, mounts []corev1.VolumeMount, name string) {
```
chatton, line 86: added this function as we were previously just checking for the number of volumemounts, not that they were the correct ones.

pkg/automationconfig/automation_config.go
```diff
@@ -34,37 +34,65 @@ type Process struct {
...
+	return p.SetArgs26Field("replication.replSetName", replSetName)
+}
+
+func (p *Process) SetSystemLog(systemLog SystemLog) *Process {
+	return p.SetArgs26Field("systemLog.path", systemLog.Path).
+		SetArgs26Field("systemLog.destination", systemLog.Destination)
+}
+
+// SetArgs26Field should be used whenever any args26 field needs to be set. It ensures
+// that the args26 map is non nil and assingns the given value.
```
bznein, line 83:
```suggestion
// that the args26 map is non nil and assigns the given value.
```

bznein (approved):
LGTM, thanks for catching this!

PR #368, opened 2021-03-03
Separate Keyfile & Password into separate secrets

chatton:
### All Submissions:

We were including the agent password + keyfile in a single secret, but these are not required to be kept together, this PR separates them into separate secrets and renames the volume to a more meaningful name.

### DIFF
--- a/pkg/kube/secret/secret.go
+++ b/pkg/kube/secret/secret.go
@@ -109,3 +109,25 @@ func HasAllKeys(secret corev1.Secret, keys ...string) bool {
 	}
 	return true
 }
+
+// EnsureSecretWithKey makes sure the Secret with the given name has a key with the given value if the key is not already present.
+// if the key is present, it will return the existing value associated with this key.
+func EnsureSecretWithKey(secretGetUpdateCreateDeleter GetUpdateCreateDeleter, nsName types.NamespacedName, key, value string) (string, error) {
+	existingSecret, err0 := secretGetUpdateCreateDeleter.GetSecret(nsName)
+	if err0 != nil {
+		if apiErrors.IsNotFound(err0) {
+			s := Builder().
+				SetNamespace(nsName.Namespace).
+				SetName(nsName.Name).
+				SetField(key, value).
+				Build()
+
+			if err1 := secretGetUpdateCreateDeleter.CreateSecret(s); err1 != nil {
+				return "", err1
+			}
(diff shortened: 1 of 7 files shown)


pkg/kube/secret/secret.go
```diff
@@ -109,3 +109,25 @@ func HasAllKeys(secret corev1.Secret, keys ...string) bool {
 	}
 	return true
 }
+
+// EnsureSecretWithKey makes sure the Secret with the given name has a key with the given value if the key is not already present.
+// if the key is present, it will return the existing value associated with this key.
+func EnsureSecretWithKey(secretGetUpdateCreateDeleter GetUpdateCreateDeleter, nsName types.NamespacedName, key, value string) (string, error) {
+	existingSecret, err0 := secretGetUpdateCreateDeleter.GetSecret(nsName)
```
bznein, line 116: [nit] I think we generally tend to reuse the name `err` every time? is there specific reason for keeping them separated here?

pkg/kube/secret/secret.go
```diff
@@ -109,3 +109,25 @@ func HasAllKeys(secret corev1.Secret, keys ...string) bool {
...
+			s := Builder().
+				SetNamespace(nsName.Namespace).
+				SetName(nsName.Name).
+				SetField(key, value).
+				Build()
+
+			if err1 := secretGetUpdateCreateDeleter.CreateSecret(s); err1 != nil {
+				return "", err1
+			}
+			return value, nil
```
bznein, line 128: this could probably be just `return value, secretGetUpdateCreateDeleter.CreateSecret(s)` in theory. Unless you want to make sure to return an empty string in case of error

bznein (approved):
LGTM

pkg/kube/secret/secret.go
chatton: there was a shadowing issue where the err1 check was referring to err0 when they were both `err`

pkg/kube/secret/secret.go
chatton: ah you're right, I generally like returning the empty string in cases like this, for example if an error check is forgotten, there is a valid value used and the calling code will get further than it should and hit errors further down the line that are likely harder to debug!

bznein: Makes sense!

PR #371, opened 2021-03-04
CLOUDP-82719: Make statefulset construction exported

bznein:
Separate common parts of the statefulset that will be imported by enterprise, and configures the rest separately

The importing of this new method has already been tested on the enterprise side 

### All Submissions:

* [n/a] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).


api/v1/mongodbcommunity_types.go
```diff
@@ -485,6 +492,10 @@ func (m MongoDBCommunity) CurrentReplicas() int {
 	return m.Status.CurrentStatefulSetReplicas
 }
 
+func (m MongoDBCommunity) GetVersion() string {
```
chatton, line 495: [nit] maybe this could be a but more specific (maybe `GetMongoDBVersion`) since we deal with a lot of different components that have different versions. (agent/mongodb/init containers and OpsManager with enterprise)

controllers/replica_set_controller.go
```diff
@@ -713,7 +713,32 @@ exec mongod -f /data/automation-mongod.conf ;
 	)
 }
 
+type statefulSetSpec interface {
```
chatton, line 716: I have a few comments/concerns about this

1. I don't think `statefulSetSpec` is a great name for what this is, with this name I would assume it is a type related to `StatefulSet.Spec`. What this actually is is an interface which defines what a resource needs to actually build a StatefulSet.

2. I don't think we should have an unexported interface that is an argument for an exported function.

3. Does it belong in this package? It's currently in the `controllers` package (which I don't think anything should be exported from, this package should just be in charge of the community operator specific controller) I think any functionality that should be shared makes more sense in the `statefulset` package, maybe `statefulset.Owner`/`statefulset.Provider` etc - what do you think?

The reason I can think of against putting it in the statefulset package is that up until now that package has contained functions that acted solely on the statefulset primitives, but I think realistically the package exists to be used between our two projects, and if there was a reason to remove it from there later we could address it then.

4. I think an interface like this needs a docstring.

controllers/replica_set_controller.go
```diff
@@ -713,7 +713,32 @@ exec mongod -f /data/automation-mongod.conf ;
...
+			podtemplatespec.Apply(
+				buildTLSPodSpecModification(mdb),
+			),
+		),
+
+		statefulset.WithCustomSpecs(mdb.Spec.StatefulSetConfiguration.SpecWrapper.Spec),
+	)
+}
+
+func BuildCommonStatefulSetModificationFunction(mdb statefulSetSpec, scaler scale.ReplicaSetScaler) statefulset.Modification {
```
chatton, line 741: common between what? I think this needs a docstring. Also, I don't think we should use the word common!

I think any of the following are more appropriate

`StandardReplicaSetStatefulset`, `UnconfiguredReplicaSetStatefulSet`, `ReplicaSetStatefulSet`, `MongoDBReplicaSetStatefulSet`

with a docstring or note explicitly stating which configuration is absent. What do you think?

chatton: Also, I think in this case it makes sense to just return an `appv1.StatefulSet`, I don't really think we get much benefit from returning the function itself, the caller can make any changes required. It should result in a simpler api.

chatton (commented):
Looking good, had a bunch of comments largely around naming and packages.

api/v1/mongodbcommunity_types.go
```diff
@@ -461,6 +461,13 @@ func (m MongoDBCommunity) NamespacedName() types.NamespacedName {
 	return types.NamespacedName{Name: m.Name, Namespace: m.Namespace}
 }
 
+func (m MongoDBCommunity) GetName() string {
+	return m.Name
+}
+func (m MongoDBCommunity) GetNamespace() string {
+	return m.Namespace
+}
```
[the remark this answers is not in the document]
bznein: Removing this makes building fails because

```	"github.com/mongodb/mongodb-kubernetes-operator/api/v1".MongoDBCommunity does not implement statefulSetSpec (GetName method has pointer receiver)```

And I think it's clearer to declare them to compatibility with the AppDB in Enterprise, which doesn't have ObjectMeta. What do you think?

File: api/v1/mongodbcommunity_types.go (hunk 1 above)
Comment (bznein) [reply api/v1/mongodbcommunity_types.go:495]: Absolutely right!

File: controllers/replica_set_controller.go (hunk 2 above)
Comment (bznein) [reply controllers/replica_set_controller.go:716]: 1. Naming is hard :) But I agree with you! I will think about a better name
2. Yeah looks like it's not the best choice by googling a bit, will change it
3. Will try to move things around, I initially tried to have it in another package but was hitting issues with a few circular dependencies, will try again and see what I can get to!
4. Yeah, 100%, thanks for noticing!

File: api/v1/mongodbcommunity_types.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (chatton) [reply api/v1/mongodbcommunity_types.go:469]: ah sure makes sense, I'm okay with it 👍

File: controllers/replica_set_controller.go (hunk 3 above)
Comment (bznein) [reply controllers/replica_set_controller.go:741]: Changed the name, but as agreed on zoom, not changing the return type for now!

File: controllers/construct/build_statefulset_test.go [hunk 5]
```diff
@@ -13,18 +13,34 @@ import (
...
 func TestMultipleCalls_DoNotCauseSideEffects(t *testing.T) {
-	_ = os.Setenv(mongodbRepoUrl, "repo")
-	_ = os.Setenv(mongodbImageEnv, "mongo")
+	_ = os.Setenv(MongodbRepoUrl, "repo")
+	_ = os.Setenv(MongodbImageEnv, "mongo")
+	_ = os.Setenv(AgentImageEnv, "agent-image")
 
 	mdb := newTestReplicaSet()
-	stsFunc := buildStatefulSetModificationFunction(mdb)
+	stsFunc := BuildMongoDBReplicaSetStatefulSetModificationFunction(&mdb, mdb)
```
chatton, line 43: [q] hmm this usage is a little weird, what if the interface was also composed of the `scale.ReplicaSetScaler` interface (since no matter what it needs to be)

controllers/construct/mongodbstatefulset.go
```diff
@@ -0,0 +1,246 @@
...
+	MongodbRepoUrl  = "MONGODB_REPO_URL"
+
+	versionUpgradeHookImageEnv = "VERSION_UPGRADE_HOOK_IMAGE"
+	headlessAgentEnv           = "HEADLESS_AGENT"
+	podNamespaceEnv            = "POD_NAMESPACE"
+	automationConfigEnv        = "AUTOMATION_CONFIG_MAP"
+)
+
+// MongoDBStatefulSetOwner is an interface which any resource which generates a MongoDB StatefulSet should implement.
+type MongoDBStatefulSetOwner interface {
```
chatton, line 48: 👍

controllers/construct/mongodbstatefulset.go
```diff
@@ -0,0 +1,246 @@
...
+	ServiceName() string
+	GetName() string
+	GetNamespace() string
+	GetMongoDBVersion() string
+	AutomationConfigSecretName() string
+	GetAgentKeyfileSecretNamespacedName() types.NamespacedName
+}
+
+// BuildMongoDBReplicaSetStatefulSetModificationFunction build the parts of the replica set that are common between every resource that implements
```
chatton, line 57:
```suggestion
// BuildMongoDBReplicaSetStatefulSetModificationFunction builds the parts of the replica set that are common between every resource that implements
```

chatton (approved):
LGTM! Just one final suggestion but looks great after that.

controllers/construct/mongodbstatefulset.go
chatton:
```golang
type MongoDBStatefulSetOwner interface {
 scale.ReplicaSetScaler
...
}

```

controllers/construct/build_statefulset_test.go
bznein: I know it looks ugly! I tried that, but on enteprise AppDB does not implement `scale.ReplicaSetScaler` as the required methods are on the opsmanager type (we need to access the `status.AppDB.Replicas` field)

bznein: (so effectively it will look like `BuildMongoDBReplicaSetStatefulSetModificationFunction(om.Spec.AppDB, om)`

chatton: ah right I remember now, the other option I guess is to have a `GetScaler()` function on the interface, I'm okay to stick with this though, we can adjust it later if it makes sense!

chatton: Or we could provide the AppDB with the OM status/spec fields in enterprise.

bznein: Can we have a `GetScaler()`? AppDB type doesn't have a right reference to OpsManager type as far as I remember. 
We can discuss this later anyway!

chatton: sure! I was more thinking now that we could add the fields to the AppDB type and let the AppDB implement the interface (the first suggestion not the GetScaler)

bznein: Yes that will make sense, will leave for a future PR in case :)

PR #378, opened 2021-03-08
Create ubi buildvariant

chatton:
### All Submissions:

This PR introduces a second build variant which runs the tests using a ubi version of the Agent. All other images are currently the same. As we add new distros to these images we can update accordingly.


scripts/dev/e2e.py
```diff
@@ -255,6 +255,12 @@ def parse_args() -> argparse.Namespace:
         help="Watch all namespaces",
         action="store_true",
     )
+    parser.add_argument(
+        "--distro",
+        help="The distro of images that should be used",
+        type=str,
```
bznein, line 261: [nit] Can we use `choices` here to make sure that we don't pass something completely different? https://docs.python.org/3/library/argparse.html#choices

bznein (approved):

PR #387, opened 2021-03-18
Refactor get update strategy and annotations

bznein:
This PR refactors the logic around the stateful set update strategy and annotations. Now all this logic is independent of the specific type and it's exportable, so we can reuse it in enterprise 

### All Submissions:

### DIFF
--- a/api/v1/mongodbcommunity_types.go
+++ b/api/v1/mongodbcommunity_types.go
@@ -490,6 +490,9 @@ func (m MongoDBCommunity) GetMongoDBVersion() string {
 	return m.Spec.Version
 }
 
+// GetMongoDBVersionForAnnotation returns the MDB version used to annotate the object.
+// Here it's the same of GetMongoDBVersion, but a different name is used in order to make
+// the usage clearer in enterprise (where it's a method of OpsManager but is used for the AppDB)
 func (m MongoDBCommunity) GetMongoDBVersionForAnnotation() string {
 	return m.GetMongoDBVersion()
 }
@@ -502,6 +505,8 @@ func (m *MongoDBCommunity) StatefulSetReplicasThisReconciliation() int {
 	return scale.ReplicasThisReconciliation(m)
 }
 
+// GetUpdateStrategyType returns the type of RollingUpgradeStrategy that the
+// MongoDB StatefulSet should be configured with.
 func (m MongoDBCommunity) GetUpdateStrategyType() appsv1.StatefulSetUpdateStrategyType {
 	if !m.IsChangingVersion() {
 		return appsv1.RollingUpdateStatefulSetStrategyType
@@ -509,11 +514,13 @@ func (m MongoDBCommunity) GetUpdateStrategyType() appsv1.StatefulSetUpdateStrate
 	return appsv1.OnDeleteStatefulSetStrategyType
 }
 
+// IsChangingVersion returns true if an attempted version change is occurring.
 func (m MongoDBCommunity) IsChangingVersion() bool {
 	prevVersion := m.getPreviousVersion()
 	return prevVersion != "" && prevVersion != m.Spec.Version
 }
 
+// GetPreviousVersion returns the last MDB version the statefulset was configured with.
 func (m MongoDBCommunity) getPreviousVersion() string {
 	return annotations.GetLastAppliedMongoDBVersion(&m)
 }
--- a/api/v1/mongodbcommunity_types.go
+++ b/api/v1/mongodbcommunity_types.go
@@ -491,7 +491,7 @@ func (m MongoDBCommunity) GetMongoDBVersion() string {
 }
 
 // GetMongoDBVersionForAnnotation returns the MDB version used to annotate the object.
-// Here it's the same of GetMongoDBVersion, but a different name is used in order to make
+// Here it's the same as GetMongoDBVersion, but a different name is used in order to make
 // the usage clearer in enterprise (where it's a method of OpsManager but is used for the AppDB)
 func (m MongoDBCommunity) GetMongoDBVersionForAnnotation() string {
 	return m.GetMongoDBVersion()
@@ -522,7 +522,7 @@ func (m MongoDBCommunity) IsChangingVersion() bool {
 
 // GetPreviousVersion returns the last MDB version the statefulset was configured with.
 func (m MongoDBCommunity) getPreviousVersion() string {
-	return annotations.GetLastAppliedMongoDBVersion(&m)
+	return annotations.GetAnnotation(&m, annotations.LastAppliedMongoDBVersion)
 }
 
 func (m MongoDBCommunity) GetAnnotations() map[string]string {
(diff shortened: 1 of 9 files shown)


api/v1/mongodbcommunity_types.go
```diff
@@ -497,6 +505,30 @@ func (m *MongoDBCommunity) StatefulSetReplicasThisReconciliation() int {
...
+func (m MongoDBCommunity) IsChangingVersion() bool {
+	prevVersion := m.getPreviousVersion()
+	return prevVersion != "" && prevVersion != m.Spec.Version
+}
+
+// GetPreviousVersion returns the last MDB version the statefulset was configured with.
+func (m MongoDBCommunity) getPreviousVersion() string {
+	return annotations.GetLastAppliedMongoDBVersion(&m)
+}
+
```
bznein, line 527: Moved from replica_set_controller.go

controllers/replica_set_controller.go
```diff
@@ -50,9 +51,6 @@ import (
 const (
 	clusterDNSName = "CLUSTER_DNS_NAME"
 
-	// lastVersionAnnotationKey should indicate which version of MongoDB was last
-	// configured
-	lastVersionAnnotationKey    = "mongodb.com/v1.lastVersion"
```
bznein, line 55: This was actually not used, moved it and renamed (`lastVersion` would have been misleading in enterprise)

controllers/replica_set_controller.go
```diff
@@ -217,8 +215,13 @@ func (r ReplicaSetReconciler) Reconcile(ctx context.Context, request reconcile.R
...
 
-	if err := r.updateCurrentSpecAnnotation(mdb); err != nil {
-		r.log.Errorf("Could not save current state as an annotation: %s", err)
+	// the last version will be duplicated in two annotations.
+	// This is needed to reuse the update strategy logic in enterprise
+	if err := annotations.UpdateLastAppliedMongoDBVersion(&mdb, r.client); err != nil {
+		r.log.Errorf("Could not save current version as an annotation: %s", err)
+	}
+	if err := r.updateLastSuccessfulConfiguration(mdb); err != nil {
+		r.log.Errorf("Could not save current spec as an annotation: %s", err)
```
bznein, line 224: This is still needed as we need to save the TLS specs for the `validate` method

pkg/kube/statefulset/statefulset.go
```diff
@@ -297,3 +298,17 @@ func VolumeMountWithNameExists(mounts []corev1.VolumeMount, volumeName string) b
 	}
 	return false
 }
+
```
bznein, line 301: we didn't need this in the `controller` package, especially since now we can export it more cleanly

pkg/kube/client/mocked_client.go
```diff
@@ -96,7 +100,50 @@ func (m *mockedClient) Update(_ context.Context, obj k8sClient.Object, _ ...k8sC
...
 }
 
-func (m *mockedClient) Patch(_ context.Context, _ k8sClient.Object, _ k8sClient.Patch, _ ...k8sClient.PatchOption) error {
+type patchValue struct {
+	Op    string      `json:"op"`
+	Path  string      `json:"path"`
+	Value interface{} `json:"value"`
+}
+
+func (m *mockedClient) Patch(_ context.Context, obj k8sClient.Object, patch k8sClient.Patch, _ ...k8sClient.PatchOption) error {
```
bznein, line 109: For now we just implement mocked patch for what was needed now. 
Adding other options is outside the scope of this PR. I tried to cover every possible case and return a `not implemented` error when needed

rodrigovalin (approved):
LGTM

api/v1/mongodbcommunity_types.go
```diff
@@ -489,6 +490,13 @@ func (m MongoDBCommunity) GetMongoDBVersion() string {
 	return m.Spec.Version
 }
 
+// GetMongoDBVersionForAnnotation returns the MDB version used to annotate the object.
+// Here it's the same of GetMongoDBVersion, but a different name is used in order to make
```
irajdeep, line 494: Did you mean "same as GetMongoDBVersion"?

api/v1/mongodbcommunity_types.go
```diff
@@ -497,6 +505,30 @@ func (m *MongoDBCommunity) StatefulSetReplicasThisReconciliation() int {
...
+
+// IsChangingVersion returns true if an attempted version change is occurring.
+func (m MongoDBCommunity) IsChangingVersion() bool {
+	prevVersion := m.getPreviousVersion()
+	return prevVersion != "" && prevVersion != m.Spec.Version
+}
+
+// GetPreviousVersion returns the last MDB version the statefulset was configured with.
+func (m MongoDBCommunity) getPreviousVersion() string {
+	return annotations.GetLastAppliedMongoDBVersion(&m)
```
irajdeep, line 525: Can we just use the `GetAnnotation` method instead of having this indirection? (Of course, after making getAnnotation public).

pkg/kube/annotations/annotations.go
```diff
@@ -0,0 +1,84 @@
...
+	Op    string      `json:"op"`
+	Path  string      `json:"path"`
+	Value interface{} `json:"value"`
+}
+
+const (
+	LastAppliedMongoDBVersion = "mongodb.com/v1.lastAppliedMongoDBVersion"
+)
+
+func getAnnotation(object Versioned, key string) string {
```
irajdeep, line 29: Can we make this method public like "SetAnnotations". It will also enable us to get the keys without auxiliary methods like `GetLastAppliedMongoDBVersion`. WDYT?

api/v1/mongodbcommunity_types.go
```diff
@@ -497,6 +505,30 @@ func (m *MongoDBCommunity) StatefulSetReplicasThisReconciliation() int {
...
+	if !m.IsChangingVersion() {
+		return appsv1.RollingUpdateStatefulSetStrategyType
+	}
+	return appsv1.OnDeleteStatefulSetStrategyType
+}
+
+// IsChangingVersion returns true if an attempted version change is occurring.
+func (m MongoDBCommunity) IsChangingVersion() bool {
+	prevVersion := m.getPreviousVersion()
+	return prevVersion != "" && prevVersion != m.Spec.Version
```
irajdeep, line 520: Q: Can `prevVersion` be empty (I think it could be in the case the first time an upgrade is performed) ? What should be the behaviour in that case?

irajdeep (commented):
Overall LGTM 👍 ,  added some comments.

api/v1/mongodbcommunity_types.go
bznein: Whoops some Italian construct crept into my English :D thanks for noticing!

api/v1/mongodbcommunity_types.go
bznein: yeah you are right, that's clearer! will do

api/v1/mongodbcommunity_types.go
bznein: yes, if there are no annotations the expected behaviour is to return false as it will mean there is no version change in progress since n version has been deployed yet

irajdeep (approved):

PR #395, opened 2021-03-23
Use emptyDir instead of volumefromsecret for keyfile

bznein:
This PR changes the volume mount for the keyfile. Instead of mounting from a secret we mount an emptyDir, and then let the agent write the keyfile with correct permissions

This avoids the issue of having the volume mounted as read-only (which always happens when mounting volumes from secrets)

### All Submissions:


chatton (approved):
LGTM once all the tests have passed!

PR #401, opened 2021-03-26
Added MANAGED_SECURITY_CONTEXT check and refactored env vars

bznein:
We weren't correctly using the `MANAGED_SECURITY_CONTEXT` env var, which is needed only in the operator and not in MongoDB pods. 

This PR uses this env var when building the Statefulset, and it adds security context to the pods only when it is not set.

It also refactors usage of env vars, to avoid duplication

### All Submissions:

### DIFF
--- a/pkg/util/envvar/envvars.go
+++ b/pkg/util/envvar/envvars.go
@@ -8,6 +8,15 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	AgentImageEnv              = "AGENT_IMAGE"
+	MongodbImageEnv            = "MONGODB_IMAGE"
+	WatchNamespaceEnv          = "WATCH_NAMESPACE"
+	VersionUpgradeHookImageEnv = "VERSION_UPGRADE_HOOK_IMAGE"
+	ReadinessProbeImageEnv     = "READINESS_PROBE_IMAGE"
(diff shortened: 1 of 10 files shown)


pkg/util/envvar/envvars.go
```diff
@@ -8,6 +8,15 @@ import (
...
 )
 
+const (
+	AgentImageEnv              = "AGENT_IMAGE"
+	MongodbImageEnv            = "MONGODB_IMAGE"
+	WatchNamespaceEnv          = "WATCH_NAMESPACE"
+	VersionUpgradeHookImageEnv = "VERSION_UPGRADE_HOOK_IMAGE"
+	ReadinessProbeImageEnv     = "READINESS_PROBE_IMAGE"
+	ManagedSecurityContextEnv  = "MANAGED_SECURITY_CONTEXT"
+)
```
chatton, line 18: I'm not sure that this is the correct package for these values, the current `envvars` package has functions related to common functionality around env vars, these are community operator specific - what do you think?

chatton (approved):
LGTM - just had one question regarding the package for the constants.

pkg/util/envvar/envvars.go
bznein: Good point, I was actually conflicted on where to put them. Where do you think they will fit best?

PR #407, opened 2021-03-31
Changed agent version

bznein:
### All Submissions:

### DIFF
--- a/test/e2e/setup/test_config.go
+++ b/test/e2e/setup/test_config.go
@@ -26,7 +26,7 @@ func loadTestConfigFromEnv() testConfig {
 		namespace:               envvar.GetEnvOrDefault(testNamespaceEnvName, "default"),
 		operatorImage:           envvar.GetEnvOrDefault(operatorImageEnvName, "quay.io/mongodb/community-operator-dev:latest"),
 		versionUpgradeHookImage: envvar.GetEnvOrDefault(construct.VersionUpgradeHookImageEnv, "quay.io/mongodb/mongodb-kubernetes-operator-version-upgrade-post-start-hook:1.0.2"),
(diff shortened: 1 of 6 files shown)


test/e2e/setup/test_config.go
```diff
@@ -26,7 +26,7 @@ func loadTestConfigFromEnv() testConfig {
 		namespace:               envvar.GetEnvOrDefault(testNamespaceEnvName, "default"),
 		operatorImage:           envvar.GetEnvOrDefault(operatorImageEnvName, "quay.io/mongodb/community-operator-dev:latest"),
 		versionUpgradeHookImage: envvar.GetEnvOrDefault(construct.VersionUpgradeHookImageEnv, "quay.io/mongodb/mongodb-kubernetes-operator-version-upgrade-post-start-hook:1.0.2"),
-		agentImage:              envvar.GetEnvOrDefault(construct.AgentImageEnv, "quay.io/mongodb/mongodb-agent:10.27.0.6772-1"), // TODO: better way to decide default agent image.
+		agentImage:              envvar.GetEnvOrDefault(construct.AgentImageEnv, "quay.io/mongodb/mongodb-agent:10.29.0.6830-1"), // TODO: better way to decide default agent image.
```
chatton, line 29: [q] maybe we shouldn't have a default at all, I don't think there is an image that could always be safe to use as we add new features, I think it makes more sense to fail loudly in the case of no agent image present, what do you think?

chatton (approved):
LGTM once all the tests are green! I left one question but I think that's outside the scope of this PR.

test/e2e/setup/test_config.go
bznein: yeah I think you might be right!

PR #420, opened 2021-04-08
Cloudp 84351 add agent images to dockerfile generator

chatton:
### All Submissions:

This PR moves the static agent Dockerfiles into templates for the dockerfile_generator

* [n/a] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).


scripts/dev/dockerfile_generator.py
```diff
@@ -9,6 +11,43 @@
 GOLANG_TAG = "1.14"
 
 
+def _shared_agent_params() -> DockerParameters:
+    with open("release.json", "r") as f:
+        release = json.loads(f.read())
+
+    return {
+        "template_path": "scripts/dev/templates/agent",
```
bznein, line 19: [q] why is this is shared if we always replace that?

bznein (approved):
LGTM only one q!

scripts/dev/dockerfile_generator.py
chatton: both agents have the same template path, but currently all the other images are just in "scripts/dev/templates"

chatton: ah just seeing now you mean we replace it in both but with the same value - this is a mistake! Will correct it.

PR #422, opened 2021-04-09
CLOUDP-83092: Add sonar inventory

chatton:
### All Submissions:

This PR adds an `inventory.yaml` file which specifies how the agent images should be built. There are no periodic builds introduced yet.

The images still have the expiry tags: https://quay.io/repository/mongodb/mongodb-agent-ubi-dev?tab=tags


scripts/ci/build_and_push_image_sonar.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+
+# shellcheck disable=SC1091
+. venv/bin/activate
+echo "${quay_password:?}" | docker login "-u=${quay_user_name:?}" quay.io --password-stdin
```
rodrigovalin, line 6: Depending on execution parameters of this shell script, this password could be printed to stdout by mistake.

Can you add an explicit `set +x` to this script, right before `docker login` and add a comment that it should not be removed, because it could cause the quay_password contents to be printed.

Optimally we would use a credentials helper, but none seems to exist for Quay.

inventory.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+      - name: agent-ubuntu-build
+        task_type: docker_build
+        dockerfile: scripts/dev/templates/agent/Dockerfile.ubuntu-$(inputs.params.version_id)
+
+        labels:
+          quay.expires-after: 48h
+
+        output:
+          - registry: $(inputs.params.registry)/mongodb-agent-dev
```
rodrigovalin, line 32: What about naming these registries in a more consistent way?

* `mongodb-agent-dev-ubuntu` and `mongodb-agent-dev-ubi` 

Instead of

* `mongodb-agent-dev` and `mongodb-agent-ubi-dev`

rodrigovalin (commented):
One concern: Let's be explicit at setting no echoing of executed commands in bash (set +x) and we should be good to go.

inventory.yaml
chatton: yes that makes more sense! Will rename to keep things consistent.

chatton:
@rodrigovalin I made changes based on your feedback, please take another look when you get a chance.

rodrigovalin (approved):
LGTM

PR #429, opened 2021-04-19
CLOUDP-87241: Fix readiness probe check for MongoDB up

bznein:
### All Submissions:

* [n/a] Have you signed our [CLA](https://www.mongodb.com/legal/contributor-agreement)?
* [] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).


pkg/readiness/health/health.go
```diff
@@ -30,7 +30,7 @@ type processHealth struct {
 	IsInGoalState   bool               `json:"IsInGoalState"`
 	LastMongoUpTime int64              `json:"LastMongoUpTime"`
 	ExpectedToBeUp  bool               `json:"ExpectedToBeUp"`
-	ReplicaStatus   *replicationStatus `json:"ReplicationStatus,omitempty"`
```
rodrigovalin, line 33: I'm concerned about this change, (please correct me if I'm wrong).

In older versions of the agent this attribute won't be available, what will happen in that case. It is possible to use this readinessprobe version with an older version of the agent (not community, but Enterprise), if deploying an old version of Ops Manager, for instance. In that case `ReplicationStatus` won't be there.

cmd/readiness/readiness_test.go
```diff
@@ -131,18 +131,6 @@ func TestPodReadiness(t *testing.T) {
 	})
 }
 
-func TestServerHasAlreadyStarted(t *testing.T) {
-	t.Run("Agent should report server has started", func(t *testing.T) {
-		healthDoc := readHealthinessFile("testdata/health-status-readable-state.json")
```
rodrigovalin, line 136: Are those files still in use? If not, then they shoud be removed.

cmd/readiness/main.go
```diff
@@ -240,61 +240,11 @@ func isInReadyState(health health.Status) bool {
 			return true
 		}
 
-		if mongoDbServerHasStarted(health) {
-			// There should be only one entry reported for this Pod.
-			return processHealth.IsReadyState()
-		}
+		// Every time the process health is created by the agent,
```
rodrigovalin, line 243: This is true only for versions of the agent that support this field; old agents won't write it.

cmd/readiness/main.go
```diff
@@ -240,61 +240,11 @@ func isInReadyState(health health.Status) bool {
...
+		// it checks if the MongoDB process is up and populates this field
+		// (https://github.com/10gen/mms-automation/blob/bb72f74a22d98cfa635c1317e623386b089dc69f/go_planner/src/com.tengen/cm/healthcheck/status.go#L43)
+		// So it's enough to check that this value is not the zero-value for int64
+		return processHealth.LastMongoUpTime != 0 && processHealth.IsReadyState()
```
rodrigovalin, line 247: Ok, I get it, we don't have to check that the process is started if the process is already on PRIMARY or SECONDARY, great improvement.

However we take into account where the agent won't write this ReplicationStatus, in which case, we'll have to check if one of the steps is `StartFresh` with a result of `success`.

pkg/readiness/health/health.go
bznein: I believe that in that case `ReplicaStatus` will just be `nil`?

I also tried to do the following change:

```
diff --git a/cmd/readiness/testdata/health-status-ok-no-replica-status.json b/cmd/readiness/testdata/health-status-ok-no-replica-status.json
index caaf67f..0398e8e 100644
--- a/cmd/readiness/testdata/health-status-ok-no-replica-status.json
+++ b/cmd/readiness/testdata/health-status-ok-no-replica-status.json
@@ -73,7 +73,6 @@
     },
     "statuses": {
         "bar": {
-            "ReplicationStatus": null,
             "ExpectedToBeUp": true,
             "LastMongoUpTime": 1568222195,
             "IsInGoalState": true
```

And to run the following unit test:

```
	t.Run("MongoDB replication state is reported by agents", func(t *testing.T) {
		assert.True(t, isPodReady(testConfig("testdata/health-status-ok-no-replica-status.json")))
	})
```

And it works fine. Does this answer the question? I can add an explicit new unit test to cover this

rodrigovalin (commented):
We need to support agents that won't write ReplicationStatus (older than). In that case we can just consider ready state to be when the MongoDB server has started.

The learned to write ReplicationStatus in version 10.14.17.6444-1, OM version 4.4.6 (CLOUDP-77580)

cmd/readiness/readiness_test.go
bznein: You are right, I will

cmd/readiness/main.go
bznein: The part when the replication status is not present was not touched by this PR. 
This was previously handled already in the code, and the function explicitly has a comment already:

```
// isInReadyState checks the MongoDB Server state. It returns true if the state
// is PRIMARY or SECONDARY.
// This function will always return true if the agent doesn't publish this state.
```

and it does
```
	if processHealth.ReplicaStatus == nil {
			// We always return true if the Agent does not publish mongodb
			// server state
			return true
		}
```

However, we could use the same logic we use for `lastMongoUpTime`

cmd/readiness/main.go
bznein: The field `lastMongoUp` is not related to the replication status, and it has been there since two years ago:

https://github.com/10gen/mms-automation/blame/bb72f74a22d98cfa635c1317e623386b089dc69f/go_planner/src/com.tengen/cm/healthcheck/status.go#L45

rodrigovalin: Sorry, I meant `processHealth.IsReadyState()` which checks for the ReplicationStatus

pkg/readiness/health/health.go
rodrigovalin: It does not hurt to add that test, let's cover this with a unit test.

cmd/readiness/main.go
bznein: That's true, but we don't get to that call if replication status is `nil`, because that is checked right above

cmd/readiness/main.go
rodrigovalin: I think you are right, if we check for `lastMongoUpTime` that should be enough.

bznein:
@rodrigovalin thanks for the feedback, I think I addressed all of the comments here, this is ready for another review when you have the time

rodrigovalin (approved):
LGTM

chatton (commented):
All the changes look good to me, would be great if we can confirm this will fix our e2e replica set upgrade test though before merging.

bznein:
I've ran the broken e2e upgrade test 10 times, and it failed 4 times.

The test is still flaky, as expected, but now it is not failing 100% of the times, merging.

PR #437, opened 2021-04-21
CLOUDP-83092: Add agent release tasks

chatton:
This PR adds agent release tasks to evergreen.yml


.evergreen.yml
```diff
@@ -368,6 +368,23 @@ tasks:
...
+          release: true
+
+  - name: release_agent_ubi
+    commands:
+      - func: clone
+      - func: setup_virtualenv
+      - func: build_and_push_image_sonar
+        vars:
+          image_type: agent-ubi
+          release: true
```
bznein, line 387: [q] I guess there's no way to have a single task and somehow pass to it the `image_type` instead of duplicating this?

bznein (approved):
LGTM!

.evergreen.yml
chatton: I think this is most re-use we can get, the main thing is the re-use of `build_and_push_image_sonar`

PR #440, opened 2021-04-23
CLOUDP-83092: Create context images for init containers

chatton:
This PR refactors the init containers to be built from a context image similar to the agent and enterprise images.


bznein (approved):
LGTM!

PR #442, opened 2021-04-23
Fix incorrect arg parse field

chatton:
### All Submissions:

The release flag was being triggered with every task. A value of true was returned if the field was specified at all, it does not account for the value of the bool.

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -1,7 +1,7 @@
 import argparse
 import json
 import sys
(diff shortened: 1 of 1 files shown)


bznein (approved):
Looks really ugly but I can't really think of a nicer way to do it, I tried some combinations and they don't work... LGTM!

rodrigovalin (approved):
does the job!

PR #446, opened 2021-04-26
Move context images into same repo as the built image

chatton:
This PR updates context images to existing in the same repository as the images it is used to build.


inventory.yaml
```diff
@@ -24,9 +24,12 @@ images:
           agent_distro: linux_x86_64
           tools_distro: ubuntu1604-x86_64
 
+        labels:
+          quay.expires-after: 48h
+
         output:
-          - registry: $(inputs.params.registry)/agent-ubuntu-context-dev
-            tag: $(inputs.params.version_id)
+          - registry: $(inputs.params.registry)/mongodb-agent-dev
```
chatton, line 31: reverting to `mongodb-agent-dev` as this is the expected format for our daily builds. It expects `<image>` for ubuntu, and `<image>-ubi` for ubi. We can re-visit this in the future.

rodrigovalin (approved):
LGTM

PR #448, opened 2021-04-27
Make docker file S3 bucket configurable

chatton:
This PR makes the S3 bucket configurable. This allows the s3 dockerfile publishing to be run against dev s3 buckets. The evergreen config file points at the one we release to as part of the release task.


rodrigovalin (approved):
LGTM

bznein (approved):

PR #450, opened 2021-04-27
Build images with no expiry

chatton:
This PR changes our tagging to release to re-building with the same Dockerfile/args. This is not ideal. The reason is due to the expiry labels: https://jira.mongodb.org/browse/CLOUDP-88562

We can re-assess in the future, but for now this is a mechanism to ensure we don't include expiry labels during the release process.


irajdeep (approved):
ugh but LGTM : -)

PR #451, opened 2021-04-27
Add operator to pipeline

chatton:
This PR moves the building of the operator and e2e images into pipeline. It does not refactor them to use context images.


inventories/e2e-inventory.yaml
```diff
@@ -0,0 +1,33 @@
+vars:
```
bznein, line 1: 2 questions: 

Why is the agent inventory in `inventory.yaml` and these two images in the `inventories` dir?

If we keep these here, can we remove `-inventory` from the filename? They are already in the `inventories` dir, we already know what they are

bznein (approved):
LGTM I just have one question, not blocking!

rodrigovalin (approved):

PR #453, opened 2021-04-28
State Machine Design #2: Implement SaveLoader interface with ReplicaSetReconciler

chatton:
This PR adds the Save/Load methods required in order to determine the correct states that should be used with the State Machine.


controllers/mongodb_statemachine_save_loader.go
```diff
@@ -0,0 +1,144 @@
...
+
+// LoadNextState fetches the next state that should be executed. This is fetched
+// from the annotations of the resource of the given NamespacedName.
+func (r *ReplicaSetReconciler) LoadNextState(nsName types.NamespacedName) (string, error) {
+	mdb, err := r.client.GetMongoDBCommunity(nsName)
+	if err != nil {
+		return "", err
+	}
+
+	nextState, err := getNextState(mdb)
```
chatton, line 32: remove this function

PR #480, opened 2021-05-11
CLOUDP-89359: Ensure Mongod logs are sent to stdout upon restart

chatton:
### All Submissions:

This PR ensures that the logs of the Mongod are displayed upon restart, previously they were being written to a log file and not displayed when the container was restarted.

### DIFF
--- a/controllers/construct/build_statefulset_test.go
+++ b/controllers/construct/build_statefulset_test.go
@@ -5,6 +5,9 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/mongodb/mongodb-kubernetes-operator/pkg/kube/container"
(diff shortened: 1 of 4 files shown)


controllers/construct/build_statefulset_test.go
```diff
@@ -57,6 +60,24 @@ func TestMultipleCalls_DoNotCauseSideEffects(t *testing.T) {
 	})
 }
 
+func TestMongod_Container(t *testing.T) {
```
chatton, line 63: unrelated test that ended up being written as part of investigation, no harm in keeping it I don't think!

irajdeep (approved):
LGTM

PR #484, opened 2021-05-12
Fix pod annotations overwrite

### LINKED ISSUE #482: Operator overwrites Pod annotations
If deploying a Replica set with the below `MongoDBCommunity`, the Pod annotations defined on the `statefulSet.template.metadata` are overwritten by the annotation `agent.mongodb.com/version` after the `mongodb-agent` reachs its goal state.

```
kind: MongoDBCommunity
metadata:
  name: sandbox-mongodb
spec:
  members: 3
  type: ReplicaSet
  version: "4.2.13"
  featureCompatibilityVersion: "4.2"
  security:
    tls:
      optional: true
      enabled: true
      certificateKeySecretRef:
        name: sandbox-mongodb-cert
      caConfigMapRef:
        name: sandbox-mongodb-ca
    authentication:
      modes: ["SCRAM"]
  users:
    - name: admin
      db: admin
      passwordSecretRef:
        name: admin-password
      roles:
        - name: clusterAdmin
          db: admin
        - name: userAdminAnyDatabase
          db: admin
      scramCredentialsSecretName: admin
    - name: exporter
      db: admin
      passwordSecretRef:
        name: exporter-password
      roles:
        - name: clusterMonitor
          db: admin
        - name: read
          db: local
      scramCredentialsSecretName: exporter
  statefulSet:
    spec:
      template:
        metadata:
          annotations:
            prometheus.io/scrape: "true"
            prometheus.io/path: "/metrics"
            prometheus.io/port: "9216"
        spec:
          terminationGracePeriodSeconds: 90
          containers:
            - name: mongod
              imagePullPolicy: IfNotPresent
              resources:
                limits:
                  cpu: "2"
                  memory: 1250M
                requests:
                  cpu: "2"
                  memory: 1000M
            - name: mongodb-agent
              imagePullPolicy: IfNotPresent
              resources:
                limits:
                  cpu: "1"
                  memory: 650M
                requests:
                  cpu: "1"
                  memory: 500M
            - name: mongodb-monitor
              env:
                - name: EXPORTER_PASSWORD
                  valueFrom:
                    secretKeyRef:
                      name: exporter-password
                      key: password
                - name: MONGODB_URI
                  value: mongodb://exporter:$(EXPORTER_PASSWORD)@127.0.0.1:27017/
              image: interal-registry.io/monitor:1.0.0
              imagePullPolicy: IfNotPresent
              ports:
              - name: metrics
                containerPort: 9216
                protocol: TCP
              resources:
                limits:
                  cpu: "0.5"
                  memory: 800M
                requests:
                  cpu: "0.5"
                  memory: 500M
```

I belive the code responsible for this behaviour is this function: https://github.com/mongodb/mongodb-kubernetes-operator/blob/ad71d9494e388327f8575dd7f284cf13f08fa8d5/pkg/readiness/pod/podannotation.go#L14-L32

**What did you expect?**
The annotations should be preserved on the Pod definition.

**What happened instead?**
The Pod annotations are being overwritten.

**Operator Information**
 - Operator Version: v0.6.0
 - MongoDB Image used: mongo:4.2.13

caiorcferreira:
Closes #482 

After this is merged a new version of the `readiness-probe` image must be released.

### DIFF
--- a/cmd/readiness/testdata/k8sobjects.go
+++ b/cmd/readiness/testdata/k8sobjects.go
@@ -7,6 +7,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Currently seems like the appending functionality on the library used by the fake
+// implementation to simulate JSONPatch is broken: https://github.com/evanphx/json-patch/issues/138
+// The short term workaround is to have the annotation empty.
+
 // These are just k8s objects used for testing. Note, that these are defined in a non "_test.go" file as they are reused
 // by other modules
 func TestSecret(namespace, name string, version int) *corev1.Secret {
@@ -21,6 +25,9 @@ func TestPod(namespace, name string) *corev1.Pod {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
 			Namespace: namespace,
+			Annotations: map[string]string{
+				"agent.mongodb.com/version": "",
+			},
 		},
 	}
 }
--- a/pkg/readiness/pod/podannotation.go
+++ b/pkg/readiness/pod/podannotation.go
@@ -1,7 +1,9 @@
 package pod
 
 import (
+	"context"
 	"go.uber.org/zap"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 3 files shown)


caiorcferreira:
Currently seems like the appending functionality on the library used by the fake implementation of the `ClienteSet` to simulate JSONPatch is broken: https://github.com/evanphx/json-patch/issues/138
The short term workaround is to have the annotation empty on the test data.

chatton:
Hi @caiorcferreira thanks for the PR!

I just merged a patch that will ensure that the readiness probe that gets configured in our tests is the one that is built from source. Would you be able to rebase your fork and we can run the full set of tests with the updated probe?

Thanks!

caiorcferreira:
Rebased made! Thanks for the fast reply, I'm deeply needed for this fix.

chatton:
Hi @caiorcferreira 

I created a pre-release image of the readiness probe based on your changes. The tests haven't finished running yet but feel free to test it out!

you can set operator deployment env var `READINESS_PROBE_IMAGE` to use this new image as there are no operator changes.

`quay.io/mongodb/mongodb-kubernetes-readinessprobe-dev:1.0.4-prerelease`

chatton:
It looks like there is an issue, the annotation is not being applied correctly to the pod, we see the error `The Pod 'mdb0-0' doesn't have annotation 'agent.mongodb.com/version' yet`.

caiorcferreira:
> It looks like there is an issue, the annotation is not being applied correctly to the pod, we see the error `The Pod 'mdb0-0' doesn't have annotation 'agent.mongodb.com/version' yet`.

I'm running an image without the rebased code and is working fine.

I will test the prerelease image you provided to see if something changes.

caiorcferreira:
Using the prerelease image `quay.io/mongodb/mongodb-kubernetes-readinessprobe-dev:1.0.4-prerelease` worked fine to me.

Immediately after (+/- 3s) the first pod of the StatefulSet was created the annotation already appeared as `agent.mongodb.com/version: "-1"`.

After all the members of my Replica Set were up the annotation then transitioned to `agent.mongodb.com/version: "1"`, which in my setup took 4 minutes approximately.

> The Pod 'mdb0-0' doesn't have annotation 'agent.mongodb.com/version' yet.

Looking at the error message the test may have a timing dependency and the correct append operation may be taking a little longer. This would be my guess.

caiorcferreira:
Hi @chatton, I can see that some checks failed. Can you provide more information so I try to diagnose the problem? Or can you verify if your tests have some temporal dependency that may be breaking?

chatton:
Hi @caiorcferreira I deployed the operator locally with the updated readiness probe and things look like they're working just fine. I've re-run the tests and will look into this further if they're red again.

chatton:
@caiorcferreira based on the information [here](https://stackoverflow.com/questions/55573724/create-a-patch-to-add-a-kubernetes-annotation) it looks like the current solution will not work when the pod does not have any annotations at all. Looking at the diagnostics files in the test run, the pods created have no annotations at all - so I suspect the issue is that in our environments the pods being created have some annotations already present.

caiorcferreira:
@caiorcferreira I cannot reproduce this in my environment because it adds an annotation to all pods, but using `kubectl` I was able to see that patchs to add annotations when they don't exist indeed fail.

I tried to used Strategy Merge Patch and Merge Patch but both fail if no container is specified.

Hence, the only solution as I see is to get the pod first, check if it has annotations and **if not**, prepand the JSONPatch with an operation like this:
```json
{
  "op": "add",
  "path": "/metadata/annotations",
  "value": {}
}
```

caiorcferreira:
@chatton can you run the tests with this new commit?

chatton:
@caiorcferreira I've re-run the tests and they are looking green! Thank you for the contribution. This looks good to merge!

Can you confirm that you've signed the [CLA](https://www.mongodb.com/legal/contributor-agreement)?

caiorcferreira:
@chatton that's great! Yes, I have signed the CLA.

chatton (approved):
LGTM! Thanks for the contribution.

PR #492, opened 2021-05-20
CLOUDP-73359: resource cleanup

### LINKED ISSUE #203: mongodb-svc doesn't get removed
Hi,

After removing the mongodb resource everything gets removed except for the "*-svc".

Is that on purpose or a bug?

tibulca:
closes #203


controllers/replica_set_controller.go
```diff
@@ -501,7 +502,7 @@ func (r ReplicaSetReconciler) buildAutomationConfig(mdb mdbv1.MongoDBCommunity)
 	}
 
 	auth := automationconfig.Auth{}
-	if err := scram.Enable(&auth, r.client, mdb); err != nil {
+	if err := scram.Enable(&auth, r.client, mdb, []metav1.OwnerReference{getOwnerReference(mdb)}); err != nil {
```
bznein, line 505: I kinda feel we should extract getOwnerReference to somewhere else and we can call it directly whenever we need it instead of passing it by to the Enable function. 

I know it wont't save too much code but it looks cleaner to me

irajdeep: +1 to the suggestion.

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -333,3 +359,14 @@ func findContainerByName(name string, containers []corev1.Container) *corev1.Con
...
+
+func assertEqualOwnerReference(t *testing.T, resourceType string, resourceNamespacedName types.NamespacedName, ownerReferences []metav1.OwnerReference, expectedOwnerReference metav1.OwnerReference) {
+	assert.Len(t, ownerReferences, 1, fmt.Sprintf("%s %s/%s doesn't have OwnerReferences", resourceType, resourceNamespacedName.Name, resourceNamespacedName.Namespace))
+
+	assert.Equal(t, expectedOwnerReference.APIVersion, ownerReferences[0].APIVersion)
+	assert.Equal(t, "MongoDBCommunity", ownerReferences[0].Kind)
+	assert.Equal(t, expectedOwnerReference.Name, ownerReferences[0].Name)
+	assert.Equal(t, expectedOwnerReference.UID, ownerReferences[0].UID)
+
+	t.Logf("%s %s/%s has the correct OwnerReference!", resourceType, resourceNamespacedName.Name, resourceNamespacedName.Namespace)
```
irajdeep, line 371: nit: do we need this extra log `assert` functions will report if there are any discrepancies

irajdeep:
Nice work overall! Added some minor comments.

test/e2e/mongodbtests/mongodbtests.go
tibulca: I removed it. You're right, it doesn't add much value (it actually makes the log a bit too verbose)

controllers/replica_set_controller.go
tibulca: refactored this part after chatting with @bznein : getOwnerReference is not part of MongoDBCommunity and the Configurable interface

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -74,21 +74,41 @@ func statefulSetIsNotReady(mdb *mdbv1.MongoDBCommunity, interval time.Duration,
 
 func StatefulSetHasOwnerReference(mdb *mdbv1.MongoDBCommunity, expectedOwnerReference metav1.OwnerReference) func(t *testing.T) {
```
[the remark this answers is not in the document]
tibulca: IMO this will be a redundant since we'll actually not test our functionality but the one k8s has. If we check that these resources were created and they have the correct OwnerReference, then the expectation would be that k8s would delete them if their owner reference is deleted.
Or maybe there are additional checkes we can get from this proposed e2e test?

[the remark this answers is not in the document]
irajdeep: Generally, we write the E2E test just to automate the manual test we would be doing anyway to test the workflow.

But in this particular case, I think it's fine to skip that.

irajdeep (approved):
LGTM 👍

pkg/authentication/scram/scram.go
```diff
@@ -43,6 +44,9 @@ type Configurable interface {
 
 	// NamespacedName returns the NamespacedName for the resource that is being configured.
 	NamespacedName() types.NamespacedName
+
+	// GetOwnerReferences returns the OwnerReference pointing to the current resource.
```
bznein, line 48:
```suggestion
	// GetOwnerReferences returns the OwnerReferences pointing to the current resource.
```

bznein (approved):
LGTM great work!

pkg/authentication/scram/scram.go
bznein: Although we actually return exactly one, so maybe not even the right suggestion :D up to you!

tibulca: changed

rodrigovalin (approved):
LGTM, well done!

config/run/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+# Value of this field is prepended to the
+# names of all resources, e.g. a deployment named
+# "wordpress" becomes "alices-wordpress".
+# Note that it should also match with the prefix (text before '-') of the namespace
+# field above.
```
chatton, line 5: can we make this comment more relevant to the purpose of this file?

chatton (commented):
Looking great! One thing, would you be able to update the `RELEASE_NOTES.md` file with the impact these changes will have on our release notes? It looks like we have not yet made any changes for release `0.6.1`. You can wipe what is currently in `RELEASE_NOTES.md` and start the `0.6.1` changes. (this file will be used to create the github release). You can use the previous release notes as a template, there is also `release-notes-template.md` that you can reference.

priyolahiri (approved):
Good stuff! LGTM

config/run/kustomization.yaml
irajdeep: @chatton this new kustomization file is for when we want to run the `manager` in localhost instead of a K8s cluster right?

chatton: yes, this is for when we don't have a k8s deployment running anywhere, the operator runs as a local go binary. We just create the required roles/serviceaccount for everything to run correctly.

tibulca: done

chatton (approved):
LGTM!

PR #493, opened 2021-05-21
CLOUDP-89260: improve readiness probe

antonlisovenko:
This PR improves the readiness probe logic - it checks the state of mongodb and if it has been down for more than 1 minute - returns "not ready". The health status file `lastMongoUpTime` is inspected for this.

Note, that's the mongod may be down intentionally (`expectedToBeUp = false`) - this may happen if the process is disabled by the user. In this case the `lastMongoUpTime` is not inspected.

The PR adds unit tests only - the e2e tests will be added in the following PR

### All Submissions:


cmd/readiness/readiness_test.go
```diff
@@ -129,23 +158,35 @@ func TestPodReadiness(t *testing.T) {
 	t.Run("If replication state is not PRIMARY or SECONDARY, Pod is not ready", func(t *testing.T) {
 		assert.False(t, isPodReady(testConfig("testdata/health-status-not-readable-state.json")))
 	})
-
```
antonlisovenko, line 132: removed this block as `health-status-readable-state.json` file doesn't exist

cmd/readiness/main.go
```diff
@@ -45,59 +44,38 @@ func init() {
 // Additionally if the previous check hasn't returned 'true' the "deadlock" case is checked to make sure the Agent is
 // not waiting for the other members.
 func isPodReady(conf config.Config) bool {
-	fd, err := os.Open(conf.HealthStatusFilePath)
+	healthStatus, err := conf.HealthStatusProvider.HealthStatus()
 	if err != nil {
-		logger.Warn("No health status file exists, assuming the Automation agent is old")
```
antonlisovenko, line 50: note, that this behavior is not true anymore as the absence of the health file will be considered as an error and the pod won't be marked as "ready".
This condition here was true a couple of years ago when the support for health file was added (middle of August 2019: https://github.com/10gen/mms-automation/pull/2307) to support "old" agents.
This refers to some 3.6/4.0 version of OM which are [not supported](https://docs.google.com/spreadsheets/d/1nTvZ2cTg2Ef2XweyV1ZBb7BMSYO4eKqXszfqkpBsOPM/edit#gid=0) anymore.

So it should be safe to consider that the version of the Agent used will always eventually write the health file (usually it happens in a couple of seconds after the Agent starts).

pkg/readiness/config/health_status_provider.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+
+	"github.com/mongodb/mongodb-kubernetes-operator/pkg/readiness/health"
+)
+
+// HealthStatusProvider is an interface for providing the 'health.Status' object to the readiness probe. Abstracting out
+// this to the interface allows to perform unit-testing easier as allows to modify the config on-the-fly.
+type HealthStatusProvider interface {
+	HealthStatus() (health.Status, error)
+}
```
chatton, line 15: I think this interface may be an unnecessary level of abstraction. We can already utilize the native `io.Reader` interface to achieve these goals E.g.

```golang
func HealthStatus(reader io.Reader) (health.Status, error) {
	var health health.Status
	fd, err := os.Open(f.HealthStatusFilePath)
	if err != nil {
		return health, err
	}
	defer fd.Close()

	data, err := reader.Read(fd)
	if err != nil {
		return health, err
	}

	err = json.Unmarshal(data, &health)
	return health, err
}
```

We can provide a mock implementation by simply marshalling an instance of health.Status and returning the byte array.

```golang

mockReader := mockReader(healthStatus)
bytes, err := mockReader.Read()
var health health.Status
err = json.Unmarshal(data, &health)
```

cmd/readiness/main.go
```diff
@@ -45,59 +44,38 @@ func init() {
...
-		return true
-	}
-	defer fd.Close()
-
-	health, err := readAgentHealthStatus(fd)
-	if err != nil {
-		logger.Errorf("Failed to read agent health status file: %s", err)
-		// panicking allows to see the problem in the events for the pod (kubectl describe pod ..)
-		panic("Failed to read agent health status file: %s")
+		panic(err)
```
chatton, line 49: why panic here? Maybe we can log an error and return false? It looks like we have a mix of styles of control flow.

cmd/readiness/main.go
```diff
@@ -45,59 +44,38 @@ func init() {
...
 		return true
 	}
 
-	// If the agent has reached the goal state - returning true
-	inGoalState, err := isInGoalState(health, conf)
+	// If the agent has reached the goal state
+	inGoalState, err := isInGoalState(healthStatus, conf)
 	if err != nil {
 		logger.Errorf("There was problem checking the health status: %s", err)
 		panic(err)
```
chatton, line 62: if we change the above panic, maybe we can do the same thing here to maintain consistency. What do you think?

cmd/readiness/readiness_test.go
```diff
@@ -54,6 +55,34 @@ func TestNotReadyHealthFileHasNoProcesses(t *testing.T) {
 	assert.False(t, isPodReady(testConfig("testdata/health-status-no-processes.json")))
 }
 
+// TestNotReadyHealthFileAbsent verifies that the readiness logic panics is the health file is absent.
+// Note, that historically the absence of the file was interpreted as "ready" when older version of agents
+// could be used. But now all supported Agent versions always produce this file.
+func TestNotReadyHealthFileAbsent(t *testing.T) {
+	assert.Panics(t, func() { isPodReady(testConfig("non-existing.json")) })
```
chatton, line 62: rather than asserting panics, maybe we could even change the signature of `isPodReady` to return `(bool, error)` I believe this would be more idiomatic and also allow to differentiate between errors and `false` because we have just not reached ready state.

cmd/readiness/main.go
```diff
@@ -233,14 +211,16 @@ func isInReadyState(health health.Status) bool {
...
+		}
 
-		// Every time the process health is created by the agent,
-		// it checks if the MongoDB process is up and populates this field
-		// (https://github.com/10gen/mms-automation/blob/bb72f74a22d98cfa635c1317e623386b089dc69f/go_planner/src/com.tengen/cm/healthcheck/status.go#L43)
-		// So it's enough to check that this value is not the zero-value for int64
-
+		timeMongoUp := time.Unix(processHealth.LastMongoUpTime, 0)
+		mongoUpThreshold := time.Now().Add(-mongodNotReadyIntervalMinutes)
+		mongoIsHealthy := timeMongoUp.After(mongoUpThreshold)
```
chatton, line 221: 👍 !

chatton (commented):
Nice work this is looking great! Just had a few comments.

irajdeep:
quick question, do we want to E2E test this probe behavior?  I imagine it would be difficult to test it accurately.

cmd/readiness/main.go
antonlisovenko: `panic` is an indication of the error - this is shown when you call `kubectl describe pod`, while logging writes the data into the `readiness.log` and no one usually cares about that file.
Happy to discuss this though 👍

cmd/readiness/readiness_test.go
antonlisovenko: yes, this makes sense, agree 👍

antonlisovenko:
> quick question, do we want to E2E test this probe behavior? I imagine it would be difficult to test it accurately.

yes, it was my plan for the next PR. My plan is to `bash` into the container and move some mdb data files - this will ensure the mongod process doesn't start

cmd/readiness/main.go
antonlisovenko: just an example: this is how the `kubectl describe` looks if the readiness probe fails: 

```
...
  Normal   Started    98s               kubelet            Started container mongodb-agent-readinessprobe
  Normal   Pulling    98s               kubelet            Pulling image "registry.hub.docker.com/library/mongo:4.2.6"
  Normal   Pulled     82s               kubelet            Successfully pulled image "registry.hub.docker.com/library/mongo:4.2.6" in 16.479752s
  Normal   Created    81s               kubelet            Created container mongod
  Normal   Started    81s               kubelet            Started container mongod
  Normal   Pulling    81s               kubelet            Pulling image "quay.io/mongodb/mongodb-agent:10.29.0.6830-1"
  Normal   Pulled     69s               kubelet            Successfully pulled image "quay.io/mongodb/mongodb-agent:10.29.0.6830-1" in 12.4341038s
  Normal   Created    67s               kubelet            Created container mongodb-agent
  Normal   Started    67s               kubelet            Started container mongodb-agent
  Warning  Unhealthy  3s (x6 over 53s)  kubelet            Readiness probe failed: panic: read /var/log/mongodb-mms-automation/healthstatus/agent-health-status.json: file already closed

goroutine 1 [running]:
main.main()
  /build/main.go:218 +0x2d3

```

chatton (approved):
Great work! It looks like things ended up quite simple in the end 👍 

Thanks for making these improvements (and improvements around the community dev-notes in the process!)

irajdeep (approved):
LGTM 👍

PR #501, opened 2021-05-24
Ensure Unique Execution ID With Each Test (Unique Password)

chatton:
When running tests locally, tests could fail if re-run because the generated password already exists. This PR ensures that existing secrets don't cause test failures.


test/e2e/setup/setup.go
```diff
@@ -112,7 +114,18 @@ func GeneratePasswordForUser(mdbu mdbv1.MongoDBUser, ctx *e2eutil.Context, names
 		SetField(passwordKey, password).
 		Build()
 
-	return password, e2eutil.TestClient.Create(context.TODO(), &passwordSecret, &e2eutil.CleanupOptions{TestContext: ctx})
+	err = e2eutil.TestClient.Create(context.TODO(), &passwordSecret, &e2eutil.CleanupOptions{TestContext: ctx})
+
+	// as we just need there to be a password, it's fine to re-use an existing one (from a previous test run if it was not deleted).
+	if apiErrors.IsAlreadyExists(err) {
+		existingSecret := corev1.Secret{}
+		if err := e2eutil.TestClient.Get(context.TODO(), types.NamespacedName{Name: mdbu.PasswordSecretRef.Name, Namespace: namespace}, &existingSecret); err != nil {
```
tibulca, line 122: should we use `nps` instead of `namespace` when building the NamespacedName since this is used when building the passwordSecret (line 111)?

test/e2e/setup/setup.go
```diff
@@ -112,7 +114,18 @@ func GeneratePasswordForUser(mdbu mdbv1.MongoDBUser, ctx *e2eutil.Context, names
...
+			return "", err
+		}
+		return string(existingSecret.Data[mdbu.PasswordSecretRef.Key]), nil
```
tibulca, line 125: same here, can we use `passwordKey`  instead of `mdbu.PasswordSecretRef.Key`?

test/e2e/setup/setup.go
chatton: good catch yes we should.

test/e2e/setup/setup.go
chatton: yes I'll fix this thanks!

tibulca (approved):
LGTM

test/e2e/setup/setup.go
```diff
@@ -112,7 +114,18 @@ func GeneratePasswordForUser(mdbu mdbv1.MongoDBUser, ctx *e2eutil.Context, names
 		SetField(passwordKey, password).
 		Build()
 
-	return password, e2eutil.TestClient.Create(context.TODO(), &passwordSecret, &e2eutil.CleanupOptions{TestContext: ctx})
+	err = e2eutil.TestClient.Create(context.TODO(), &passwordSecret, &e2eutil.CleanupOptions{TestContext: ctx})
```
rodrigovalin, line 117: I'm not 100% sure about this. I think we should decide if this test wants a new password or not, or we'll depend on some kind of previous state.

Each test's executing environment should be explicit, and not have any hidden state. In the case of this snippet, we will be adding a conditioning state to the test run, possible making it not deterministic.

rodrigovalin (changes requested):
I think the correct behaviour of the test is to fail if one of the expected preconditions was not met, in this case, an environment where this particular resource (Secret object) **can be created successfully**.

Instead of this we should make sure every Secret with a password has a different name.

test/e2e/setup/setup.go
chatton: very good point about consistency, I think there's no harm in adding some unique string to the name of each secret instead of re-using existing ones. What do you think?

rodrigovalin: I think that's a much better approach.

rodrigovalin (approved):
LGTM

However, I think that the title of the PR is not what is happening right now :)

PR #502, opened 2021-05-24
Deploy all dev images with make deploy

chatton:
### All Submissions:

This PR adds functionality to deploy all of the dev images when using `make deploy`.


scripts/dev/create_kustomize_dev_yaml.py
```diff
@@ -0,0 +1,58 @@
...
+
+import sys
+import yaml
+
+from dev_config import load_config
+
+
+def main() -> int:
+    dev_config = load_config()
+    dev_yaml = {
```
chatton, line 11: kustomize does not provide a native mechanism to substitute values in patch files. This isn't an ideal solution, but I think it does exactly what we require for now.

scripts/dev/create_kustomize_dev_yaml.py
```diff
@@ -0,0 +1,60 @@
...
+
+import sys
+import yaml
+
+from dev_config import load_config
+
+
+def main() -> int:
+
+    # TODO: CLOUDP-86212 this script should be removed when Helm replaces Kustomize is introduced.
```
chatton, line 11: Helm will make this script redundant. We can remove it afterwards.

PR #507, opened 2021-05-26
CLOUDP-89260: e2e test for readiness probe

antonlisovenko:
This is an e2e test for readiness improvements. 

The test executes into the container, moves the important data file and kills the mongod process (otherwise mongod will work fine until any data is written).

Then the test makes sure the `mongodb-agent` container becomes not ready.
Finally the test fixes the situation by moving the file back

**Update** I tried to set the `failureThreshold` to a much lower value (4) but this broke upgrade procedures, so I had to keep it high (though decreased from 60 to 40)

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -163,7 +163,8 @@ task_groups:
     - func: create_kind_cluster
   tasks:
     - e2e_test_replica_set
-    - e2e_test_replica_set_readiness_probe
+    - e2e_test_replica_set_recovery
+    - e2e_test_replica_set_mongod_readiness
     - e2e_test_replica_set_scale
     - e2e_test_replica_set_scale_down
     - e2e_test_replica_set_change_version
@@ -262,11 +263,17 @@ tasks:
         vars:
           test: replica_set
 
-  - name: e2e_test_replica_set_readiness_probe
+  - name: e2e_test_replica_set_recovery
     commands:
       - func: run_e2e_test
         vars:
-          test: replica_set_readiness_probe
+          test: replica_set_recovery
+
+  - name: e2e_test_replica_set_mongod_readiness
+    commands:
+      - func: run_e2e_test
+        vars:
+          test: replica_set_mongod_readiness
 
   - name: e2e_test_replica_set_scale
     commands:
--- a/test/e2e/client.go
+++ b/test/e2e/client.go
@@ -159,7 +159,7 @@ func (c *E2ETestClient) Execute(pod corev1.Pod, containerName, command string) (
 	errBuf := &bytes.Buffer{}
 	exec, err := remotecommand.NewSPDYExecutor(c.restConfig, "POST", req.URL())
 	if err != nil {
-		panic(err)
+		return "", err
 	}
 	err = exec.Stream(remotecommand.StreamOptions{
 		Stdout: buf,
--- a/test/e2e/client.go
+++ b/test/e2e/client.go
@@ -1,15 +1,19 @@
 package e2eutil
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"testing"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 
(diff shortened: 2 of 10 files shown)


.evergreen.yml
```diff
@@ -262,11 +263,17 @@ tasks:
         vars:
           test: replica_set
 
-  - name: e2e_test_replica_set_readiness_probe
```
antonlisovenko, line 265: as discussed with Cian - have renamed this test to match it's logic better

test/e2e/client.go
```diff
@@ -79,6 +83,10 @@ func (ctx *Context) AddCleanupFunc(fn func() error) {
 // E2ETestClient is a wrapper on client.Client that provides cleanup functionality.
 type E2ETestClient struct {
 	Client client.Client
+	// We need the core API client for some operations that the controller-runtime client doesn't support
+	// (e.g. exec into the container)
+	CoreV1Client corev1client.CoreV1Client
```
irajdeep, line 88: 👍

test/e2e/client.go
```diff
@@ -127,6 +139,42 @@ func (c *E2ETestClient) Get(ctx context.Context, key types.NamespacedName, obj c
...
+	exec, err := remotecommand.NewSPDYExecutor(c.restConfig, "POST", req.URL())
+	if err != nil {
+		panic(err)
+	}
+	err = exec.Stream(remotecommand.StreamOptions{
+		Stdout: buf,
+		Stderr: errBuf,
+	})
+	if err != nil {
+		return "", fmt.Errorf(`failed executing command "%s" on %v/%v: %s ("%s")`, command, pod.Namespace, pod.Name, err, errBuf.String())
```
irajdeep, line 169: Q: Wondering if we need to return the `errBuf` as well, `err` should contain it right?

test/e2e/e2eutil.go
```diff
@@ -140,6 +140,22 @@ func waitForStatefulSetCondition(t *testing.T, mdb *mdbv1.MongoDBCommunity, retr
 	})
 }
 
+func WaitForPodReadiness(t *testing.T, isReady bool, containerName string, timeout time.Duration, pod corev1.Pod) error {
```
irajdeep, line 143: Q: wondering if `timeout` needs to be a parameter, time for the pod to reach ready should be irrespective of the caller? -- maybe we can assume some sane defaults.

irajdeep (commented):
Looks great to me. Added some minor comments but nothing blocking.

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -359,6 +353,27 @@ func StatefulSetContainerConditionIsTrue(mdb *mdbv1.MongoDBCommunity, containerN
 	}
 }
 
+func PodContainerBecomesNotReady(mdb *mdbv1.MongoDBCommunity, podNum int, containerName string, timeout time.Duration) func(*testing.T) {
```
irajdeep, line 356: nit: sake of consistency in the file maybe add a comment to the public methods.

test/e2e/client.go
irajdeep: No strong opinion here, but if you think `errBuf` doesn't have any explicit useful info, I am slightly in favor of just returning the `err`.

test/e2e/e2eutil.go
antonlisovenko: I agree with you in general though this method has a caveat that it's called from both `PodContainerBecomesReady` and `PodContainerBecomesNotReady` (so both directions: true->false and false->true). So the former passes 3 minutes and the latter - 10 as we have a very long `failureThreshold` for readiness probe. We could try to default each of those two methods instead?

irajdeep: > We could try to default each of those two methods instead?

+1

irajdeep (approved):
Nice one! LGTM.

test/e2e/client.go
```diff
@@ -127,6 +139,42 @@ func (c *E2ETestClient) Get(ctx context.Context, key types.NamespacedName, obj c
...
+			Stdout:    true,
+			Stderr:    true,
+			TTY:       true,
+		}, scheme.ParameterCodec)
+
+	buf := &bytes.Buffer{}
+	errBuf := &bytes.Buffer{}
+	exec, err := remotecommand.NewSPDYExecutor(c.restConfig, "POST", req.URL())
+	if err != nil {
+		panic(err)
```
chatton, line 162: [q] why panic here when error is already part of method signature, is there a specific reason to panic instead of return the error?

chatton (approved):
Great work! I just had a few additional comments, but no blockers

test/e2e/client.go
antonlisovenko: good catch! Just bad copy-paste 🤦

test/e2e/replica_set_mongod_readiness/replica_set_mongod_readiness_test.go
```diff
@@ -0,0 +1,56 @@
...
+	defer ctx.Teardown()
+
+	mdb, user := e2eutil.NewTestMongoDB(ctx, "mdb0", "")
+
+	_, err := setup.GeneratePasswordForUser(ctx, user, "")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = FromResource(t, mdb)
```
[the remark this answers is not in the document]
antonlisovenko: ok 👍

PR #514, opened 2021-05-30
Add GitHub Action to Release All Images

chatton:
This PR adds a GitHub Action which builds and pushes all of our images to Quay. The release process is as follows.

* Manually trigger the `Create Release PR` action (master branch release.json should contain the desired versions, any that do not exist, will be built and released. no-op for already released images.)
* Approve PR once all tasks have run. (triggers github release to be drafted, release text comes from docs/RELEASE_NOTES.md)
* Publish draft (Release PR will be auto-merged and deleted)

This PR also includes a bump for the operator to version 0.6.1, so when the release task is called we can release this new operator version.

The evergreen release tasks have been removed.


scripts/ci/determine_required_releases.py
```diff
@@ -0,0 +1,61 @@
...
+    release["mongodb-agent"] = release["mongodb-agent"]["version"]
+    return release
+
+
+def main() -> int:
+    if len(sys.argv) != 2:
+        raise ValueError("usage: determine_required_releases.py [image-name]")
+    image_name_map = _load_image_name_to_version_map()
+
+    if sys.argv[1] not in image_name_map:
```
tibulca, line 42: can we define a variable image_name = sys.argv[1] and use it in all 6 places?

chatton: yes, good idea 👍

scripts/ci/add_supported_release.py
```diff
@@ -72,18 +69,23 @@ def add_release_version(image: str, version: str) -> None:
...
 
-    if args.image not in VALID_IMAGES:
-        raise ValueError("Image {} not supported".format(args.image))
+    if args.image_name not in VALID_IMAGES:
+        print(
+            "Image {} not supported. Not adding release version.".format(
+                args.image_name
+            )
+        )
+        return 0
```
tibulca, line 82: do we want to exit here with code 0 or an error code?

rodrigovalin (approved):
LGTM

scripts/ci/add_supported_release.py
chatton: right good catch, the reason for a 0 here rather than error, is that we run this task for every image we are going to release. This will update the release as a support version (for agent only currently), but we need this to be a no-op rather than failure for the other images.

This task will currently only fail if there is an error when interacting with the release database in Atlas.

tibulca (approved):

PR #518, opened 2021-05-31
add detailed upgrade steps

### LINKED ISSUE #352: upgrade documentation is way too sparse
I got a operator instance running (v0.2.2) with a production mongoDB cluster.
I tried to figure out what to do to upgrade to latest 0.5.1 in my test environment, but aparently there were several versions needing manual steps and no real documentation for the whole thing.

**What did you expect?**
Documentation describing the different migration steps needed.

**What happened instead?**
Some unclear Notes about mongo resource requiring a new value and "we renamed the resource".

**Operator Information**
 - Operator Version 0.2.2
 - MongoDB Image used 4.2.10

fireba11:
closes #352

### DIFF
--- a/docs/install-upgrade.md
+++ b/docs/install-upgrade.md
@@ -155,12 +155,47 @@ To install the MongoDB Community Kubernetes Operator:
 
 ## Upgrade the Operator
 
-To upgrade the MongoDB Community Kubernetes Operator to a specific version:
+The release v0.6.0 had some breaking changes (see https://github.com/mongodb/mongodb-kubernetes-operator/releases/tag/v0.6.0) requiring special steps to upgrade from a pre-0.6.0 Version.
+As always, have backups.
+Make sure you run commands in the correct namespace.
 
-1. Change the version in the Operator [resource definition](../config/manager/manager.yaml)
-2. Change to the directory in which you cloned the repository.
-3. Checkout the specific tag matching the operator version (e.g. `v0.5.1`)
-4. Invoke the following `kubectl` command to upgrade the [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/).
-   ```
-   kubectl apply -f config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
-   ```
+1. Prepare for the upgrade.
+
+   a. Migrate your cr by updating apiVersion and kind to
+      ```
+      apiVersion: mongodbcommunity.mongodb.com/v1
+      kind: MongoDBCommunity
+      ```
+      If you upgrade from pre-0.3.0 you need to also add the field spec.users[n].scramCredentialsSecretName for each resource. This will be used to determine the name of the generated secret which stores MongoDB user credentials. This field must comply with DNS-1123 rules (must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character).
+   b. Plan a downtime.
+2. Out with the old
+
+   a. Delete the old operator.
+      ```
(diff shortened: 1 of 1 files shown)


docs/install-upgrade.md
```diff
@@ -155,12 +155,47 @@ To install the MongoDB Community Kubernetes Operator:
...
-   kubectl apply -f config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
-   ```
+1. Prepare for the upgrade.
+
+   a. Migrate your cr by updating apiVersion and kind to
+      ```
+      apiVersion: mongodbcommunity.mongodb.com/v1
+      kind: MongoDBCommunity
+      ```
+      If you upgrade from pre-0.3.0 you need to also add the field spec.users[n].scramCredentialsSecretName for each resource. This will be used to determine the name of the generated secret which stores MongoDB user credentials. This field must comply with DNS-1123 rules (must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character).
```
rodrigovalin, line 169: Can you reference https://kubernetes.io/docs/concepts/overview/working-with-objects/names/ and maybe remove the parenthesis comment.

docs/install-upgrade.md
```diff
@@ -155,12 +155,47 @@ To install the MongoDB Community Kubernetes Operator:
...
+      kubectl delete statefulsets.apps mongodb
+      ```
+   c. Delete the old customResourceDefinition. Not strictly needed but no need to keep it around anymore (unless you got more installations of operator in your cluster)
+      ```
+      kubectl delete crd mongodb.mongodb.com
+      ```
+3. In with the new
+
+   Follow the normal installation procedure above.
+4. Start up your cluster again
```
rodrigovalin, line 188: Maybe specify that you are referring to the MongoDB cluster or Replica Set. `cluster` can be confused with "Kubernetes cluster".

docs/install-upgrade.md
```diff
@@ -155,12 +155,47 @@ To install the MongoDB Community Kubernetes Operator:
...
+   a. Re-create your cr using the new Version from Step 1.a
+   b. Patch your statefulset to have it update the permissions
```
rodrigovalin, line 190: Is this something specific you need for your environment? The Operator should create the StatefulSet with the right SecurityContext at this stage, and I'm not sure why you would need to patch it manually.

rodrigovalin (commented):
Hello and thanks for your contribution!

This is looking really good; I have left some comments and I would really appreciate if you could address them.

Thanks again

docs/install-upgrade.md
fireba11: The patch is from release notes of v0.6.0 to fix filesystem rights on the PV.
It is actually needed for mongod to be able to start with the old data.

docs/install-upgrade.md
fireba11: Replaced the text in parenthesis with the reference link.

docs/install-upgrade.md
fireba11: changed

rodrigovalin (approved):
Thanks for your contribution @fireba11

rodrigovalin:
@fireba11 we only need for your commits to be signed before being able to merge the patch.

fireba11:
@rodrigovalin i realized. first time that was required, i just set up a key to do that ... and rebase the commits to sign them (at least that's the way i found to retroactively do that). hope i did everything correct :-)

bznein:
Hi @fireba11 

the signing looks correct now! Will merge tomorrow after all the required checks pass :)

Thanks again for your contribution!

PR #520, opened 2021-06-01
Fixed image_name field

chatton:


bznein (approved):
Thanks for the quick fix LGTM!

PR #521, opened 2021-06-01
Create role/rolebinging in namespace configured in config.json and create namespace if it doesn't exist.

irajdeep:
### All Submissions:

### DIFF
--- a/scripts/dev/e2e.py
+++ b/scripts/dev/e2e.py
@@ -49,8 +49,12 @@ def _prepare_test_environment(config_file: str) -> None:
     )
 
     print("Creating Role Binding")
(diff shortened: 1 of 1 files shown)


chatton (commented):
Nice fix, it looks like `_load_test_service_account()`  is still using whatever ns is configured in your current context and not the one coming from your dev config. I think we can update that too. What do you think? 

Other than that looks good!

chatton (approved):
It looks like we already have that covered when we create the service account LGTM

irajdeep:
@chatton I think we have that case covered since while creating the SA we pass the namespace.

PR #522, opened 2021-06-01
Add GitHub action to create release pr

chatton:
This PR introduces a new Github Action which can be run manually to create a release PR.

This PR will increment the versions in the provided yaml manifests with the values from release.json. 

There are no semantical changes to the deployment manifests, the differences are just the result of running the script.


config/manager/manager.yaml
```diff
@@ -13,31 +12,31 @@ spec:
...
+        - name: READINESS_PROBE_IMAGE
+          value: quay.io/mongodb/mongodb-kubernetes-readinessprobe:1.0.4
+        - name: MONGODB_IMAGE
+          value: library/mongo
+        - name: MONGODB_REPO_URL
+          value: registry.hub.docker.com
+        image: quay.io/mongodb/mongodb-kubernetes-operator:0.6.0
+        imagePullPolicy: Always
+        name: mongodb-kubernetes-operator
+      serviceAccountName: mongodb-kubernetes-operator
```
chatton, line 42: the new release tool was run in order to prevent differences in formatting going forward, there are no changes in versions here.

scripts/ci/update_release.py
```diff
@@ -0,0 +1,65 @@
...
+RELATIVE_PATH_TO_MANAGER_YAML = "config/manager/manager.yaml"
+RELATIVE_PATH_TO_OPENSHIFT_MANAGER_YAML = "deploy/openshift/operator_openshift.yaml"
+
+
+def _load_yaml_file(path: str) -> Dict:
+    with open(path, "r") as f:
+        return yaml.safe_load(f.read())
+
+
+def _dump_yaml(operator: Dict, path: str) -> None:
```
rodrigovalin, line 17: return type is not None

scripts/ci/update_release.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def _load_release() -> Dict:
+    with open("release.json", "r") as f:
+        return json.loads(f.read())
+
+
+def _replace_tag(image: str, new_tag: str) -> str:
+    split_image = image.split(":")
+    return f"{split_image[0]}:{new_tag}"
```
rodrigovalin, line 36: please avoid Python expressions in f-strings. I think it is better to do:

`split_image[0] + ":" + new_tag` (it can't get easier to read than this)

.github/workflows/create_release_pr.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.6'
+          architecture: 'x64'
+      - run: pip install -r requirements.txt
+      - name: Create Release PR
+        uses: technote-space/create-pr-action@v2
+        with:
+          EXECUTE_COMMANDS: |
+            python scripts/ci/update_release.py
```
rodrigovalin, line 27: the changes that you do in this file will not be staged at this point, will they be incorporated into the patch?

scripts/ci/update_release.py
chatton: you are correct! Thanks for catching that

.github/workflows/create_release_pr.yml
```diff
@@ -0,0 +1,33 @@
+name: Create Release PR
+on:
+  workflow_dispatch
+
+jobs:
+  CreateReleasePR:
```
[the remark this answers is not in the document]
chatton: It looks like in our other actions, the job names specifically do not use kebab case, however I am in favour of making that transition.

.github/workflows/create_release_pr.yml
chatton: The github action used here will stage the changes after the commands are executed. The only input we need to provide is the commands that make changes.

rodrigovalin (approved):
LGTM

PR #523, opened 2021-06-02
CLOUDP-89492 Expose connection string for deployments for easy of use

tibulca:


controllers/mongodb_users.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+
+	mdbv1 "github.com/mongodb/mongodb-kubernetes-operator/api/v1"
+	"github.com/mongodb/mongodb-kubernetes-operator/pkg/kube/secret"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (r ReplicaSetReconciler) ensureUserResources(mdb mdbv1.MongoDBCommunity) error {
```
bznein, line 12: [q] can we add short docstring for every function introduced in this PR? (exported ones especially, but I would add a comment for each function in general)

controllers/mongodb_users.go
```diff
@@ -0,0 +1,51 @@
...
+	for _, user := range mdb.GetScramUsers() {
+		secretNamespacedName := types.NamespacedName{Name: user.PasswordSecretName, Namespace: mdb.NamespacedName().Namespace}
```
bznein, line 14: [nit] we can just do `mdb.Namespace` instead of passing through `NamespacedName()`

controllers/mongodb_users.go
```diff
@@ -0,0 +1,51 @@
...
+		if _, err := secret.ReadKey(r.client, user.PasswordSecretKey, secretNamespacedName); err != nil {
+			if apiErrors.IsNotFound(err) {
+				r.log.Errorf(`User password secret "%s" not found`, secretNamespacedName)
```
bznein, line 17: I'm not sure about logging here, we will return the error anyway at we will both see an error and a log message?

What do you think if we returned fmt.Errorf(...) here?

controllers/replica_set_controller.go
```diff
@@ -218,6 +228,10 @@ func (r ReplicaSetReconciler) Reconcile(ctx context.Context, request reconcile.R
 		return res, err
 	}
 
+	if err := r.updateConnectionStringSecrets(mdb); err != nil {
+		r.log.Errorf("Could not update connection string secrets: %s", err)
```
bznein, line 232: If I remember correctly, error messages should not start with an uppercase letter

controllers/mongodb_users.go
```diff
@@ -0,0 +1,51 @@
...
+func (r ReplicaSetReconciler) updateConnectionStringSecrets(mdb mdbv1.MongoDBCommunity) error {
+	for _, user := range mdb.GetScramUsers() {
+		secretNamespacedName := types.NamespacedName{Name: user.PasswordSecretName, Namespace: mdb.NamespacedName().Namespace}
+		pwd, err := secret.ReadKey(r.client, user.PasswordSecretKey, secretNamespacedName)
+		if err != nil {
+			return err
+		}
+
+		operatorSecret := secret.Builder().
+			SetName(fmt.Sprintf("mdbc-%s-%s", user.Database, user.Username)).
```
bznein, line 37: I almost think we could extract this `fmt.Sprintf` call to a method inside `user`. We also duplicate this in the testing. What do you think?

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -202,6 +215,13 @@ func CreateMongoDBResource(mdb *mdbv1.MongoDBCommunity, ctx *e2eutil.Context) fu
 	}
 }
 
+func GetConnectionStringForUser(mdb mdbv1.MongoDBCommunity, username string, database string) string {
```
bznein, line 218: [q] why are we passing username and database here? can't we directly pass a scram user?

bznein (commented):
Looking good, nice addition! Just a few comments here and there but great work!

controllers/mongodb_users.go
tibulca: done

controllers/mongodb_users.go
tibulca: done

controllers/mongodb_users.go
tibulca: done

controllers/mongodb_users.go
tibulca: implemented, good idea

test/e2e/mongodbtests/mongodbtests.go
tibulca: passing now the scram user (since I created that method inside the user)

controllers/replica_set_controller.go
tibulca: this is a log and there are many (10+) similar error logs in replica_set_controller.go
I can update all to use the lowercase format if this is part of our style guide

api/v1/mongodbcommunity_types.go
```diff
@@ -455,6 +451,31 @@ func (m MongoDBCommunity) MongoSRVURI() string {
...
+func (m MongoDBCommunity) MongoAuthUserURI(user scram.User, password string) string {
+	return fmt.Sprintf("mongodb://%s:%s@%s/%s?ssl=%t",
+		url.QueryEscape(user.Username),
+		url.QueryEscape(password),
+		strings.Join(m.Hosts(), ","),
+		user.Database,
+		m.Spec.Security.TLS.Enabled)
+}
+
+// MongoSRVURI returns a mongo srv uri which can be used to connect to this deployment
```
bznein, line 465: inconsistent function name in this docstring

controllers/mongodb_users.go
```diff
@@ -0,0 +1,56 @@
...
+// The client applications can mount these secrets and connect to the mongodb cluster
+func (r ReplicaSetReconciler) updateConnectionStringSecrets(mdb mdbv1.MongoDBCommunity) error {
+	for _, user := range mdb.GetScramUsers() {
+		secretNamespacedName := types.NamespacedName{Name: user.PasswordSecretName, Namespace: mdb.Namespace}
+		pwd, err := secret.ReadKey(r.client, user.PasswordSecretKey, secretNamespacedName)
+		if err != nil {
+			return err
+		}
+
+		operatorSecret := secret.Builder().
```
bznein, line 40: [nit] operatorSecret doesn't sound like the best name here, it's not closely related to the operator as much as it is related to the scram user

controllers/replica_set_controller.go
bznein: Maybe let's leave it for next time :) I think this is more for when returning errors

pkg/authentication/scram/scram.go
```diff
@@ -336,3 +337,8 @@ func convertMongoDBUserToAutomationConfigUser(secretGetUpdateCreateDeleter secre
 	acUser.ScramSha256Creds = &sha256Creds
 	return acUser, nil
 }
+
+// GetConnectionStringSecretName returns the name of the secret where the operator stores the connection string for current user
+func (u User) GetConnectionStringSecretName() string {
+	return fmt.Sprintf("mdbc-%s-%s", u.Database, u.Username)
+}
```
bznein, line 344: I have a question here: could we have a case here where an user has two MongoDBCommunity resources: mdbcA and mdbcB, both with username `foo` on database `foo`, but different passwords, would this generate conflicting secrets?

bznein (commented):
I think we are almost there, just a few more questions :)

controllers/mongodb_users.go
tibulca: renamed to connectionStringSecret

api/v1/mongodbcommunity_types.go
tibulca: fixed, thanks

pkg/authentication/scram/scram.go
tibulca: I included the resource name and also implemented spec validation to avoid name collision

irajdeep (approved):
LGTM 👍

bznein (approved):
LGTM I think we should document it somewhere though! Ok to do in a separate PR

controllers/replica_set_controller.go
```diff
@@ -457,14 +471,15 @@ func buildService(mdb mdbv1.MongoDBCommunity) corev1.Service {
...
+// If there is no a previous Spec, then the function assumes this is the first version
+// and runs the initial spec validations otherwise it validates that the new Spec,
+// corresponding to the existing one is still valid
+func (r ReplicaSetReconciler) validateSpec(mdb mdbv1.MongoDBCommunity) error {
 	lastSuccessfulConfigurationSaved, ok := mdb.Annotations[lastSuccessfulConfiguration]
 	if !ok {
-		// First version of Spec, no need to validate
-		return nil
+		// First version of Spec
+		return validation.ValidateInitalSpec(mdb)
```
chatton, line 482: This validation will only apply a single time, (the first reconciliation) so we'll be skipping this reconciliation every other time. The other validation (ValidateUpdate) only makes sense to happen all times but the first, but here I believe we want to check this every reconciliation.

chatton (commented):
Nice job this is great work! I only had one concern, everything else LGTM.

controllers/replica_set_controller.go
tibulca: you're right since users can be also added after the initial reconcile process was successful.
I made the small change to validate users in both scenarios

chatton (approved):
LGTM! nice work!

bznein:
We should also add it to the Release Notes!

PR #544, opened 2021-06-09
Remove Dockerfile Generator

chatton:
The Dockerfile generator is no longer need (functionality now exists within sonar/pipeline)


bznein (approved):
LVGTM!

PR #551, opened 2021-06-10
Move Wait functions into separate package

chatton:
This PR moves waiting related code out of the generic `util` package into the new `wait` package.

### DIFF
--- a/test/e2e/util/wait/wait.go
+++ b/test/e2e/util/wait/wait.go
@@ -1,0 +1,138 @@
+package wait
+
+import (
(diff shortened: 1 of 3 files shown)


test/e2e/util/wait/wait.go
```diff
@@ -0,0 +1,138 @@
+package wait
```
chatton, line 1: code moved as is, with the `Wait` prefix removed from function names.

tibulca (approved):

PR #552, opened 2021-06-10
Include MongoDBCommunity resource spec in diagnostics

chatton:
We were previously only dumping the StatefulSet spec, not the MongoDBCommunity spec in our e2e tests.


bznein (approved):
LGTM great idea

PR #556, opened 2021-06-11
Added Single Image Github Action

chatton:
This is largely a duplicate of the release all images github action that was added recently. 

This action allows us to manually run any image providing the pipeline argument and the key in release.json


rodrigovalin (approved):
LGTM

PR #560, opened 2021-06-13
Run E2E tests with GitHub actions

chatton:
Since we moved our release / image build process to use Github Actions, I think it makes sense to move our e2e test suite to use Github Actions too.

There are few important things to mention about the changes made in this PR.

1. The E2E tests here are contained with a **single workflow**. Currently, it is not possible to re-run a single job (in this case an **e2e test**), only a single workflow i.e. all of the tests. A workaround was put in place to cache the result of successful tests so that upon re-run, they are skipped.
2. We previously used a python script to dump diagnostic files, I replaced this with bash (I'm sorry!). The reasoning behind this is that the direct kubectl output of `-o yaml` and `describe` is what we are used to having. Just fetching the pod/sts manifests does not include all of the information that `kubectl describe` does (events etc. ) I am very open to discussing this, I would rather have a non-bash solution if possible!
3. Evergreen.yml and all ci scripts specific to evg have been removed.
4. Test diagnostics are uploaded on **failure only**. The way test artefacts are presented does not make it easy to link a given artefact to a given test run (without manually searching for the name). This decision was to increase ease of use.

Other options.

To get around the test artefact and re-run `hack`, another option is to make each **test** a single workflow. At this time there is currently no way of templating github action jobs or workflows to achieve this in a way that doesn't involve creating a massive amount of  workflows with duplicated steps. (only difference would be the string passed to e2e.py). However as it is not possible to depend on a job from another workflow, there would need to be some workaround implemented to ensure we don't re-build all the images for each test.


rodrigovalin (approved):
LGTM

I think we have to remove Evergreen integration from [here](https://github.com/mongodb/mongodb-kubernetes-operator/settings/hooks).

irajdeep (approved):
LGTM -- I think there is some pending cleanup? -- failing some tests

irajdeep:
Also one question -- with evergreen we could run one of the tests with evergreen cli. Is it possible with GH actions?

chatton:
@irajdeep, the pending task is just the failing evg ones (as there is no evergreen.yml anymore)

Regarding running tasks with cli, I will add a new workflow that we can manually trigger to achieve the same thing!

PR #561, opened 2021-06-14
Fixing the small issue with *

killian2k:
Added quotes to * because it created an error without it.


chatton (approved):
LGTM! Thanks for the fix.

PR #564, opened 2021-06-14
Add Manually Triggerable E2E Workflow

chatton:
This PR adds a workflow that we can trigger manually to run a single test.

Note: Until the workflow is merged into master it is not possible to actually trigger it! I will test it once merged and create a follow up PR if necessary.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -20,6 +20,11 @@ jobs:
           - pipeline-argument: agent-ubuntu
           - pipeline-argument: e2e
     steps:
+      - name: Cancel Previous Runs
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e.yml
```diff
@@ -20,6 +20,11 @@ jobs:
           - pipeline-argument: agent-ubuntu
           - pipeline-argument: e2e
     steps:
+      - name: Cancel Previous Runs
```
chatton, line 23: cancel existing runs on this branch when we pushed new code

irajdeep (approved):
LGTM 👍

PR #576, opened 2021-06-19
Run GithubActions correctly from forks (Templating of GithubActions)

chatton:
One big limitation of github actions is the lack of reusing existing components. There is no way to reuse `jobs`, `steps` or `events` between workflow files. Which can result in a lot of duplication and error prone workflow files when a single change results in needing to change many files.

`composite` jobs exist, but these are limited to only running shell scripts (not using other github actions)

This PR refactors our Github Actions to allow templating of shared `jobs`, `steps` and `events` (`on` section of workflow file) using [this library](https://github.com/chatton/GithubActionTemplates). 

The diff is large, but this is primarily due to auto generation of our github actions.

The following files will been genereted upon running the pre-commit hook

* `.github/workflows/e2e.yaml`
* `.github/workflows/e2e-fork.yaml`
* `.github/workflows/e2e-dispatch.yaml`

The contents of our e2e workflows have been extracted out as per the README in the above link. They are now under `.action_templates`

The pre-commit hook has been updated to generate these workflows. Any changes should happen through these templates.

Notes on forks.

**This will allow us to run tests for the dependabot PRs.**

In order to run tests correctly on forks, we need to ensure the following things are true.

* The `event` triggering the workflow must be `pull_request_target`. This ensures that the github secrets are passed (otherwise the test will fail every time)
* We checkout code from the **fork**.
* We ensure that the workflow requires a manual trigger. In this case, adding the `safe-to-test` label will trigger tests from contributors using forks. 

The new `e2e-fork.yaml` workflow is almost identical to the existing `e2e.yaml` except for these three differences.

The introduced templating allows for re-use of all of the existing components.


.github/workflows/e2e-dispatch.yml
```diff
@@ -1,103 +1,115 @@
...
+# 
+# This file is automatically generated using templates. Changes to this file
+# should happen through editing the templates under .action_templates/*
+# Manual edits will be overwritten.
+#
+##################################################################################
+
 name: Run Single E2E
 on:
+  # template: .action_templates/events/single-e2e-workflow-dispatch.yaml
```
chatton, line 12: template path is added to each section to aid in debugging/clarity

.github/workflows/e2e-fork.yml
```diff
@@ -0,0 +1,237 @@
...
+      fail-fast: false
+      matrix:
+        include:
+        - pipeline-argument: operator-ubi
+        - pipeline-argument: version-post-start-hook-init
+        - pipeline-argument: readiness-probe-init
+        - pipeline-argument: agent-ubi
+        - pipeline-argument: agent-ubuntu
+        - pipeline-argument: e2e
+    if: github.event.pull_request.head.repo.full_name != 'mongodb/mongodb-kubernetes-operator'
```
chatton, line 33: we only run this workflow if it is being triggered by a fork

.github/workflows/e2e-fork.yml
```diff
@@ -0,0 +1,237 @@
...
+    - name: Cancel Previous Runs
+      uses: styfle/cancel-workflow-action@0.9.0
+      with:
+        access_token: ${{ github.token }}
+    # template: .action_templates/steps/checkout-fork.yaml
+    - name: Checkout Code
+      uses: actions/checkout@v2
+      with:
+        ref: ${{github.event.pull_request.head.ref}}
+        repository: ${{github.event.pull_request.head.repo.full_name}}
```
chatton, line 46: we checkout the fork, not this repo.

.github/workflows/e2e.yml
```diff
@@ -1,218 +1,237 @@
...
-          - pipeline-argument: agent-ubi
-          - pipeline-argument: agent-ubuntu
-          - pipeline-argument: e2e
+        - pipeline-argument: operator-ubi
+        - pipeline-argument: version-post-start-hook-init
+        - pipeline-argument: readiness-probe-init
+        - pipeline-argument: agent-ubi
+        - pipeline-argument: agent-ubuntu
+        - pipeline-argument: e2e
+    if: github.event.pull_request.head.repo.full_name == 'mongodb/mongodb-kubernetes-operator'
```
chatton, line 40: run only on **this** repo

.github/workflows/release-images.yml
```diff
@@ -85,7 +86,7 @@ jobs:
     - name: Create Github Release
       uses: ncipollo/release-action@v1
       with:
-        tag: ${{ steps.release_tag.outputs.OUTPUT }}
+        tag: "v${{ steps.release_tag.outputs.OUTPUT }}"
```
chatton, line 89: unrelated fix: we had the tag incorrect.

chatton (commented):
I've added comments on all of the differences introduced in the actual workflow files, everything else is format/whitespace changes due to the auto-generation, and the extracted template files.

.github/workflows/e2e-fork.yml
```diff
@@ -0,0 +1,237 @@
...
+
+name: Run E2E Fork
+on:
+  # template: .action_templates/events/labeled-pull-request-target.yaml
+  pull_request_target:
+    types: [labeled]
+    branches:
+    - master
+    paths-ignore:
+    - docs/**
```
chatton, line 18: The only way a pull request will run on a fork, is if someone with write access labels it.

.action_templates/e2e-fork-template.yaml
```diff
@@ -0,0 +1,25 @@
+name: Run E2E Fork
+jobs:
+  - template: setup
+    if: github.event.pull_request.head.repo.full_name != 'mongodb/mongodb-kubernetes-operator' && contains(github.event.pull_request.labels.*.name, 'safe-to-test')
```
chatton, line 4: the condition for forks, we run only if it has been labeled and is coming from a different repo.

.action_templates/e2e-pr-template.yaml
```diff
@@ -0,0 +1,27 @@
+name: Run E2E
+jobs:
+  - template: setup
+    if: github.event.pull_request.head.repo.full_name == 'mongodb/mongodb-kubernetes-operator'
```
chatton, line 4: this will run only if it is a branch from this repo.

irajdeep (approved):
My knowledge on GH actions is limited -- I assume it works and will approve : -)

Issue #217, opened 2020-10-09
scram-sha-1 User


rgherta:
How can we create users with  SCRAM-SHA-1 authentication mechanism?

irajdeep:
Hi @rgherta 
does this documentation: https://github.com/mongodb/mongodb-kubernetes-operator/blob/master/docs/users.md help?

rgherta:
Hello @irajdeep 

Current version allows only SHA-256 that is better secured, it makes sense, but still the same mongo version supports both.

We must add this legacy SHA-1 hash method to the cluster-configuration.json file that is mounted to the mongodb-agent
It looks like we must add this item "SCRAM-SHA-1" in the 2 arrays created by scram_enabler.go but I do not know if there is some other logic in the middle of this.

andreasevers:
This workaround mentioned by @chatton can be automated using a MutatingAdmissionWebhook.
This repository does just that: https://github.com/andreasevers/mongodb-operator-scram-sha-1-mutatingadmissionwebhook

lchdev:
Hi all,

Bumping this topic to know if there are any plan to add SCRAM-SHA-1 support directly in the operator ? Honestly, since this auth mechanism is supported and not deprecated by MongoDB, I don't this why the operator should prevent its use.

Support for SCRAM-SHA-1 would help in migrating old applications not yet supporting SCRAM-SHA-256. Relying on a mutating webhook workaround feels a bit odd.

Thanks

rgherta:
I say the operator will see a better adoption by legacy systems running in private clusters.

bodomic2:
I guess I'm the first to report that feature is still not supported in 0.7.0 even if announced in release notes...

```
{"t":{"$date":"2021-07-12T17:22:28.484+00:00"},"s":"I",  "c":"ACCESS",   "id":20249,   "ctx":"conn35649","msg":"Authentication failed","attr":{"mechanism":"SCRAM-SHA-1","principalName":"rt","authenticationDatabase":"rt","client":"100.64.0.30:44118","result":"BadValue: SCRAM-SHA-1 authentication is disabled"}}
```

The error message is the very same as before upgrade.

```
Containers:
  mongodb-kubernetes-operator:
    Container ID:  containerd://8401da467efd26639f12dc61d8a46eae957477fe07d23376b49e5af9eb90fa7d
    Image:         quay.io/mongodb/mongodb-kubernetes-operator:0.7.0
[...]

PR #598, opened 2021-06-28
Cloudp 78288 scram sha 1 user

killian2k:
### All Submissions:

* [x] Have you signed all of your commits?


killian2k (commented):
We discussed with Cian how to improve readability of the test files. That's how I ended up with authOptions. We also discussed the fact that a single function runs all the tests one after the other and modify the created resources instead of creating new ones.

api/v1/mongodbcommunity_types.go
```diff
@@ -402,12 +423,39 @@ func (m MongoDBCommunity) GetScramOptions() scram.Options {
 		ignoreUnknownUsers = *m.Spec.Security.Authentication.IgnoreUnknownUsers
 	}
 
+	//Manage the Auth
```
chatton, line 426: this comment is a little vague, I don't think it provides much value, we should include comments if it is not immediately clear what the code is doing.

.github/workflows/e2e.yml
```diff
@@ -126,6 +126,8 @@ jobs:
           distro: ubuntu
         - test-name: replica_set_arbiter
           distro: ubuntu
+        - test-name: replica_set_authentication
+          distro: ubuntu
```
chatton, line 130: 👍

chatton: Ah! I don't see the corresponding changes the workflow files, have you set up the pre-commit hook? (see the end of contributing.md)

docs/RELEASE_NOTES.md
```diff
@@ -5,6 +5,7 @@
 - Changes
   - Members of a Replica Set can be configured as arbiters.
   - Reduce the number of permissions for operator role.
+  - Support SHA-1 as an authentication method.
```
chatton, line 8: Thanks for keeping the release notes updated!

controllers/validation/validation.go
```diff
@@ -76,3 +85,27 @@ func validateArbiterSpec(mdb mdbv1.MongoDBCommunity) error {
...
 }
+
+// validateAuthModeSpec checks that the list of modes does not contain duplicates.
+func validateAuthModeSpec(mdb mdbv1.MongoDBCommunity) error {
+	allModes := mdb.Spec.Security.Authentication.Modes
+
+	// if no mode is specified in the specs, it's valid
+	if len(allModes) == 0 {
+		return nil
+	}
```
chatton, line 96: I believe modes is a required CRD field, so we actually will require at least one entry!

api/v1/mongodbcommunity_types.go
```diff
@@ -402,12 +423,39 @@ func (m MongoDBCommunity) GetScramOptions() scram.Options {
...
+				}
+			}
+		}
+		if containsDefaultMode {
+			AutoAuthMechanism = defaultSystemName
+		} else {
+			AutoAuthMechanism, _ = ConvertAuthModeLabelToAuthModeSystemName(listModes[0])
+		}
+	}
+
```
chatton, line 452: I have some concerts with this block. The variable names seem quite confusing, it's not immediately clear what everything is. Remember, we have AuthenticationModes (specified in CR) but also AuthenticationMechanisms (specified in automation config). It might be a good idea to distinguish between these two domains.

We appear to be a bit inconsistent with terminology used in the variable names (system/authMode/authMechanism) we should make sure that it is very clear wha is going on.

For example I was a bit confused by what `defaultSystemName`.  

So far in existing code we use `authentication.modes` to refer to the CR, and `authenticationMechanism` when dealing with the agent/automation config.

controllers/validation/validation.go
chatton: Maybe you can double check this by trying to create a resource manually without specifying an entry for `authentication.modes`

api/v1/mongodbcommunity_types.go
chatton: I would also say that this block of code is not that easy to follow as currently written. I think ideally we want to keep our go code as flat as possible. In this case that would involve extracting out some well named helper functions to aid in readability. 

I left another comment below about actually needing a single entry in the modes array, with this being the case, we might be able to simplify this logic a bit! (please let me know if my understanding is correct and that we need at least one entry)

chatton (commented):
This is great work! Thank you for writing both e2e and unit tests.

I left plenty questions/comments, please let me know if you would like more clarification about any of them!

controllers/validation/validation.go
```diff
@@ -19,6 +20,10 @@ func ValidateInitalSpec(mdb mdbv1.MongoDBCommunity) error {
 		return err
 	}
 
+	if err := validateAuthModeSpec(mdb); err != nil {
```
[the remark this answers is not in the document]
killian2k: Yes it makes sense to write a function especially for these verifications

api/v1/mongodbcommunity_types.go
```diff
@@ -40,6 +40,12 @@ const (
 	defaultPasswordKey = "password"
 )
 
+// SCRAM-SHA-256 and SCRAM-SHA-1 are the supported auth modes.
+const (
+	NumberAuthModeSupported          = 2
```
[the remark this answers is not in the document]
killian2k: Yes you're right. I initially used it in the validation, but it's not necessary anymore.

api/v1/mongodbcommunity_types.go
```diff
@@ -346,9 +352,24 @@ type Authentication struct {
...
+func ConvertAuthModeLabelToAuthModeSystemName(authModeLabel AuthMode) (string, bool) {
+	switch authModeLabel {
+	case "SCRAM":
+		return scram.Sha256, true
+	case "SCRAM-SHA-256":
+		return scram.Sha256, true
+	case "SCRAM-SHA-1":
+		return scram.Sha1, true
+	default:
+		return "", false
```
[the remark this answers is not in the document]
killian2k: I initially used a boolean to imitate the output of map objects, but it's simpler, like you said

test/e2e/replica_set_authentication/replica_set_authentication_test.go
```diff
@@ -0,0 +1,128 @@
+package replica_set_tls
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	v1 "github.com/mongodb/mongodb-kubernetes-operator/api/v1"
```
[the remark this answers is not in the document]
killian2k: Sure, should I rename it also in the two existing test files that name it v1?

test/e2e/replica_set_authentication/replica_set_authentication_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+func TestReplicaSetAuthentication(t *testing.T) {
+	ctx := setup.Setup(t)
+	defer ctx.Teardown()
+
+	mdb, user := e2eutil.NewTestMongoDB(ctx, "mdb0", "")
+	pw, err := setup.GeneratePasswordForUser(ctx, user, "")
+	if err != nil {
+		t.Logf("ERROR password generation")
+		t.Fatal(err)
```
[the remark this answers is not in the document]
killian2k: Yep my mistake. I added this for debug and forgot to remove it

test/e2e/util/mongotester/mongotester.go
```diff
@@ -398,6 +409,19 @@ func WithScram(username, password string) OptionApplier {
 	}
 }
 
+func WithScramWithAuth(username, password string, authenticationMechanism string) OptionApplier {
```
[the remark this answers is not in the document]
killian2k: Yes we do but should we use it as the alias is in the mongodbcommunity_types and this function is in mongotester? We discussed trying to separate the tests and the operator. Is it different concerning aliases?

test/e2e/replica_set_authentication/replica_set_authentication_test.go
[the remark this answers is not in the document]
chatton: @killian2k I think that would be best! We should make sure we have a consistent style across the whole codebase.

api/v1/mongodbcommunity_types.go
killian2k: Thanks for the reminder about the difference in the names with the CR and automation config. I simplified a bit the code. Let me know if you think it is still difficult to read.

killian2k:
closes #217

chatton (approved):
LGTM! This is great work @killian2k

durera:
Is there any information about the timeline for the 0.6.3 release that will carry this support?

Issue #464, opened 2021-05-05
unable to create replicaset from sample


umrikar:
Hi,
   I followed steps on https://github.com/mongodb/mongodb-kubernetes-operator/blob/master/docs/deploy-configure.md#deploy-a-replica-set

but pods remained in pending state, as volume claims were not available , is there any additional step required to perform before this?

Events:
  Type     Reason            Age                    From               Message
  ----     ------            ----                   ----               -------
  Warning  FailedScheduling  2m39s (x800 over 19h)  default-scheduler  0/3 nodes are available: 3 pod has unbound immediate PersistentVolumeClaims.

Image:      registry.hub.docker.com/library/mongo:4.2.6

jwenjian:
Same  issue, attach below output of `kubectl describe pvc logs-volume-example-mongodb-0`

```
Name:          logs-volume-example-mongodb-0
Namespace:     default
StorageClass:
Status:        Pending
Volume:
Labels:        app=example-mongodb-svc
Annotations:   <none>
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:
Access Modes:
VolumeMode:    Filesystem
Used By:       example-mongodb-0
Events:
  Type    Reason         Age                     From                         Message
  ----    ------         ----                    ----                         -------
  Normal  FailedBinding  3m16s (x962 over 4h3m)  persistentvolume-controller  no persistent volumes available for this claim and no storage class is set

```

chatton:
Hi @umrikar @jwenjian 

by default the operator will use whatever the current default storageclass is. This can be different based on your provider. You are able to specify a storageclass of your choosing.

You can refer to #157 with examples for this.

Please let me know if this helps you fix the problem!

umrikar:
@chatton , can i specify my own persistent volume and claim in statefulset, i did try it but its taking data-volume and logs-volume only,

apiVersion: mongodb.com/v1
kind: MongoDBCommunity
metadata:
  name: example-mongodb
spec:
  members: 3
  type: ReplicaSet
  version: "4.0.6

can i specify persistent volumes and claims inside above manifest, i did try but it didnt change anything it was still using default volume and claims,is there any other way to use my own persistent volumes and claims

umrikar:
> 
> 
> Hi @umrikar @jwenjian
> 
> by default the operator will use whatever the current default storageclass is. This can be different based on your provider. You are able to specify a storageclass of your choosing.
> 
[...]

PR #603, opened 2021-06-29
CLOUDP-89483: tests for custom volumes

bznein:
# Context
#464 

The user was trying to create a deployment that uses custom PVs/PVCs by defining volumes of type `hostpath`. The approach was not working for some different reasons that I will summarize here (and then I will reply to the issue when this get merged, with more details).

```
    spec:
          volumes:
              - name: logs-volume
                persistentVolumeClaim:
                    claimName: storage-volume-claim
              - name: data-volume
                persistentVolumeClaim:
                    claimName: storage-volume-claim
```

Specifying the PVC name here makes all pods trying to use the same PVCs (doesn't automatically add `-X` so this can't be done.

```
kind: PersistentVolume
apiVersion: v1
metadata:
  name: storage-volume
  labels:
    type: local
spec:
  storageClassName: manual
  capacity:
    storage: 10Gi
  accessModes:
    - ReadWriteOnce
  hostPath:
    path: "/opt/data/lxa/mongo/"
```

`hostPath` volumes are only writable by `root`, and our containers all run with non-root permissions. It is possible in theory to add an `initContainer` through the sts override that changes the ownership of the volume to user 2000, but that's a non-necessary workaround.

The user was trying to use a single PV/PVC for both containers within a Pod and all Pods within a Replica Set.
While this can be obtained by using `hostPath` and some hacks, it won't work (apart of not making much sense) as the second `mongod` will crash as soon as it tries to write on the lock file in `/data/`, since there is another `mongod` running there.

# Solution
There is not actually a need to use custom PVCs, but only custom PV. If an user has a Persistent Volume (one for each container in each pod), it can use the statefulset override to change the spec of the data-volume and logs-volume claim template we create to be whatever they want. If they want to make sure this PVC bind to their own volume, they can use match labels.

More comments inline


.action_templates/steps/setup-kind-cluster.yaml
```diff
@@ -3,6 +3,9 @@
     curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.11.1/kind-linux-amd64
     chmod +x ./kind
     ./kind create cluster
+- name: Create Directories
+  run: |
+    docker exec kind-control-plane  mkdir -p /opt/data/mongo-data-0 /opt/data/mongo-data-1 /opt/data/mongo-data-2 /opt/data/mongo-logs-0 /opt/data/mongo-logs-1 /opt/data/mongo-logs-2
```
bznein, line 8: The only way we can test this on Kind is to exec into the container and create six volumes (3 pods with two volumes each for logs and data).

Note: This will be done for each test, even the one that do not need it, but I don't think it is worth separating the logic

test/e2e/replica_set_custom_persistent_volume/replica_set_custom_persistent_volume.go
```diff
@@ -0,0 +1,141 @@
...
+									Operator: "In",
+									Values:   []string{"kind-control-plane"},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
```
bznein, line 61: This method returns a specific PersistentVolume that uses localPath on kind

test/e2e/replica_set_custom_persistent_volume/replica_set_custom_persistent_volume.go
```diff
@@ -0,0 +1,141 @@
...
+	volumes := make([]corev1.PersistentVolume, members)
+	for i := 0; i < members; i++ {
+		volumes[i] = getPersistentVolumeLocal(
+			fmt.Sprintf("%s-volume-%d", volumeType, i),
+			fmt.Sprintf("/opt/data/mongo-%s-%d", volumeType, i),
+			volumeType,
+		)
+	}
+	return volumes
+}
```
bznein, line 74: Returns a slice of 2 volumes for each member of the replicaset, one for the logs and one for data

test/e2e/replica_set_custom_persistent_volume/replica_set_custom_persistent_volume.go
```diff
@@ -0,0 +1,141 @@
...
+
+func getPvc(pvcType string, mdb v1.MongoDBCommunity) corev1.PersistentVolumeClaim {
```
bznein, line 76: This function returns a custom PVC, either for `data` or for `logs`

test/e2e/replica_set_custom_persistent_volume/replica_set_custom_persistent_volume.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	defaultString := "default"
+	return corev1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Spec: corev1.PersistentVolumeClaimSpec{
+			AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{"type": pvcType},
```
bznein, line 91: use `MatchLabels` as a way to ensure binding to the volume that has been created separately

test/e2e/replica_set_custom_persistent_volume/replica_set_custom_persistent_volume.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	os.Exit(code)
+}
+
+func getPersistentVolumeLocal(name string, localPath string, label string) corev1.PersistentVolume {
+	return corev1.PersistentVolume{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: e2eutil.OperatorNamespace,
+			Labels:    map[string]string{"type": label},
```
bznein, line 35: Use `labels` as a way to then use `matchlabels` in the PVC to bind to this specific volume

requirements.txt
```diff
@@ -13,4 +13,5 @@ dnspython==2.0.0
 requests==2.24.0
 pyyaml==5.4.1
 ruamel.yaml==0.17.9
+ghat==0.1.3
```
chatton, line 16: [q] what is this for?

test/e2e/replica_set_custom_persistent_volume/replica_set_custom_persistent_volume.go
chatton: ^^ this could be a useful docstring

test/e2e/replica_set_custom_persistent_volume/replica_set_custom_persistent_volume.go
chatton: ^^ also good docstring!

test/e2e/replica_set_custom_persistent_volume/replica_set_custom_persistent_volume.go
chatton: can we make `pvcType` an enum?

test/e2e/replica_set_custom_persistent_volume/replica_set_custom_persistent_volume.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func getPvc(pvcType string, mdb v1.MongoDBCommunity) corev1.PersistentVolumeClaim {
+	name := ""
+	if pvcType == "logs" {
+		name = mdb.LogsVolumeName()
+	} else {
+		name = mdb.DataVolumeName()
+	}
+	defaultString := "default"
```
chatton, line 83: this isn't a very useful variable name, maybe `defaultStorageClassName`?

test/e2e/replica_set_custom_persistent_volume/replica_set_custom_persistent_volume.go
```diff
@@ -0,0 +1,141 @@
...
+		getPvc("logs", mdb),
+	}
+	volumesToCreate := getVolumes(ctx, "data", mdb.Spec.Members)
+	volumesToCreate = append(volumesToCreate, getVolumes(ctx, "logs", mdb.Spec.Members)...)
+
+	for i := range volumesToCreate {
+		err := e2eutil.TestClient.Create(context.TODO(), &volumesToCreate[i], &e2eutil.CleanupOptions{TestContext: ctx})
+		assert.NoError(t, err)
+	}
+	scramUser := mdb.GetScramUsers()[0]
```
chatton, line 117: [q] isn't `user` which is already in scope the same as `scramUser`?

chatton (commented):
Looks good, just had a few questions!

requirements.txt
bznein: Good question! I think I got an error that something was missing and then added the wrong thing to requirements since this doesn't seem to have anything to do with us! WIll double chek

bznein: nvm found it

test/e2e/replica_set_custom_persistent_volume/replica_set_custom_persistent_volume.go
bznein: I copied the skeleton for this test from another one, but let me double check!

bznein: they have different types

`user` is `MongoDBUser`
`scramUser` is `scram.User`

I guess we could avoid duplicating and just converting one to the other, but this is basically what `GetScramUsers()` does, I think it's ok to leave it like this, what do you think?

chatton (approved):
LGTM 👍

PR #605, opened 2021-06-29
CLOUDP-90457 Create kubelinter-check.yml

killian2k:
Added the kubelinter action + reduce the CPU request when doing the e2e tests. Now, the manager.yaml and the openshift operator are in the mongodb namespace

### All Submissions:

* [x] Have you signed all of your commits?


.github/workflows/kubelinter-check.yml
```diff
@@ -0,0 +1,50 @@
+---
+#
+# Docs: https://github.com/marketplace/actions/close-stale-issues
```
bznein, line 3: non relevant comment

config/manager/manager.yaml
```diff
@@ -39,4 +58,15 @@ spec:
         image: quay.io/mongodb/mongodb-kubernetes-operator:0.6.2
         imagePullPolicy: Always
         name: mongodb-kubernetes-operator
+        resources:
+          limits:
+            cpu: 1100m
+            memory: 1Gi
+          requests:
+            cpu: 500m
+            memory: 200Mi
```
bznein, line 67: [q] How did we derive these values?

deploy/clusterwide/role.yaml
```diff
@@ -6,10 +6,10 @@ rules:
 - apiGroups:
```
bznein, line 6: The diff on this file is a bit noisy, not your fault. Can you summarize what changed? I see we completely removed permissions on secrets but is there a specific reason we keep secrets in with `verbs: []`?

deploy/e2e/role.yaml
```diff
@@ -104,3 +104,4 @@ rules:
   - patch
   - update
   - watch
+
```
bznein, line 107: [micro-nit] Can we make github happy here by just restoring this file as it was?

test/e2e/setup/setup.go
```diff
@@ -238,6 +240,22 @@ func createOrUpdate(obj client.Object) error {
 	return nil
 }
 
+// withCPURequest assumes that the underlying type is an appsv1.Deployment.
```
bznein, line 243: Can we maybe log something in case this is not a deployment? Or maybe even fail loudly so that is clear that this function has been called on the wrong type?

bznein (commented):
Overall looks good and I look forward to seeing this in action, just a few comments!

config/manager/manager.yaml
killian2k: According to Cian these are the values used for the enterprise operator

bznein: makes sense, I guess at some point we'd want to perform some actual testing of the community operator to come with a precise number, but definitely out of scope here

deploy/clusterwide/role.yaml
killian2k: Removed  create for pods, Removed everything from secret

test/e2e/setup/setup.go
killian2k: Good idea, I added the error management to the options functions

.github/config_files/config_lint.yaml
```diff
@@ -0,0 +1,13 @@
+checks:
+  addAllBuiltIn: true
+
+#Reasons to exclude:
+  # non-existent-service-account because the service account is created in another file
+  # minimum-three-replicas because the deployment contains only 1 replica of the operator
+  # no-readiness-probe & no-liveness-probe because for now, it brings nothing to add theses probes
```
chatton, line 7:
```suggestion
  # no-readiness-probe & no-liveness-probe because for now, it brings nothing to add these probes
```

.github/workflows/kubelinter-check.yml
```diff
@@ -0,0 +1,46 @@
+name: Kubelinter-check
+
+on:
+  push:
+    branches:
+    - master
+    paths-ignore: 
+    - docs/**
```
chatton, line 8: I think maybe we don't need need to exclude any directories, it's possible that we could have yaml samples under `docs`

deploy/e2e/service_account.yaml
```diff
@@ -2,3 +2,4 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: e2e-test
+  namespace: default
```
chatton, line 5: can we add a note here saying why we are still using the default ns for tests (TLS fixtures assume default ns)

chatton (commented):
This looks great! I just had a few small comments but I think we're good to merge afterwards.

bznein (commented):
LGTM for what concerns my comments :) Leaving to Cian final LGTM since I see he has some comments

chatton (approved):
LGTM! Nice work on this!

PR #607, opened 2021-06-30
CLOUDP-91632: E2E tests should fail fast if the operator pod doesn't reach READY state

FlorentinaSimlinger:
### All Submissions:

Tested it by changing the parameters of the field "operatorImage" in the testConfig on line 28 in test/e2e/setup/setup.go to "blah".

### DIFF
--- a/test/e2e/setup/setup.go
+++ b/test/e2e/setup/setup.go
@@ -170,7 +170,7 @@ func deployOperator() error {
 		return errors.Errorf("error building operator deployment: %s", err)
 	}
 
-	if err := wait.PollImmediate(time.Second, time.Duration(30)*time.Second, hasDeploymentRequiredReplicas(dep)); err != nil {
+	if err := wait.PollImmediate(time.Second, 30*time.Second, hasDeploymentRequiredReplicas(dep)); err != nil {
 		return errors.New("error building operator deployment: the deployment does not have the required replicas")
 	}
 
--- a/test/e2e/setup/setup.go
+++ b/test/e2e/setup/setup.go
@@ -187,6 +187,9 @@ func hasDeploymentRequiredReplicas(dep *appsv1.Deployment) wait.ConditionFunc {
 				Namespace: e2eutil.OperatorNamespace},
 			dep)
 		if err != nil {
+			if apiErrors.IsNotFound(err) {
+				return false, nil
+			}
 			return false, errors.Errorf("error getting operator deployment: %s", err)
 		}
 		if dep.Status.ReadyReplicas == *dep.Spec.Replicas {
--- a/test/e2e/setup/setup.go
+++ b/test/e2e/setup/setup.go
@@ -156,8 +156,10 @@ func deployOperator() error {
 	}
 	fmt.Println("Successfully created the operator Role Binding")
 
+	dep := &appsv1.Deployment{}
+
 	if err := buildKubernetesResourceFromYamlFile(path.Join(deployDir(), "manager.yaml"),
-		&appsv1.Deployment{},
+		dep,
 		withNamespace(testConfig.namespace),
 		withOperatorImage(testConfig.operatorImage),
 		withEnvVar("WATCH_NAMESPACE", watchNamespace),
@@ -168,13 +170,8 @@ func deployOperator() error {
 		return errors.Errorf("error building operator deployment: %s", err)
 	}
 
-	// TODO: check if we could get Name from service account
-
-	// Note: PollImmediate tries a condition func until it returns true, an error, or the timeout is reached.
-	// PollImmediate always checks 'condition' before waiting for the interval. 'condition' will always be invoked at least once.
-	// Some intervals may be missed if the condition takes too long or the time window is too short.
-	if err := wait.PollImmediate(time.Second, time.Duration(30)*time.Second, hasDeploymentRequiredReplicas(&appsv1.Deployment{})); err != nil {
-		return errors.New("error building operator deployment: the deployment does not have the required replicas (possibly because the image is not correct)")
+	if err := wait.PollImmediate(time.Second, time.Duration(30)*time.Second, hasDeploymentRequiredReplicas(dep)); err != nil {
+		return errors.New("error building operator deployment: the deployment does not have the required replicas")
 	}
 
 	fmt.Println("Successfully created the operator Deployment")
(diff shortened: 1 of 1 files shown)


test/e2e/setup/setup.go
```diff
@@ -164,6 +166,45 @@ func deployOperator() error {
...
+	// TODO: check if we could get Name from service account
+	// TODO: check if actualReplicas is getting updated inside the loop
+	dep := appsv1.Deployment{}
+
+	// retrieve required replicas -- they will already be there
+	if err := e2eutil.TestClient.Get(context.TODO(),
+		types.NamespacedName{Name: "mongodb-kubernetes-operator",
+			Namespace: e2eutil.OperatorNamespace},
+		&dep); err != nil {
+		return errors.Errorf("error building operator deployment: %s", err)
```
irajdeep, line 183: we should return `failed to get` error here

test/e2e/setup/setup.go
```diff
@@ -164,6 +166,45 @@ func deployOperator() error {
...
+			break
+		}
+	}
+
+	// check if now the condition is fulfilled
+	actualReplicas := dep.Status.ReadyReplicas
+	if actualReplicas != requiredReplicas {
+		return errors.New("error building operator deployment: actual number of replicas does not match required number of replicas")
+	}
+
```
irajdeep, line 207: I was just looking at if we can use some existing function from k8s and I found this: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait . I think we can use this and pass the function of getting the deployment and checking it's status. I think it will make the code more robust and easy to follow.

irajdeep (commented):
LGTM overall -- added some comments inline.

test/e2e/setup/setup.go
```diff
@@ -164,10 +167,50 @@ func deployOperator() error {
...
+func hasDeploymentRequiredReplicas(dep *appsv1.Deployment) wait.ConditionFunc {
+	// Note: ConditionFunc returns true if the condition is satisfied, or an error if the loop should be aborted.
+	return func() (bool, error) {
+		if err := getDeployment(dep); err != nil {
+			fmt.Printf("returning error\n")
+			return false, errors.Errorf("error getting operator deployment: %s\n", err)
+		}
+		fmt.Printf("actual replicas %d\n", dep.Status.ReadyReplicas)
+		fmt.Printf("required replicas %d\n", *dep.Spec.Replicas)
+		switch dep.Status.ReadyReplicas {
```
irajdeep, line 195: we can just do:
``` err := e2eutil.TestClient.Get(context.TODO(),
		types.NamespacedName{Name: "mongodb-kubernetes-operator",
			Namespace: e2eutil.OperatorNamespace},
		dep)
		if err != nil {}```

File: test/e2e/setup/setup.go [hunk 4]
```diff
@@ -164,10 +167,50 @@ func deployOperator() error {
...
+			return true, nil
+		default:
+			fmt.Printf("returning false\n")
+			return false, nil
+		}
+	}
+}
+
+// getDeployment fills the empty deployment object with fields from the resource and returns an error if it cannot get it
+func getDeployment(dep *appsv1.Deployment) error {
```
irajdeep, line 207: We can avoid this method altogether and make the call in the function.

irajdeep (commented):
Added some more comments.

test/e2e/setup/setup.go
```diff
@@ -164,10 +169,33 @@ func deployOperator() error {
...
+	if err := wait.PollImmediate(time.Second, time.Duration(30)*time.Second, hasDeploymentRequiredReplicas(dep)); err != nil {
+		return errors.New("error building operator deployment: the deployment does not have the required replicas")
+	}
+
 	fmt.Println("Successfully created the operator Deployment")
 	return nil
 }
 
+// hasDeploymentRequiredReplicas returns a condition function that indicates whether the given deployment
+// currently has the required amount of replicas
```
irajdeep, line 182:
```suggestion
// currently has the required amount of replicas in the ready state as specified in spec.replicas
```

test/e2e/setup/setup.go
```diff
@@ -164,10 +169,33 @@ func deployOperator() error {
 	); err != nil {
 		return errors.Errorf("error building operator deployment: %s", err)
 	}
+
+	if err := wait.PollImmediate(time.Second, time.Duration(30)*time.Second, hasDeploymentRequiredReplicas(dep)); err != nil {
```
tibulca, line 173: in this repo, I see in most of the cases we're using this format: `30 * time.Second` instead of `time.Duration(30)*time.Second`

test/e2e/setup/setup.go
```diff
@@ -164,10 +169,33 @@ func deployOperator() error {
...
+			types.NamespacedName{Name: dep.Name,
+				Namespace: e2eutil.OperatorNamespace},
+			dep)
+		if err != nil {
+			return false, errors.Errorf("error getting operator deployment: %s", err)
+		}
+		if dep.Status.ReadyReplicas == *dep.Spec.Replicas {
+			return true, nil
+		}
+		return false, nil
```
tibulca, line 195: nit: what do you think about replacing this with something like this
```
depNamespacedName := types.NamespacedName{Name: dep.Name, Namespace: e2eutil.OperatorNamespace}
if err := e2eutil.TestClient.Get(context.TODO(), depNamespacedName, dep); err != nil {
    return false, errors.Errorf("error getting operator deployment: %s", err)
}
replicasReachedTheGoalState := dep.Status.ReadyReplicas == *dep.Spec.Replicas
return replicasReachedTheGoalState, nil
```
it's really minor but it improves a bit the readability

test/e2e/setup/setup.go
FlorentinaSimlinger: fixed! :)

test/e2e/setup/setup.go
irajdeep: > replicasReachedTheGoalState := dep.Status.ReadyReplicas == *dep.Spec.Replicas

I think it's better to be explicit here about the if check, also to align with the existing pattern in the codebase -- I am inclined towards having multiple explicit returns instead of assigning it to a temporary variable.

irajdeep (approved):

test/e2e/setup/setup.go
```diff
@@ -164,10 +169,33 @@ func deployOperator() error {
...
+// currently has the required amount of replicas in the ready state as specified in spec.replicas
+func hasDeploymentRequiredReplicas(dep *appsv1.Deployment) wait.ConditionFunc {
+	return func() (bool, error) {
+		err := e2eutil.TestClient.Get(context.TODO(),
+			types.NamespacedName{Name: dep.Name,
+				Namespace: e2eutil.OperatorNamespace},
+			dep)
+		if err != nil {
+			return false, errors.Errorf("error getting operator deployment: %s", err)
+		}
```
chatton, line 191: I have one concern here, since we call the wait function which performs a Get operation immediately after we create the Deployment, it could be possible that we receive a not found error here. When an error is returned in `wait.Poll` it immediately exits with the error. We could maybe try to simply return `false` in the case of an `NotFound` error. @irajdeep @FlorentinaSimlinger what do you think?

test/e2e/setup/setup.go
```diff
@@ -164,10 +169,33 @@ func deployOperator() error {
 	); err != nil {
 		return errors.Errorf("error building operator deployment: %s", err)
 	}
+
+	if err := wait.PollImmediate(time.Second, 30*time.Second, hasDeploymentRequiredReplicas(dep)); err != nil {
+		return errors.New("error building operator deployment: the deployment does not have the required replicas")
```
chatton, line 174: we don't expose the underlying reason (i.e. the contents of `err` ). In go it is idiomatic to wrap errors with additional context. Something like

```golang
return errors.Errorf("error building operator deployment: the deployment does not have the required replicas: %s", err)
```

test/e2e/setup/setup.go
irajdeep: Sure, to be extra cautious we can check for `NotFound` and return nil here.

test/e2e/setup/setup.go
tibulca: I suggested a temp variable to express the intention, but agree consistency is more important

test/e2e/setup/setup.go
irajdeep: I don't think we need to wrap the error here, the code makes no decision based on the error content -- pretty much when this error happens, the user would go and describe the object.

chatton: I think that's a good point about the lack of decision! This message as is should be enough for what we need. 👍

PR #609, opened 2021-06-30
Use "run" script instead of GH Action to install kind

chatton:
Kind GithubAction was causing failures during setup of cluster.

Setting up the cluster manually seems to fix this issue.


bznein (approved):
LGTM once the tests are green, thanks for the fix!

PR #616, opened 2021-07-05
CLOUDP-93590: Service name configurable

FlorentinaSimlinger:
### All Submissions:

Tested by adding 
statefulSet:
    spec:
      serviceName: database
to the custom resource yaml file and running make run to observe that the serviceName field now contains database and by setting the service name in statefulset_arbitrary_config_test.go and checking the name is correctly set in mongodbtests.go

### DIFF
--- a/api/v1/mongodbcommunity_types.go
+++ b/api/v1/mongodbcommunity_types.go
@@ -528,11 +528,10 @@ func (m MongoDBCommunity) Hosts() []string {
 // this resource which is either the default name resourcename-svc or the user defined name
 func (m MongoDBCommunity) ServiceName() string {
 	serviceName := m.Spec.StatefulSetConfiguration.SpecWrapper.Spec.ServiceName
-	if serviceName == "" {
-		return m.Name + "-svc"
-	} else {
+	if serviceName != "" {
 		return serviceName
 	}
+	return m.Name + "-svc"
 }
 
 func (m MongoDBCommunity) AutomationConfigSecretName() string {
--- a/api/v1/mongodbcommunity_types.go
+++ b/api/v1/mongodbcommunity_types.go
@@ -524,8 +524,7 @@ func (m MongoDBCommunity) Hosts() []string {
 	return hosts
 }
 
-// ServiceName returns the name of the Service that should be created for
-// this resource which is either the default name resourcename-svc or the user defined name
+// ServiceName returns the name of the Service that should be created for this resource
 func (m MongoDBCommunity) ServiceName() string {
 	serviceName := m.Spec.StatefulSetConfiguration.SpecWrapper.Spec.ServiceName
 	if serviceName != "" {
--- a/api/v1/mongodbcommunity_types.go
+++ b/api/v1/mongodbcommunity_types.go
@@ -525,9 +525,14 @@ func (m MongoDBCommunity) Hosts() []string {
 }
 
 // ServiceName returns the name of the Service that should be created for
-// this resource
+// this resource which is either the default name resourcename-svc or the user defined name
 func (m MongoDBCommunity) ServiceName() string {
-	return m.Name + "-svc"
+	serviceName := m.Spec.StatefulSetConfiguration.SpecWrapper.Spec.ServiceName
+	if serviceName == "" {
+		return m.Name + "-svc"
+	} else {
+		return serviceName
+	}
 }
 
 func (m MongoDBCommunity) AutomationConfigSecretName() string {
--- a/config/samples/mongodb.com_v1_mongodbcommunity_cr.yaml
+++ b/config/samples/mongodb.com_v1_mongodbcommunity_cr.yaml
@@ -23,6 +23,9 @@ spec:
       scramCredentialsSecretName: my-scram
   additionalMongodConfig:
     storage.wiredTiger.engineConfig.journalCompressor: zlib
+  statefulSet:
+    spec:
+      serviceName: newdatabasename
 
 # the user credentials will be generated from this secret
 # once the credentials are generated, this secret is no longer required
--- a/config/samples/mongodb.com_v1_mongodbcommunity_cr.yaml
+++ b/config/samples/mongodb.com_v1_mongodbcommunity_cr.yaml
@@ -23,9 +23,6 @@ spec:
       scramCredentialsSecretName: my-scram
   additionalMongodConfig:
     storage.wiredTiger.engineConfig.journalCompressor: zlib
-  statefulSet:
-    spec:
-      serviceName: newdatabasename
 
 # the user credentials will be generated from this secret
 # once the credentials are generated, this secret is no longer required
--- a/test/e2e/mongodbtests/mongodbtests.go
+++ b/test/e2e/mongodbtests/mongodbtests.go
@@ -289,6 +289,13 @@ func BasicFunctionality(mdb *mdbv1.MongoDBCommunity) func(*testing.T) {
 	}
 }
 
+// ServiceHasCorrectName asserts whether the actual service name equals the expected service name
+func ServiceHasCorrectName(mdb *mdbv1.MongoDBCommunity, expectedName string) func(t *testing.T) {
+	return func(t *testing.T) {
+		assert.Equal(t, expectedName, mdb.ServiceName())
+	}
+}
+
 // DeletePod will delete a pod that belongs to this MongoDB resource's StatefulSet
 func DeletePod(mdb *mdbv1.MongoDBCommunity, podNum int) func(*testing.T) {
 	return func(t *testing.T) {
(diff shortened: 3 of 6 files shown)


config/samples/mongodb.com_v1_mongodbcommunity_cr.yaml
```diff
@@ -23,6 +23,9 @@ spec:
       scramCredentialsSecretName: my-scram
   additionalMongodConfig:
     storage.wiredTiger.engineConfig.journalCompressor: zlib
+  statefulSet:
+    spec:
+      serviceName: newdatabasename
```
irajdeep, line 28: I don't think we need to push this change, can you please revert it?

controllers/replica_set_controller.go
```diff
@@ -129,6 +129,7 @@ func (r ReplicaSetReconciler) Reconcile(ctx context.Context, request reconcile.R
 	}
 
 	r.log = zap.S().With("ReplicaSet", request.NamespacedName)
+
```
irajdeep, line 132: revert this new line?

api/v1/mongodbcommunity_types.go
```diff
@@ -523,9 +523,14 @@ func (m MongoDBCommunity) Hosts() []string {
...
-// this resource
+// this resource which is either the default name resourcename-svc or the user defined name
 func (m MongoDBCommunity) ServiceName() string {
-	return m.Name + "-svc"
+	serviceName := m.Spec.StatefulSetConfiguration.SpecWrapper.Spec.ServiceName
+	if serviceName == "" {
+		return m.Name + "-svc"
+	} else {
+		return serviceName
+	}
```
irajdeep, line 533: I think it would be simpler to write
```
if serviceName  !=  "" {
		return serviceName
}
return m.Name + "-svc"
```
and get rid of the extra else check.

test/e2e/statefulset_arbitrary_config/statefulset_arbitrary_config_test.go
```diff
@@ -22,9 +22,11 @@ func TestMain(m *testing.M) {
 }
 
 func TestStatefulSetArbitraryConfig(t *testing.T) {
+	// creates operator/deployment from the yaml files
 	ctx := setup.Setup(t)
 	defer ctx.Teardown()
 
+	// creates basic mongodb resource/custom resource
```
irajdeep, line 29: I think it's fairly clear here what the method `NewTestMongoDB` does -- so I think the comment is redundant. we can remove it.

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -289,6 +289,13 @@ func BasicFunctionality(mdb *mdbv1.MongoDBCommunity) func(*testing.T) {
 	}
 }
 
+// ServiceHasCorrectName asserts whether the actual service name equals the expected service name
+func ServiceHasCorrectName(mdb *mdbv1.MongoDBCommunity, expectedName string) func(t *testing.T) {
+	return func(t *testing.T) {
+		assert.Equal(t, expectedName, mdb.ServiceName())
+	}
+}
+
```
irajdeep, line 298: We need the get the service here and check the name

irajdeep (commented):
The functionality looks good -- added some comments.

config/samples/mongodb.com_v1_mongodbcommunity_cr.yaml
FlorentinaSimlinger: Oops sorry, I thought I had made sure not to push it :)

test/e2e/statefulset_arbitrary_config/statefulset_arbitrary_config_test.go
FlorentinaSimlinger: leftovers of my own debugging process :)

api/v1/mongodbcommunity_types.go
```diff
@@ -523,8 +523,12 @@ func (m MongoDBCommunity) Hosts() []string {
 }
 
 // ServiceName returns the name of the Service that should be created for
-// this resource
+// this resource which is either the default name resourcename-svc or the user defined name
```
chatton, line 526: [nit] I don't think implementation details necessarily need to be included in the docstring. The new text added is re-stating the body of the function. What do you think?

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -289,6 +289,19 @@ func BasicFunctionality(mdb *mdbv1.MongoDBCommunity) func(*testing.T) {
 	}
 }
 
+// ServiceWithNameExists checks whether a service with the name serviceName exists
+func ServiceWithNameExists(mdb *mdbv1.MongoDBCommunity, serviceName string) func(t *testing.T) {
```
chatton, line 293: [nit] we are introducing a dependency on  `*mdbv1.MongoDBCommunity` when in realsity all we actually care about is the namespace. This signature could be `(namespace, serviceName string)` or even `(nsName types.NamespacedName)`

test/e2e/statefulset_arbitrary_config/statefulset_arbitrary_config_test.go
```diff
@@ -34,13 +34,17 @@ func TestStatefulSetArbitraryConfig(t *testing.T) {
...
+	mdb.Spec.StatefulSetConfiguration.SpecWrapper.Spec.ServiceName = serviceName
+
 	tester, err := mongotester.FromResource(t, mdb)
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	t.Run("Create MongoDB Resource", mongodbtests.CreateMongoDBResource(&mdb, ctx))
 	t.Run("Basic tests", mongodbtests.BasicFunctionality(&mdb))
+	t.Run("Test setting Service Name", mongodbtests.ServiceWithNameExists(&mdb, serviceName))
```
chatton, line 47: we are now **only** testing the service creation when it is specified in the StatefulSetSpec override, we could add this to the `BasicTests` tests which are run in every test to make sure the service name is as expected when not overriding through sts override. WDYT?

chatton (commented):
Looks great! Had a few small comments/suggestions

bznein (commented):
The overall implementation LGTM, I have only one main question here: what happens if a user here creates two resources with the same name service name? 

I guess this is the user's responsibility so no need to implement weird logic to cover this case, but is there maybe a clean way to identify this situation?

Other two minor things:
- We should add it to the release notes doc
- We could maybe provide an example in the `samples` directory

irajdeep:
> what happens if a user here creates two resources with the same name service name?

How would that be possible? you mean create the service manually?

bznein:
> > what happens if a user here creates two resources with the same name service name?
> 
> How would that be possible? you mean create the service manually?

I mean creating two resources and use the same value for both in `.Spec.StatefulSetConfiguration.SpecWrapper.Spec.ServiceName`.

irajdeep:
> I mean creating two resources and use the same value for both in .Spec.StatefulSetConfiguration.SpecWrapper.Spec.ServiceName

In this case, the reconcile should just fail and report it to status -- which is what will happen currently.

bznein:
> > I mean creating two resources and use the same value for both in .Spec.StatefulSetConfiguration.SpecWrapper.Spec.ServiceName
> 
> In this case, the reconcile should just fail and report it to status -- which is what will happen currently.

SGTM!

test/e2e/mongodbtests/mongodbtests.go
irajdeep: This should happen more often in functions, I think `mongodbtests.go` has quite a few such examples -- e.g `StatefulSetHasOwnerReference` ...

api/v1/mongodbcommunity_types.go
FlorentinaSimlinger: agree

test/e2e/statefulset_arbitrary_config/statefulset_arbitrary_config_test.go
FlorentinaSimlinger: do you mean adding a call to ServiceWithNameExists to BasicFunctionality and call it with m.Name + "-svc"?

test/e2e/statefulset_arbitrary_config/statefulset_arbitrary_config_test.go
```diff
@@ -34,8 +34,8 @@ func TestStatefulSetArbitraryConfig(t *testing.T) {
 
 	mdb.Spec.StatefulSetConfiguration.SpecWrapper.Spec.Template.Spec.Containers[1].ReadinessProbe = &corev1.Probe{TimeoutSeconds: 100}
 
-	serviceName := "database"
-	mdb.Spec.StatefulSetConfiguration.SpecWrapper.Spec.ServiceName = serviceName
+	customServiceName := "database"
+	mdb.Spec.StatefulSetConfiguration.SpecWrapper.Spec.ServiceName = customServiceName
```
FlorentinaSimlinger, line 38: I changed the parameter to customServiceName to make it clearer to the reader that this is a name that is set by the user. Happy to revert if original parameter name is preferred.

irajdeep (approved):
LGTM

bznein (approved):
LGTM!

PR #623, opened 2021-07-08
CLOUDP 90175: cleanup all resources

FlorentinaSimlinger:
### All Submissions:

Tested by running e2e replica_set test and checking after that the following resources are not existing:
e2e-test pod, serviceaccount, clusterrole, clusterrolebinding
as well as the
mongodb-kubernetes-operator role, service account, role binding and deployment.
I did not delete the namespace.

### DIFF
--- a/scripts/dev/e2e.py
+++ b/scripts/dev/e2e.py
@@ -11,6 +11,9 @@ import sys
 import yaml
 
 TEST_POD_NAME = "e2e-test"
+TEST_CLUSTER_ROLE_NAME= "e2e-test"
+TEST_CLUSTER_ROLE_BINDING_NAME="e2e-test"
+TEST_SERVICE_ACCOUNT_NAME="e2e-test"
 
 
 def load_yaml_from_file(path: str) -> Dict:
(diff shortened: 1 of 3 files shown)


scripts/dev/e2e.py
```diff
@@ -11,6 +11,9 @@
 import yaml
 
 TEST_POD_NAME = "e2e-test"
+TEST_CLUSTER_ROLE_NAME= "e2e-test"
+TEST_CLUSTER_ROLE_BINDING_NAME="e2e-test"
+TEST_SERVICE_ACCOUNT_NAME="e2e-test"
```
FlorentinaSimlinger, line 16: We could just make this one constant, something like TEST_NAME. I left it like this for clarity but open to change it.

scripts/dev/e2e.py
```diff
@@ -69,18 +71,6 @@ def _prepare_test_environment(config_file: str) -> None:
...
-    """
-    _delete_testrunner_pod deletes the test runner pod
-    if it already exists.
-    """
-    dev_config = load_config(config_file)
-    corev1 = client.CoreV1Api()
-    k8s_conditions.ignore_if_doesnt_exist(
-        lambda: corev1.delete_namespaced_pod(TEST_POD_NAME, dev_config.namespace)
-    )
-
```
FlorentinaSimlinger, line 82: I moved _delete_test_pod down so that it is below _delete_test_environment because create_test_pod is also below _create_test_environment

test/e2e/setup/setup.go
```diff
@@ -46,7 +46,7 @@ func Setup(t *testing.T) *e2eutil.Context {
 		t.Fatal(err)
 	}
 
-	if err := deployOperator(); err != nil {
+	if err := deployOperator(ctx); err != nil {
```
FlorentinaSimlinger, line 49: I'm passing ctx because before createOrUpdate created an empty context which is why the shouldPerformCleanup flag was not propagated.

scripts/dev/e2e.py
```diff
@@ -69,18 +71,6 @@ def _prepare_test_environment(config_file: str) -> None:
     )
```
FlorentinaSimlinger, line 71: I think for clarity the print statement should say "Creating Cluster Role Binding", similar for the role. Similarly the comments should reflect that and should also be corrected to say test pod instead of test runner. Happy to do this in a later PR (if wanted at all), just wanted to bring it up :)

PR #624, opened 2021-07-08
CLOUDP-90175: cleanup all resources

FlorentinaSimlinger:
### All Submissions:

Tested by running replica_set test and observing that after the following resources do not exist:
e2e-test: pod, clusterrole, clusterrolebinding, serviceaccount
mongodb-kubernetes-operator: deployment, role, role binding, service account
I did not delete the namespace.
Note that it will only delete those resources if they do not exist before (the delete functions are registered upon creation, not when existing resources are only updated). Thus, please make sure to delete any e2e-related resources before running a new e2e test.

### DIFF
--- a/scripts/dev/e2e.py
+++ b/scripts/dev/e2e.py
@@ -10,10 +10,12 @@ import argparse
 import sys
 import yaml
 
-TEST_POD_NAME = "e2e-test"
-TEST_CLUSTER_ROLE_NAME = "e2e-test"
-TEST_CLUSTER_ROLE_BINDING_NAME = "e2e-test"
-TEST_SERVICE_ACCOUNT_NAME = "e2e-test"
+(
+    TEST_POD_NAME,
+    TEST_CLUSTER_ROLE_NAME,
+    TEST_CLUSTER_ROLE_BINDING_NAME,
+    TEST_SERVICE_ACCOUNT_NAME,
+) = "e2e-test"
 
 
 def load_yaml_from_file(path: str) -> Dict:
@@ -35,7 +37,7 @@ def _load_test_role_binding() -> Dict:
 
 def _prepare_test_environment(config_file: str) -> None:
     """
-    _prepare_testrunner_environment ensures that the namespace, cluster role,
+    _prepare_test_environment ensures that the namespace, cluster role,
     cluster role binding and service account are created for the test pod.
     """
     rbacv1 = client.RbacAuthorizationV1Api()
@@ -148,7 +150,7 @@ def create_test_pod(args: argparse.Namespace, dev_config: DevConfig) -> None:
         timeout=60,
         exceptions_to_ignore=ApiException,
     ):
-        raise Exception("Could not create test_runner pod!")
+        raise Exception("Could not create test pod!")
 
 
 def wait_for_pod_to_be_running(
--- a/scripts/dev/e2e.py
+++ b/scripts/dev/e2e.py
@@ -11,9 +11,9 @@ import sys
 import yaml
 
 TEST_POD_NAME = "e2e-test"
-TEST_CLUSTER_ROLE_NAME= "e2e-test"
-TEST_CLUSTER_ROLE_BINDING_NAME="e2e-test"
-TEST_SERVICE_ACCOUNT_NAME="e2e-test"
+TEST_CLUSTER_ROLE_NAME = "e2e-test"
+TEST_CLUSTER_ROLE_BINDING_NAME = "e2e-test"
+TEST_SERVICE_ACCOUNT_NAME = "e2e-test"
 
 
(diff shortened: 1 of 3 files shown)


scripts/dev/e2e.py
```diff
@@ -11,6 +11,9 @@
 import yaml
 
 TEST_POD_NAME = "e2e-test"
+TEST_CLUSTER_ROLE_NAME = "e2e-test"
+TEST_CLUSTER_ROLE_BINDING_NAME = "e2e-test"
+TEST_SERVICE_ACCOUNT_NAME = "e2e-test"
```
FlorentinaSimlinger, line 17: We could summarise those in one constant, eg TEST_NAME -- I kept it as is for clarity, but happy to make changes

scripts/dev/e2e.py
```diff
@@ -69,18 +71,6 @@ def _prepare_test_environment(config_file: str) -> None:
...
-    """
-    _delete_testrunner_pod deletes the test runner pod
-    if it already exists.
-    """
-    dev_config = load_config(config_file)
-    corev1 = client.CoreV1Api()
-    k8s_conditions.ignore_if_doesnt_exist(
-        lambda: corev1.delete_namespaced_pod(TEST_POD_NAME, dev_config.namespace)
-    )
-
```
FlorentinaSimlinger, line 82: I moved this function to the bottom for file structure reasons: it is now below delete_test_environment just like create_test_pod is below create_test_environment

test/e2e/setup/setup.go
```diff
@@ -46,7 +46,7 @@ func Setup(t *testing.T) *e2eutil.Context {
 		t.Fatal(err)
 	}
 
-	if err := deployOperator(); err != nil {
+	if err := deployOperator(ctx); err != nil {
 		t.Fatal(err)
 	}
```
FlorentinaSimlinger, line 52: I had to pass ctx because before createOrUpdate was called with an empty context which meant that the shouldPerformCleanup flag was not used.

scripts/dev/e2e.py
```diff
@@ -69,18 +71,6 @@ def _prepare_test_environment(config_file: str) -> None:
     )
```
FlorentinaSimlinger, line 71: There are two things I would like to improve (if wanted at all) in either this or a separate PR:
- changing the print statements so that they accurately reflect that a cluster role binding and a cluster role instead of a role binding and role are created (makes it easier to spot it right away in future debugging proceses)
- fixing the comments of existing functions: some function names have changed and the comments do not reflect that. The comment for _prepare_test_environment could also be more specific.

irajdeep: feel free to do them in the same PR if they are relatively small changes.

bznein (approved):
LGTM!

test/e2e/setup/setup.go
```diff
@@ -228,8 +229,8 @@ func marshalRuntimeObjectFromYAMLBytes(bytes []byte, obj runtime.Object) error {
 	return json.Unmarshal(jsonBytes, &obj)
 }
 
-func createOrUpdate(obj client.Object) error {
-	if err := e2eutil.TestClient.Create(context.TODO(), obj, &e2eutil.CleanupOptions{}); err != nil {
+func createOrUpdate(obj client.Object, ctx *e2eutil.Context) error {
```
chatton, line 232: [nit] we should keep argument order consistent, ctx has been the first argument everywhere else.

chatton (commented):
Looks great! Just had a few small comments/questions

irajdeep (approved):
LGTM! just make `ctx` as the first parameter in the function as already pointed out.

scripts/dev/e2e.py
```diff
@@ -230,6 +255,14 @@ def prepare_and_run_test(args: argparse.Namespace, dev_config: DevConfig) -> Non
         print(line.decode("utf-8").rstrip())
 
 
+def delete_test_resources(config_file: str) -> None:
+    """
+    delete_test_resources ensures that the test environment and test pod are deleted.
+    """
+    _delete_test_environment(config_file)
+    _delete_test_pod(config_file)
```
chatton, line 263: [q] do we want to delete the pod? This means after a test run we will lose the pod and access to the logs, what do you think?

irajdeep: yeah -- we can skip deleting the test pod.

FlorentinaSimlinger: Good point. I'll revert that change. Should we maybe do a separate ticket to look into options to create a flag to keep the pod alive after the test ran?

irajdeep: sure we can do it as a follow-up later.

chatton (approved):
LGTM!

PR #628, opened 2021-07-09
Fix Github Actions for Release

chatton:
There were a few issues with the release process. This PR makes some changes.

* Remove the merge action. This was not working correctly as if there are other E2E tests running, this can cause the merge to fail and interrupt the release process. We can leave this as a manual action.
* Only comment on releases **after** the PR is merged. Previously it was commenting too early.

### DIFF
--- a/.github/workflows/comment-release-pr.yml
+++ b/.github/workflows/comment-release-pr.yml
@@ -1,0 +1,21 @@
+name: Link Github Releases
+on:
+  pull_request:
+    types: [closed]
+
+jobs:
+  comment:
+    # only link releases on release PRs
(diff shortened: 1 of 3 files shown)


.github/workflows/comment-release-pr.yml
```diff
@@ -0,0 +1,21 @@
+name: Link Github Releases
+on:
+  pull_request:
+    types: [closed]
+
+jobs:
+  comment:
```
chatton, line 7: moved into separate action which runs only when the Release PR is merged.

irajdeep (approved):

PR #635, opened 2021-07-14
CLOUDP-78289: cr annotations update

FlorentinaSimlinger:
I was not able to reproduce the bug. I changed the spec.version field in the cr yaml file multiple times and it correctly updated the annotation fields lastSuccessfulConfiguration and lastAppliedMongoDBVersion.

However, in the process of debugging I noted the interchangeable use of "last" and "previous" as well as the occasional use of (in this case) ambiguous "old" and "new". 
This was confusing as the field lastSuccessfulConfiguration only becomes the "previous" config once the reconciliation is done/the configuration is successful. Before the reconciliation is done, it is just the current config.
I tried to clear this ambiguity by using "attempted" for the new spec we want, "last successful" for the spec that is in place until it is replaced, and generally avoided the use of "last" for "previous".

I spent quite some time tracing the changing meaning of lastSuccessfulConfiguration and lastAppliedMongoDBVersion and its use in the code, so what I'm hoping for is that it's easier to follow along now. Let me know what you think.

### All Submissions:

### DIFF
--- a/api/v1/mongodbcommunity_types.go
+++ b/api/v1/mongodbcommunity_types.go
@@ -593,6 +593,9 @@ func (m MongoDBCommunity) GetUpdateStrategyType() appsv1.StatefulSetUpdateStrate
 // IsChangingVersion returns true if an attempted version change is occurring.
 func (m MongoDBCommunity) IsChangingVersion() bool {
 	prevVersion := m.getPreviousVersion()
+	fmt.Printf("Previous version of MongoDB is %v \n", string(prevVersion))
+	fmt.Printf("Attempted version of MongoDB is %v \n", string(m.Spec.Version))
+
 	return prevVersion != "" && prevVersion != m.Spec.Version
 }
 
--- a/api/v1/mongodbcommunity_types.go
+++ b/api/v1/mongodbcommunity_types.go
@@ -592,16 +592,13 @@ func (m MongoDBCommunity) GetUpdateStrategyType() appsv1.StatefulSetUpdateStrate
 
 // IsChangingVersion returns true if an attempted version change is occurring.
 func (m MongoDBCommunity) IsChangingVersion() bool {
-	prevVersion := m.getPreviousVersion()
-	fmt.Printf("annotations.LastAppliedMongoDBVersion aka previous version is %v \n", string(prevVersion))
-	fmt.Printf("Attempted version of MongoDB aka m.Spec.Version is %v \n", string(m.Spec.Version))
+	lastVersion := m.getLastVersion()
 
-	return prevVersion != "" && prevVersion != m.Spec.Version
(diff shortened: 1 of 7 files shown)


controllers/replica_set_controller.go
```diff
@@ -248,7 +248,8 @@ func (r ReplicaSetReconciler) Reconcile(ctx context.Context, request reconcile.R
 	return res, err
 }
 
-// updateLastSuccessfulConfiguration annotates the MongoDBCommunity resource with the latest configuration
+// updateLastSuccessfulConfiguration annotates the MongoDBCommunity resource with the configuration
+// that was used to create the resource
```
irajdeep, line 252: > that was used to create the resource

This is called for every reconcile(whenever the spec is updated) -- why is it only annotating with the configuration that was used to create the resource?

FlorentinaSimlinger: Good point

api/v1/mongodbcommunity_types.go
```diff
@@ -592,12 +592,13 @@ func (m MongoDBCommunity) GetUpdateStrategyType() appsv1.StatefulSetUpdateStrate
 
 // IsChangingVersion returns true if an attempted version change is occurring.
 func (m MongoDBCommunity) IsChangingVersion() bool {
-	prevVersion := m.getPreviousVersion()
-	return prevVersion != "" && prevVersion != m.Spec.Version
+	lastVersion := m.getLastVersion()
+
```
irajdeep, line 596: Let's remove this empty line

controllers/replica_set_controller.go
```diff
@@ -261,6 +261,7 @@ func (r *ReplicaSetReconciler) updateLastSuccessfulConfiguration(mdb mdbv1.Mongo
 		// This is needed to reuse the update strategy logic in enterprise
 		lastAppliedMongoDBVersion: mdb.Spec.Version,
 	}
+
```
irajdeep, line 264: same

irajdeep (approved):
LGTM

controllers/validation/validation.go
```diff
@@ -9,12 +9,12 @@ import (
...
 )
 
-// ValidateInitalSpec checks if the resource initial Spec is valid
+// ValidateInitalSpec checks if the resource's initial Spec is valid.
 func ValidateInitalSpec(mdb mdbv1.MongoDBCommunity) error {
 	return validateSpec(mdb)
 }
 
-// ValidateUpdate validates that the new Spec, corresponding to the existing one is still valid
+// ValidateUpdate checks if the attempted Spec is valid and if it is invalidated by the last successful Spec's Security field.
```
irajdeep, line 17: let's revert this comment -- I think the old one is more apt

PR #646, opened 2021-07-20
CLOUDP-95679: Added support for pem entry in tls secret

bznein:
This PR adds the possibility of providing directly a tls.pem entry. 
Existing secrets with tls.crt and tls.key will continue to work, this possibility is already present in enterprise and will bridge the two implementations.

There is also unit and e2e testing.

### All Submissions:


docs/RELEASE_NOTES.md
```diff
@@ -4,7 +4,8 @@
 
 - Changes
   - MongoDB database of the statefulSet is managed using distinct Role, ServiceAccount and RoleBinding.
-
+  - TLS Secret can also contain a single "tls.pem" entry, containing the concatenation of the certificate and key
+    - If a TLS secret contains all three entries, the operator will raise an error if the tls.pem entry is not equal to the concatenation of the tls.crt and tls.key entries
```
irajdeep, line 8: > If a TLS secret contains all three entries

this is a concatenation of `crt` and `key` right? so three ?

api/v1/mongodbcommunity_types.go
```diff
@@ -321,6 +321,8 @@ type TLS struct {
 	// CertificateKeySecret is a reference to a Secret containing a private key and certificate to use for TLS.
 	// The key and cert are expected to be PEM encoded and available at "tls.key" and "tls.crt".
 	// This is the same format used for the standard "kubernetes.io/tls" Secret type, but no specific type is required.
+	// Alternatively, an entry tls.pem, containing the concatenation of cert and key, can be provided.
+	// If all three entries are provided, the tls.pem one needs to be equal to the concatenation fo tls.crt and tls.key
```
irajdeep, line 325:
```suggestion
	// If all three entries are provided, the tls.pem one needs to be equal to the concatenation of tls.crt and tls.key
```

controllers/mongodb_tls.go
```diff
@@ -90,7 +87,7 @@ func getTLSConfigModification(getUpdateCreator secret.GetUpdateCreator, mdb mdbv
 		return automationconfig.NOOP(), nil
 	}
 
-	certKey, err := getCertAndKey(getUpdateCreator, mdb)
+	certKey, err := getPemOrConcatCrtAndKey(getUpdateCreator, mdb)
```
irajdeep, line 90: I think the name `getPemOrConcatCrtAndKey` is a bit confusing , maybe we can remove the `OrConcatCrtAndKey` since it is an implementation details of the function ?

controllers/mongodb_tls.go
```diff
@@ -99,18 +96,27 @@ func getTLSConfigModification(getUpdateCreator secret.GetUpdateCreator, mdb mdbv
 }
 
 // getCertAndKey will fetch the certificate and key from the user-provided Secret.
-func getCertAndKey(getter secret.Getter, mdb mdbv1.MongoDBCommunity) (string, error) {
+func getCertAndKey(getter secret.Getter, mdb mdbv1.MongoDBCommunity) string {
 	cert, err := secret.ReadKey(getter, tlsSecretCertName, mdb.TLSSecretNamespacedName())
 	if err != nil {
-		return "", err
+		return ""
```
irajdeep, line 102: 👍

test/e2e/setup/setup.go
```diff
@@ -56,7 +61,7 @@ func Setup(t *testing.T) *e2eutil.Context {
 
 // CreateTLSResources will setup the CA ConfigMap and cert-key Secret necessary for TLS
 // The certificates and keys are stored in testdata/tls
-func CreateTLSResources(namespace string, ctx *e2eutil.Context) error { //nolint
+func CreateTLSResources(namespace string, ctx *e2eutil.Context, secretType tlsSecretType) error { //nolint
```
irajdeep, line 64: now related to this PR, curious why do we have a `//nolint` here ?

irajdeep (commented):
LGTM overall, added some comments inline.

controllers/mongodb_tls_test.go
```diff
@@ -215,29 +224,89 @@ func TestCombineCertificateAndKey(t *testing.T) {
...
-	err := c.Create(context.TODO(), &s)
-	if err != nil {
-		return err
-	}
+		r := NewReconciler(client.NewManagerWithClient(c))
+
+		err = r.ensureTLSResources(mdb)
+		assert.NoError(t, err)
+
+		// Operator-managed secret should have been created and contain the
```
irajdeep, line 241:
```suggestion
		// Operator-managed secret should have been created and contains the
```

test/e2e/setup/setup.go
```diff
@@ -77,22 +82,29 @@ func CreateTLSResources(namespace string, ctx *e2eutil.Context) error { //nolint
...
+		cert, err := ioutil.ReadFile(path.Join(testDataDir, "server.crt"))
+		if err != nil {
+			return err
+		}
+		key, err := ioutil.ReadFile(path.Join(testDataDir, "server.key"))
+		if err != nil {
+			return err
+		}
+		certKeySecretBuilder.SetField("tls.crt", string(cert)).SetField("tls.key", string(key))
+	} else if secretType == Pem {
```
irajdeep, line 100: do we need an `else if` instead of just `if` , since the secretType can be only one ?

docs/RELEASE_NOTES.md
bznein: I don't get the question😅 We both wrote "three"

irajdeep: of sorry I mean `two` : -)

controllers/mongodb_tls.go
bznein: I called it `getFinalPem` earlier but I didn't like it. We already have `getCertAndKey` and `getPem` so I wanted a name that was clearly different.

test/e2e/setup/setup.go
bznein: good question :) next commit I'll try to remove it and see what happens!

controllers/mongodb_tls.go
irajdeep: I see in that case maybe s/Concat/Concatenated? since the current signature, it looks like the function is performing concatenation -- will leave it upto you

docs/RELEASE_NOTES.md
bznein: I think three is correct: the error happens if the secret contains:

- tls.key
- tls.crt
- tls.pem

and pem is different from key+crt.

Do you think I should reword it to be more clear?

test/e2e/setup/setup.go
bznein: golanci-lint does not raise any error, removed!

docs/RELEASE_NOTES.md
irajdeep: ah got it 
> Do you think I should reword it to be more clear?

yes, please I think it's mentioned at multiple places , we can make it a bit more clearer

bznein: will do, thanks!

irajdeep (approved):
LGTM 👍

PR #647, opened 2021-07-21
Add test labels to test resources and cleanup before running a new test

chatton:
This PR adds the label `e2e-test: true` to resources created by the tests. This allows for easy cleanup. A new make target has been added which runs before running and tests. This causes cleanup of resources before any new tests run.

### DIFF
--- a/pkg/kube/configmap/configmap_builder.go
+++ b/pkg/kube/configmap/configmap_builder.go
@@ -10,6 +10,7 @@ type builder struct {
 	name            string
 	namespace       string
 	ownerReferences []metav1.OwnerReference
+	labels          map[string]string
 }
 
 func (b *builder) SetName(name string) *builder {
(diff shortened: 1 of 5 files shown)


pkg/kube/configmap/configmap_builder.go
```diff
@@ -32,12 +33,22 @@ func (b *builder) SetOwnerReferences(ownerReferences []metav1.OwnerReference) *b
...
 
+func (b *builder) SetLabels(labels map[string]string) *builder {
+	newLabels := make(map[string]string)
+	for k, v := range labels {
+		newLabels[k] = v
+	}
+	b.labels = newLabels
+	return b
+}
+
```
killian2k, line 44: If we use SetLabels multiple times in a builder object. Should we keep only the labels from the last function call or merge all the maps and labels together?

killian2k (commented):
LGTM

pkg/kube/configmap/configmap_builder.go
chatton: That's a good question, I think with a name like `Set` we indicate that this will overwrite all labels, if we need the functionality to add labels incrementally this might be an `AddLabel` function.

bznein (approved):
LGTM!

PR #654, opened 2021-07-26
Cloudp 95948 integrate helm to makefile

killian2k:
I modified the makefile to replace kustomize with helm


Makefile
```diff
@@ -31,27 +31,43 @@ manager: generate fmt vet
...
 	eval $$(scripts/dev/get_e2e_env_vars.py $(cleanup)); \
 	go run ./cmd/manager/main.go
 
 # Install CRDs into a cluster
-install: manifests kustomize
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
+install: manifests helm install_crd
+	
+
+install_crd:
```
chatton, line 42: small nit, the other make targets use `kebab-case`, we should make sure that our naming conventions are consistent!

Makefile
```diff
@@ -31,27 +31,43 @@ manager: generate fmt vet
...
+uninstall: manifests helm uninstall_rbac uninstall_crd
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-deploy: manifests kustomize
-	cd config/manager && $(KUSTOMIZE) edit set image quay.io/mongodb/mongodb-kubernetes-operator=$(IMG):latest
-	$(KUSTOMIZE) build config/default | kubectl apply -n $(NAMESPACE) -f -
+deploy: manifests helm
+	install_crd
+	install_rbac
+	install_manager
```
chatton, line 66: previously we were dynamically subbing in a value from `IMG` (which comes from the json config file). What it looks like this will do is just apply manager.yaml, is that correct?

One other thing to note, I see that you have using several helm template commands, one for each template file, we don't necessarily need to separate these out this much. We could do a single `helm install helm-chart --set key1=val1 --set key2=val2` etc.

Makefile
```diff
@@ -31,27 +31,43 @@ manager: generate fmt vet
...
+
+install_rbac:
+	$(HELM) template -s templates/tests/operator_roles.yaml helm-chart | kubectl  -n $(NAMESPACE) apply -f -
+	$(HELM) template -s templates/tests/database_roles.yaml helm-chart | kubectl  -n $(NAMESPACE) apply -f -
+
+install_manager:
+	kubectl apply -f -n $(NAMESPACE) config/manager/manager.yaml
+
+uninstall_crd:
+	k delete crd mongodbcommunity.mongodbcommunity.mongodb.com
```
chatton, line 53: we can't assume everyone will have the alias `k` set! We should use `kubectl` here

Makefile
```diff
@@ -31,27 +31,43 @@ manager: generate fmt vet
...
+
+uninstall_rbac: 
+	$(HELM) template -s templates/tests/operator_roles.yaml helm-chart | kubectl delete -f -
```
chatton, line 56: Is it possible to use `helm uninstall` https://helm.sh/docs/helm/helm_uninstall/ for this rather than templating and deleting?

Makefile
```diff
@@ -124,10 +140,23 @@ CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
...
-KUSTOMIZE = $(shell pwd)/bin/kustomize
-kustomize:
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.2.0)
+# Download helm locally if necessary
+HELM = $(shell pwd)/bin/helm
+helm:
+	$(call install_helm)
+
+
+define install_helm
```
chatton, line 149: 👍

chatton (commented):
Looking good so far!

Makefile
killian2k: Ah yes, I missed that one 🤭

Makefile
killian2k: we also need to install the CRD and RBAC as well here I think. Because in the current version of the makefile we call `$(KUSTOMIZE) build config/default | kubectl apply -n $(NAMESPACE) -f -` which install the manager but also the crd and rbac

Makefile
```diff
@@ -31,27 +34,36 @@ manager: generate fmt vet
...
-install: manifests kustomize
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
+install: manifests helm install-crd
+	
+
+install-crd:
+	kubectl apply -f config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
+
+install-chart:
+#	$(HELM) template --set namespace=$(NAMESPACE) helm-chart  | kubectl  --namespace=$(NAMESPACE) apply -f -
```
chatton, line 49: I think we can remove this

chatton (approved):
LGTM nice work getting this set up 👍

PR #655, opened 2021-07-26
CLOUDP-95181: version is missing

FlorentinaSimlinger:
This PR adds Version to the Status field and correspondingly changes the tests.

### All Submissions:

### DIFF
--- a/controllers/mongodb_status_options.go
+++ b/controllers/mongodb_status_options.go
@@ -33,7 +33,7 @@ func (o *optionBuilder) GetOptions() []status.Option {
 	return o.options
 }
 
-// StatusOptions returns an initialized optionBuilder
+// statusOptions returns an initialized optionBuilder
 func statusOptions() *optionBuilder {
 	return &optionBuilder{
 		options: []status.Option{},
--- a/controllers/mongodb_status_options.go
+++ b/controllers/mongodb_status_options.go
@@ -60,8 +60,8 @@ func (m mongoUriOption) GetResult() (reconcile.Result, error) {
 	return result.OK()
 }
 
-func (o *optionBuilder) withVersion() *optionBuilder {
-	version := getVersion()
+func (o *optionBuilder) withVersion(mdb *mdbv1.MongoDBCommunity) *optionBuilder {
+	version := mdb.GetMongoDBVersion()
 	o.options = append(o.options,
 		versionOption{
 			version: version,
--- a/controllers/mongodb_status_options.go
+++ b/controllers/mongodb_status_options.go
@@ -22,7 +22,7 @@ const (
 	None  severity = "NONE"
 )
 
-// optionBuilder is in charge of constructing a slice of options that
+// OptionBuilder is in charge of constructing a slice of options that
 // will be applied on top of the MongoDB resource that has been provided
 type optionBuilder struct {
 	options []status.Option
@@ -33,7 +33,7 @@ func (o *optionBuilder) GetOptions() []status.Option {
 	return o.options
 }
 
-// options returns an initialized optionBuilder
+// StatusOptions returns an initialized optionBuilder
 func statusOptions() *optionBuilder {
 	return &optionBuilder{
 		options: []status.Option{},
@@ -57,6 +57,27 @@ func (m mongoUriOption) ApplyOption(mdb *mdbv1.MongoDBCommunity) {
 }
 
 func (m mongoUriOption) GetResult() (reconcile.Result, error) {
+	return result.OK()
+}
+
+func (o *optionBuilder) withVersion() *optionBuilder {
+	version := getVersion()
+	o.options = append(o.options,
+		versionOption{
+			version: version,
+		})
+	return o
+}
+
+type versionOption struct {
+	version string
+}
+
+func (v versionOption) ApplyOption(mdb *mdbv1.MongoDBCommunity) {
+	mdb.Status.Version = v.version
+}
+
+func (v versionOption) GetResult() (reconcile.Result, error) {
 	return result.OK()
 }
 
--- a/controllers/mongodb_status_options.go
+++ b/controllers/mongodb_status_options.go
@@ -22,7 +22,7 @@ const (
 	None  severity = "NONE"
 )
 
-// OptionBuilder is in charge of constructing a slice of options that
+// optionBuilder is in charge of constructing a slice of options that
 // will be applied on top of the MongoDB resource that has been provided
 type optionBuilder struct {
 	options []status.Option
--- a/controllers/mongodb_status_options.go
+++ b/controllers/mongodb_status_options.go
@@ -60,8 +60,7 @@ func (m mongoUriOption) GetResult() (reconcile.Result, error) {
 	return result.OK()
 }
 
-func (o *optionBuilder) withVersion(mdb *mdbv1.MongoDBCommunity) *optionBuilder {
-	version := mdb.GetMongoDBVersion()
+func (o *optionBuilder) withVersion(version string) *optionBuilder {
 	o.options = append(o.options,
 		versionOption{
 			version: version,
--- a/controllers/mongodb_status_options_test.go
+++ b/controllers/mongodb_status_options_test.go
@@ -9,9 +9,11 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const testVersion string = "4.2.6"
+
 func TestMongoUriOption_ApplyOption(t *testing.T) {
 
-	mdb := newReplicaSet(3, "4.2.6", "my-rs", "my-ns")
+	mdb := newReplicaSet(3, testVersion, "my-rs", "my-ns")
 
 	opt := mongoUriOption{
 		mongoUri: "my-uri",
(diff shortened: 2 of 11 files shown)


controllers/mongodb_status_options.go
```diff
@@ -22,7 +22,7 @@ const (
 	None  severity = "NONE"
 )
 
-// optionBuilder is in charge of constructing a slice of options that
+// OptionBuilder is in charge of constructing a slice of options that
```
irajdeep, line 25: this should match the function signature -- so should be `optionBuilder` -- let's revert this change.

controllers/mongodb_status_options.go
```diff
@@ -33,7 +33,7 @@ func (o *optionBuilder) GetOptions() []status.Option {
 	return o.options
 }
 
-// options returns an initialized optionBuilder
+// StatusOptions returns an initialized optionBuilder
```
irajdeep, line 36:
```suggestion
// statusOptions returns an initialized optionBuilder
```

controllers/mongodb_status_options_test.go
```diff
@@ -23,30 +23,42 @@ func TestMongoUriOption_ApplyOption(t *testing.T) {
 }
 
 func TestOptionBuilder_RunningPhase(t *testing.T) {
-	mdb := newReplicaSet(3, "my-rs", "my-ns")
+	mdb := newReplicaSet(3, "4.2.6", "my-rs", "my-ns")
```
irajdeep, line 26: Can we make `4.2.6` a constant? -- would be easier to make changes later

controllers/mongodb_status_options_test.go
```diff
@@ -23,30 +23,42 @@ func TestMongoUriOption_ApplyOption(t *testing.T) {
...
 }
 
-func newReplicaSet(members int, name, namespace string) mdbv1.MongoDBCommunity {
+func TestVersion_ApplyOption(t *testing.T) {
+	mdb := newReplicaSet(3, "4.2.6", "my-rs", "my-ns")
+
+	opt := versionOption{
+		version: "4.2.6",
+	}
+
```
irajdeep, line 55: nit: let's remove empty line here

test/e2e/replica_set_scale/replica_set_scaling_test.go
```diff
@@ -53,6 +53,7 @@ func TestReplicaSetScaleUp(t *testing.T) {
 			mdbv1.MongoDBCommunityStatus{
 				MongoURI:                   mdb.MongoURI(),
 				Phase:                      mdbv1.Running,
+				Version:                    mdb.Spec.Version,
```
irajdeep, line 56: we can call `mdb.GetMongoDBVersion` here

irajdeep (commented):
LGTM overall,  added some minor comments.

test/e2e/replica_set_recovery/replica_set_recovery_test.go
```diff
@@ -58,6 +58,7 @@ func TestReplicaSetRecovery(t *testing.T) {
 			mdbv1.MongoDBCommunityStatus{
 				MongoURI:                   mdb.MongoURI(),
 				Phase:                      mdbv1.Running,
+				Version:                    mdb.Spec.Version,
```
irajdeep, line 61: we can call mdb.GetMongoDBVersion here

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -310,6 +310,7 @@ func BasicFunctionality(mdb *mdbv1.MongoDBCommunity) func(*testing.T) {
 			mdbv1.MongoDBCommunityStatus{
 				MongoURI:                   mdb.MongoURI(),
 				Phase:                      mdbv1.Running,
+				Version:                    mdb.Spec.Version,
```
irajdeep, line 313: we can use `mdb.GetMongoDBVersion` here as well

test/e2e/replica_set_scale_down/replica_set_scale_down_test.go
```diff
@@ -56,6 +56,7 @@ func TestReplicaSetScaleDown(t *testing.T) {
 			mdbv1.MongoDBCommunityStatus{
 				MongoURI:                   mdb.MongoURI(),
 				Phase:                      mdbv1.Running,
+				Version:                    mdb.Spec.Version,
```
irajdeep, line 59: ditto

controllers/mongodb_status_options.go
```diff
@@ -60,6 +60,27 @@ func (m mongoUriOption) GetResult() (reconcile.Result, error) {
 	return result.OK()
 }
 
+func (o *optionBuilder) withVersion(mdb *mdbv1.MongoDBCommunity) *optionBuilder {
```
bznein, line 63: [q] why are we passing `mdb` instead of simply the version string?

controllers/replica_set_controller.go
```diff
@@ -224,7 +224,8 @@ func (r ReplicaSetReconciler) Reconcile(ctx context.Context, request reconcile.R
 			withMongoDBMembers(mdb.AutomationConfigMembersThisReconciliation()).
 			withStatefulSetReplicas(mdb.StatefulSetReplicasThisReconciliation()).
 			withMessage(None, "").
-			withRunningPhase(),
+			withRunningPhase().
+			withVersion(&mdb),
```
bznein, line 228: related to the other comment, I think `withVersion(mdb.GetMongoDBVersion())` would be much clearer

bznein (commented):
Looks good! I just have one comment!

rodrigovalin (commented):
@FlorentinaSimlinger is it possible to display what we have in `spec.version` instead of adding a new attribute to status -- an attribute that will be identical to what we already have in `spec.`


cc'ing @bznein as the reporter.

test/e2e/replica_set_scale_down/replica_set_scale_down_test.go
FlorentinaSimlinger: good catch, I'll check again to find all other occurrences

controllers/mongodb_status_options.go
FlorentinaSimlinger: good point, I'll change it to passing the string

irajdeep:
> is it possible to display what we have in spec.version instead of adding a new attribute to status -- an attribute that will be identical to what we already have in spec.

Do you mean not to have the `version` field in the CR status altogether? I think this is more to align with what we have in the enterprise where we show the `version` in the CR status as well ( -- @bznein could correct here if wrong)

bznein:
> > is it possible to display what we have in spec.version instead of adding a new attribute to status -- an attribute that will be identical to what we already have in spec.
> 
> Do you mean not to have the `version` field in the CR status altogether? I think this is more to align with what we have in the enterprise where we show the `version` in the CR status as well ( -- @bznein could correct here if wrong)

I said "status" in the ticket after a quick discussion with Cian. I don't have a strong opinion on this, I think both solutions are ok

irajdeep:
I think one advantage of having the version in the status is to view it at "one place" when running `kubectl get mdbc` when there are multiple CRs in the cluster.

rodrigovalin:
We originally added `version` and a few other attributes to the status object of the enterprise resources because we were holding a "state"; when we designed the community operator we decided to use the "status" as an informational resource that won't have repeated data to hold a state for the reconcile process. This data is stored mostly in annotations right now.

Considering this design, I think it is better if we use `spec.version` to display the version in the mongodb resource.

> I think one advantage of having the version in the status is to view it at "one place" when running kubectl get mdbc when there are multiple CRs in the cluster.

We can have it displayed with `kubectl get mdbc` if we add `.jsonPath: .spec.version` to `additionalPrinterColumns`

irajdeep:
> We can have it displayed with kubectl get mdbc if we add .jsonPath: .spec.version to additionalPrinterColumns

 SGTM

bznein:
> We can have it displayed with `kubectl get mdbc` if we add `.jsonPath: .spec.version` to `additionalPrinterColumns`

SGTM Although this might be slightly not in sync with the correct status. We update the status when the reconciliation is finished, while `spec.version` would reflect the desired version rather than the current one

rodrigovalin (approved):
> SGTM Although this might be slightly not in sync with the correct status. We update the status when the reconciliation is finished, while spec.version would reflect the desired version rather than the current one

Ok I think this is the right answer to base our conclussion ... If this is the case, I think we should keep displaying the status field, beause it is the "current" version, while the `spec` one is the "desired" one.

In that case we should be good to merge.

LGTM

chatton (commented):
I think displaying the version that will be seen in the describe output is a strong reason to keep it. Maintaining multiple resources could become more difficult to manage without it. It is also quite common to list versions with different resources when relevant. (e.g. nodes etc.)

FlorentinaSimlinger:
> SGTM Although this might be slightly not in sync with the correct status. We update the status when the reconciliation is finished, while spec.version would reflect the desired version rather than the current one

Good point, we could use "lastSuccessfulMondoDBVersion" in annotations then -- that would mean that it accurately shows the version even though we might change it during the reconciliation

bznein:
I just noticed that we didn't include this in the Release Notes, I think it's a small fix but worth mentioning, would you mind doing a quick PR to add it?

PR #662, opened 2021-07-29
CLOUDP-95657: Make TLS code exportable

bznein:
This PR makes TLS code exportable in order to import it into enterprise. Most of the diff is just renaming, plus a few adjustment to make it possible to use it in enterprise. Comments inline

### All Submissions:


api/v1/mongodbcommunity_types.go
```diff
@@ -620,6 +620,30 @@ func (m MongoDBCommunity) LogsVolumeName() string {
...
+	return "tls.key"
+}
+
+func (m MongoDBCommunity) TLSSecretCertName() string {
+	return "tls.crt"
+}
+
+func (m MongoDBCommunity) TLSSecretPEMName() string {
+	return "tls.pem"
+}
```
bznein, line 641: All of these are constant but wrapped in a function as enterprise might provide different entries

controllers/tls/mongodb_tls.go
```diff
@@ -15,50 +16,66 @@ import (
...
 		return false, err
 	}
 
-	// Ensure ConfigMap has a "ca.crt" field
-	if cert, ok := caData[tlsCACertName]; !ok || cert == "" {
-		r.log.Warnf(`ConfigMap "%s" should have a CA certificate in field "%s"`, mdb.TLSConfigMapNamespacedName(), tlsCACertName)
-		return false, nil
+	// Ensure ConfigMap has all the required fields
+	for _, key := range mdb.TLSConfigMapRequiredEntries() {
+		if cert, ok := caData[key]; !ok || cert == "" {
```
bznein, line 68: Enterprise requires two entries in the CM, so we need to add a loop which here will execute only once

controllers/tls/mongodb_tls.go
```diff
@@ -96,44 +113,44 @@ func getTLSConfigModification(getUpdateCreator secret.GetUpdateCreator, mdb mdbv
...
 	}
 
-	return combineCertificateAndKey(cert, key)
+	return CombineCertificateAndKey(cert, key)
 }
 
 // getPem will fetch the pem from the user-provided secret
-func getPem(getter secret.Getter, mdb mdbv1.MongoDBCommunity) string {
-	pem, err := secret.ReadKey(getter, tlsSecretPemName, mdb.TLSSecretNamespacedName())
+func getPem(getter secret.Getter, mdb TLSResource, secretName types.NamespacedName) string {
```
bznein, line 131: We need to add the `secretName` as a parameter here since in enterprise we also support shared clusters, and in that there isn't a single secret name we can get from the resource, so we will pass it to this function

controllers/tls/mongodb_tls.go
```diff
@@ -96,44 +113,44 @@ func getTLSConfigModification(getUpdateCreator secret.GetUpdateCreator, mdb mdbv
...
 
-// getPemOrConcatenatedCrtAndKey will get the final PEM to write to the secret.
+// GetPemOrConcatenatedCrtAndKey will get the final PEM to write to the secret.
 // This is either the tls.pem entry in the given secret, or the concatenation
 // of tls.crt and tls.key
 // It performs a basic validation on the entries.
-func getPemOrConcatenatedCrtAndKey(getter secret.Getter, mdb mdbv1.MongoDBCommunity) (string, error) {
-	certKey := getCertAndKey(getter, mdb)
-	pem := getPem(getter, mdb)
+func GetPemOrConcatenatedCrtAndKey(getter secret.Getter, mdb TLSResource, secretName types.NamespacedName) (string, error) {
```
bznein, line 149: same thing about `secretName` here

controllers/tls/mongodb_tls.go
```diff
@@ -15,50 +16,66 @@ import (
...
 const (
-	tlsCAMountPath             = "/var/lib/tls/ca/"
-	tlsCACertName              = "ca.crt"
-	tlsOperatorSecretMountPath = "/var/lib/tls/server/" //nolint
-	tlsSecretCertName          = "tls.crt"              //nolint
-	tlsSecretKeyName           = "tls.key"
-	tlsSecretPemName           = "tls.pem"
+	TLSCAMountPath             = "/var/lib/tls/ca/"
+	TLSCACertName              = "ca.crt"
+	TLSOperatorSecretMountPath = "/var/lib/tls/server/" //nolint
```
chatton, line 30: [nit] we no longer need the `TLS` prefix since we are in a `tls` package.

controllers/tls/mongodb_tls.go
```diff
@@ -15,50 +16,66 @@ import (
...
+	IsTLSOptional() bool
+	TLSConfigMapNamespacedName() types.NamespacedName
+	TLSSecretNamespacedName() types.NamespacedName
+	NamespacedName() types.NamespacedName
+	TLSOperatorSecretNamespacedName() types.NamespacedName
+	GetOwnerReferences() []metav1.OwnerReference
+	TLSSecretKeyName() string
+	TLSSecretCertName() string
+	TLSSecretPEMName() string
+	TLSConfigMapRequiredEntries() []string
```
chatton, line 44: I think this interface could hugely benefit from docstrings for each of the methods.

controllers/tls/mongodb_tls.go
```diff
@@ -15,50 +16,66 @@ import (
...
-	tlsOperatorSecretMountPath = "/var/lib/tls/server/" //nolint
-	tlsSecretCertName          = "tls.crt"              //nolint
-	tlsSecretKeyName           = "tls.key"
-	tlsSecretPemName           = "tls.pem"
+	TLSCAMountPath             = "/var/lib/tls/ca/"
+	TLSCACertName              = "ca.crt"
+	TLSOperatorSecretMountPath = "/var/lib/tls/server/" //nolint
 )
 
+type TLSResource interface {
```
chatton, line 33: this interface is a bit bulky still, do you think it would be possible/worthwhile to try to break it down into several smaller interfaces?

chatton (approved):
LGTM! Just had a few small comments, but no blockers 👍

PR #686, opened 2021-08-16
Add all helm values to makefile

chatton:
We were missing a few values when using helm in a dev environment. This PR adds a set of values requires to run `make deploy` and then deploy a CR.


irajdeep (approved):
LGTM

PR #689, opened 2021-08-17
CLOUDP-95948: Refactor E2E tests to use HELM

chatton:
This PR migrates our tests to use the new HELM chart.

* The operator deployment, roles/service account/role bindings are now created using the helm chart.
* Helm is bundled into the e2e image and is used for installation.


test/e2e/setup/setup.go
```diff
@@ -160,52 +191,41 @@ func deployOperator(ctx *e2eutil.Context) error {
...
-	if err := buildKubernetesResourceFromYamlFile(ctx, path.Join(roleDir(), "service_account_database.yaml"), &corev1.ServiceAccount{}, withNamespace(testConfig.namespace)); err != nil {
-		return errors.Errorf("error building mongodb database service account: %s", err)
+	quantityCPU, err := resource.ParseQuantity("50m")
+	if err != nil {
+		return err
 	}
-	if err := buildKubernetesResourceFromYamlFile(ctx, path.Join(roleDir(), "role_binding_database.yaml"), &rbacv1.RoleBinding{}, withNamespace(testConfig.namespace)); err != nil {
-		return errors.Errorf("error building mongodb database role binding: %s", err)
+
+	for _, cont := range dep.Spec.Template.Spec.Containers {
```
chatton, line 215: reduce resources used by the operator as this can cause issues with the tests if they are too high.

Makefile
```diff
@@ -9,12 +9,12 @@ READINESS_PROBE_IMG := $(shell jq -r .readiness_probe_image < ~/.community-opera
 REGISTRY := $(shell jq -r .repo_url < ~/.community-operator-dev/config.json)
 AGENT_IMAGE_NAME := $(shell jq -r .agent_image_ubuntu < ~/.community-operator-dev/config.json)
 
-STRING_SET_VALUES := --set namespace=$(NAMESPACE),version_upgrade_hook.name=$(UPGRADE_HOOK_IMG),readiness_probe.name=$(READINESS_PROBE_IMG),registry.operator=$(REPO_URL),operator.operator_image_name=$(OPERATOR_IMAGE),operator.version=latest,registry.agent=$(REGISTRY),registry.version_upgrade_hook=$(REGISTRY),registry.readiness_probe=$(REGISTRY),registry.operator=$(REGISTRY),version_upgrade_hook.version=latest,readiness_probe.version=latest,agent.version=latest,agent.name=$(AGENT_IMAGE_NAME)
+STRING_SET_VALUES := --set namespace=$(NAMESPACE),versionUpgradeHook.name=$(UPGRADE_HOOK_IMG),readinessProbe.name=$(READINESS_PROBE_IMG),registry.operator=$(REPO_URL),operator.operatorImageName=$(OPERATOR_IMAGE),operator.version=latest,registry.agent=$(REGISTRY),registry.versionUpgradeHook=$(REGISTRY),registry.readinessProbe=$(REGISTRY),registry.operator=$(REGISTRY),versionUpgradeHook.version=latest,readinessProbe.version=latest,agent.version=latest,agent.name=$(AGENT_IMAGE_NAME)
```
chatton, line 12: changed everything to be consistent with camelCase

helm-chart/templates/operator_roles.yaml
```diff
@@ -70,6 +70,9 @@ metadata:
 subjects:
 - kind: ServiceAccount
   name: {{ .Values.operator.name }}
+{{ if eq (.Values.operator.watchNamespace | default "") "*" }}
```
chatton, line 73: the namespace if we are running in clusterwide mode

scripts/dev/templates/Dockerfile.e2e
```diff
@@ -3,3 +3,12 @@
...
 {% endblock -%}
+
+
+{% block install_helm -%}
+RUN curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl" \
+    && chmod +x ./kubectl \
+    && mv ./kubectl /usr/local/bin/kubectl \
+    && curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 \
+    && chmod +x get_helm.sh && ./get_helm.sh
+{% endblock -%}
```
bznein, line 14: I just trust you on this :D 

minor nit: can you put `./get_helm.sh` on a separate line as we have each command on a separate line apart from the last 2?

pkg/helm/helm.go
```diff
@@ -0,0 +1,45 @@
...
+	helmArgs := []string{"uninstall", chartName}
+	cmd := exec.Command("helm", helmArgs...)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		if strings.Contains(string(output), "not found") {
+			return nil
+		}
+		return fmt.Errorf("error executing command: %s %s", err, output)
+	}
+	return nil
```
bznein, line 21: [q/nit] WDYT of extracting this and the one below to a function like

```golang
executeHelmCommand(args []string, allowedErrors string...) error {
	cmd := exec.Command("helm", helmArgs...)
	output, err := cmd.CombinedOutput()
	if err != nil {
		// check if the error contains one of the allowedErrors, return nil in that case
		return fmt.Errorf("error executing command: %s %s", err, output)
	}
	return nil
}
```

?

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -36,7 +36,7 @@ func SkipTestIfLocal(t *testing.T, msg string, f func(t *testing.T)) {
 func StatefulSetBecomesReady(mdb *mdbv1.MongoDBCommunity, opts ...wait.Configuration) func(t *testing.T) {
 	defaultOpts := []wait.Configuration{
 		wait.RetryInterval(time.Second * 15),
-		wait.Timeout(time.Minute * 15),
+		wait.Timeout(time.Minute * 20),
```
bznein, line 39: [q] I know you changed this a few times for testing, is it needed now?

test/e2e/setup/setup.go
```diff
@@ -140,18 +133,55 @@ func GeneratePasswordForUser(ctx *e2eutil.Context, mdbu mdbv1.MongoDBUser, names
...
-func deployDir() string {
-	return envvar.GetEnvOrDefault(deployDirEnv, "/workspace/config/manager")
+	splitString = strings.Split(splitString[len(splitString)-1], ":")
+	version := "latest"
+	if len(splitString) > 1 {
+		version = splitString[len(splitString)-1]
+	}
+	name := splitString[0]
+	return registry, name, version
 }
```
bznein, line 148: This relies on the image being a repo:tag and not with the SHA digest right?
I think this is fine but we could maybe document it in the docstring?

We could even check and maybe return an error, but since this is only in the e2e setup I think we are fine for now!

bznein (approved):
LGTM a few small comments!

pkg/helm/helm.go
chatton: yes makes sense!

test/e2e/mongodbtests/mongodbtests.go
chatton: ah nice catch, right we shouldn't need this anymore.

PR #706, opened 2021-09-02
CLOUDP-98684 helm chart - integrate cert-manager

tibulca:


cndoit18 (approved):
I like this PR

controllers/mongodb_tls.go
```diff
@@ -37,11 +39,20 @@ func (r *ReplicaSetReconciler) validateTLSConfig(mdb mdbv1.MongoDBCommunity) (bo
...
+	var caResourceName types.NamespacedName
+	var caResourceType string
+	if mdb.Spec.Security.TLS.CaCertificateSecret != nil {
+		caResourceName = mdb.TLSCaCertificateSecretNamespacedName()
+		caResourceType = "Secret"
+	} else {
+		caResourceName = mdb.TLSConfigMapNamespacedName()
+		caResourceType = "ConfigMap"
+	}
+	caData, err := configmap.ReadData(r.client, caResourceName)
```
bznein, line 52: [q] how does this work? I understand if the CA is passed through config map, but if it's a secret how can we read it? I expect it to fail?

bznein (commented):
LGTM in general, but there is one line I can't understand

controllers/mongodb_tls.go
tibulca: thanks for spotting this. It was working probably because the secret and the config map have a similar struct signature but I made the change so that we're explicitly reading a secret or a configmap

helm-chart/templates/cr.yaml
```diff
@@ -0,0 +1,79 @@
+{{- if and .Values.createResource .Values.resource.tls.enabled .Values.resource.tls.useCertManager }}
```
bznein, line 1: can we give a better name to this file? `cr` is very generic

bznein (approved):

helm-chart/templates/cr.yaml
tibulca: renamed to mongodbcommunity_cr_with_tls.yaml

PR #709, opened 2021-09-06
CLOUDP-99765 use cert-manager to generate tls certificates for e2e tests

tibulca:


irajdeep (approved):
Looks good, just one question: do we need to modify the getting started docs with this change? for ex: previously we were pinned to `default` namespace for tls tests.

pkg/helm/helm.go
```diff
@@ -8,16 +8,22 @@ import (
...
+func Uninstall(chartName string, namespace string) error {
+	helmArgs := []string{"uninstall", chartName, "-n", namespace}
 	return executeHelmCommand(helmArgs, isNotFoundMessage)
 }
 
 // Install a helm chert at the given path with the given name and the provided set arguments.
-func Install(chartPath, chartName string, args map[string]string) error {
-	helmArgs := []string{"install", "--skip-crds"}
-	helmArgs = append(helmArgs, mapToHelmArgs(args)...)
+func Install(chartPath, chartName string, flags map[string]string, values map[string]string) error {
```
chatton, line 17: do we need to pass two different sets of flags? I think this could be a single map right? I believe the current `values` map could be refactored slightly to pass `--set` with the value to the new `flags` map with the new approach. What do you think?

test/e2e/setup/setup.go
```diff
@@ -228,6 +199,25 @@ func deployOperator() error {
 	return nil
 }
 
+func deployCertManager(config TestConfig) error {
+	const helmChartName = "cert-manager"
+	if err := helm.Uninstall(helmChartName, config.CertManagerNamespace); err != nil {
+		return err
```
chatton, line 205: [q] do you think it's worth wrapping the error with context saying whether or not the error occurred during uninstall vs/install?

test/e2e/util/mongotester/mongotester.go
```diff
@@ -254,21 +227,19 @@ func (m *Tester) connectivityCheck(shouldSucceed bool, opts ...OptionApplier) fu
...
 		// Reject all server certificates that don't have the expected serial number
 		tls.VerifyPeerCertificate = func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
 			cert := verifiedChains[0][0]
-			if expectedSerial.Cmp(cert.SerialNumber) != 0 {
-				return errors.Errorf("expected certificate serial number %s, got %s", expectedSerial, cert.SerialNumber)
+			if initialCertSerialNumber.Cmp(cert.SerialNumber) == 0 {
+				return errors.Errorf("certificate serial number has not changed: %s", cert.SerialNumber)
+			} else {
+				return nil
 			}
```
chatton, line 242: redundant else clause, we can just `return nil` here

test/e2e/util/mongotester/mongotester.go
```diff
@@ -470,20 +450,36 @@ func WithReplicaSet(rsname string) OptionApplier {
...
-	// Read the CA certificate from test data
-	testDataDir := e2eutil.TlsTestDataDir()
-	caPEM, err := ioutil.ReadFile(path.Join(testDataDir, "ca.crt"))
+func getClientTLSConfig(mdb mdbv1.MongoDBCommunity) (*tls.Config, error) {
+	caSecret := corev1.Secret{}
+	caSecretName := types.NamespacedName{Name: mdb.Spec.Security.TLS.CaCertificateSecret.Name, Namespace: mdb.Namespace}
+	err := e2eutil.TestClient.Get(context.TODO(), caSecretName, &caSecret)
 	if err != nil {
 		return nil, err
+	} else {
```
chatton, line 459: no need for else here

test/e2e/util/mongotester/mongotester.go
```diff
@@ -470,20 +450,36 @@ func WithReplicaSet(rsname string) OptionApplier {
...
-		RootCAs: caPool,
-	}, nil
+// GetClientCert reads the client key certificate
+func GetClientCert(mdb mdbv1.MongoDBCommunity) (*x509.Certificate, error) {
+	certSecret := corev1.Secret{}
+	certSecretName := types.NamespacedName{Name: mdb.Spec.Security.TLS.CertificateKeySecret.Name, Namespace: mdb.Namespace}
+	err := e2eutil.TestClient.Get(context.TODO(), certSecretName, &certSecret)
+	if err != nil {
+		return nil, err
+	} else {
```
chatton, line 476: same comment as above

test/e2e/util/mongotester/mongotester.go
```diff
@@ -432,7 +403,16 @@ func WithHosts(hosts []string) OptionApplier {
 }
 
 // WithTls configures the client to use tls
-func WithTls() OptionApplier {
+func WithTls(mdb mdbv1.MongoDBCommunity) OptionApplier {
+	if !mdb.Spec.Security.TLS.Enabled {
```
chatton, line 407: one thing to consider, we have the different options so that we can test resources in arbitrary states with arbitrary mongo connectivity options, the panic here means we wouldn't be able to perform a test which ensures that attempting a connection to a non tls enabled MDB would fail. I don't think we'll ever need to test it, but the entire test pod itself will crash here rather than the specific test failing. What do you think?

chatton (commented):
Looks great! Thanks for the hard work on this, I had a few questions/comments/suggestions

pkg/helm/helm.go
tibulca: as discussed on zoom, I will rename values to templateValues and keep 2 maps for keeping a clear distinction between helm arguments and template arguments

test/e2e/setup/setup.go
tibulca: good point, done

test/e2e/util/mongotester/mongotester.go
tibulca: removed

test/e2e/util/mongotester/mongotester.go
tibulca: removed

test/e2e/util/mongotester/mongotester.go
tibulca: removed

test/e2e/util/mongotester/mongotester.go
tibulca: agree, I removed this check

chatton (approved):
LGTM! 🥇

PR #718, opened 2021-09-15
CLOUDP-100089 make MongoDBCommunityStatus.Version optional

tibulca:


config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
```diff
@@ -340,7 +340,6 @@ spec:
             - currentStatefulSetReplicas
             - mongoUri
             - phase
-            - version
```
tibulca, line 343: looks like this field was manually removed from the required list by this [PR 684](https://github.com/mongodb/mongodb-kubernetes-operator/pull/684/files) but we missed to update the type (tag the version field with "omitempty")

irajdeep (approved):

PR #745, opened 2021-10-08
CLOUDP-101992: Adds helm chart as submodule

rodrigovalin:
## Summary

Required before merge: https://github.com/mongodb/helm-charts/pull/61 (and point helm-charts to commit SHA in main after merging)

This patch exports the contents of the `helm-chart` into the [MongoDB Helm Charts repo](https://github.com/mongodb/helm-charts.git). The `helm-charts` repo will be the only place where people will find Helm charts for the different MongoDB Operators.

Currently `helm-repos` hosts the `atlas-operator` Helm chart. And will hold Community (this) and Enterprise in the future. Ideally users will do:

```
helm repo add mongodb https://mongodb.github.io/helm-charts
helm install <any mongodb operator>
```

In this patch we remove the `helm-chart` directory and replaced it with a git submodule that points at the `helm-charts` repository. The tests have been modified so the contents of the submodule are fetched (using `with: submodule: true` in `actions/checkout`). This is a process that should be done locally as well with:

```
git submodule init
git submodule update
```

## Next

Next steps are:

1. Update the Helm release process on `helm-charts` repo.
2. Update documentation on `helm-charts` indicating how to use this chart.


.github/workflows/e2e-dispatch.yml
```diff
@@ -50,6 +50,8 @@ jobs:
     # template: .action_templates/steps/checkout.yaml
```
chatton, line 50: the submodule changes in this file should instead be added to https://github.com/mongodb/mongodb-kubernetes-operator/blob/fb205a37567b9cc8edc89503bc8f3ddcba249317/.action_templates/steps/checkout.yaml#L1 and https://github.com/mongodb/mongodb-kubernetes-operator/blob/fb205a37567b9cc8edc89503bc8f3ddcba249317/.action_templates/steps/checkout-fork.yaml#L1-L0 instead.

These are the templates that are used to generate the files under `.github` in the pre-commit hook

PR #752, opened 2021-10-19
Update path to helm-chart

chatton:
### All Submissions:

This assumes you have checked out `mongodb/helm-charts` at the same level as the community repo.


rodrigovalin:
@chatton you should do

```
git submodule init
git submodule update
```

PR #755, opened 2021-10-21
Changed SetAnnotations method signature

chatton:


chatton:
@rodrigovalin this is so that we can use this to apply to annotations to arbitrary client.Object instances.

PR #763, opened 2021-10-28
Uses latest version of helm-charts.

rodrigovalin:
# Summary

Uses latest helm-chart, where CRDs has been separated into own Chart.


irajdeep (approved):
LGTM

PR #777, opened 2021-11-11
Updates multiple libraries.

rodrigovalin:
# Summary

Updates all libraries as proposed by dependabot.


bznein (approved):
LGTM once evg is happy.

But I really think we should look into using this: https://github.com/hrvey/combine-prs-workflow

rodrigovalin:
@bznein let's discuss about that solution; there are a few requirements:

1. We only want to _merge_ go.mod, but not using `git merge` because some patches update the same packages (merge will not work with conflicts)
2. We need to generate go.sum from whatever go.mod we end up with (by running `go mod download` and `go mod tidy`)
3. Create a new patch with changes
4. I think dependabot will close the tickets once it realizes that the versions have been updated already

PR #782, opened 2021-11-18
Add "WithVolumes" API to podspec template

irajdeep:
This would enable us to patch volumes more cleanly, currently, it would be used to configure volumes when vault is configured or not.

### All Submissions:

### DIFF
--- a/pkg/kube/podtemplatespec/podspec_template.go
+++ b/pkg/kube/podtemplatespec/podspec_template.go
@@ -107,6 +107,22 @@ func WithServiceAccount(serviceAccountName string) Modification {
 	}
 }
 
+// WithVolumes appends the given volumes to the existing volume, it ensures not
+// to override existing volumes
+func WithVolumes(volumes []corev1.Volume) Modification {
+	return func(template *corev1.PodTemplateSpec) {
+		present := make(map[string]struct{})
+		for _, v := range template.Spec.Volumes {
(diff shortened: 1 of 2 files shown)


pkg/kube/podtemplatespec/podspec_template.go
```diff
@@ -107,6 +107,22 @@ func WithServiceAccount(serviceAccountName string) Modification {
 	}
 }
 
+// WithVolumes appends the given volumes to the existing volume, it ensures not
+// to override existing volumes
```
chatton, line 111: Does it make sense to ensure we don't override volumes? I would think it might be desirable to be able to replace volumes with the new ones.

irajdeep: I don't have a strong opinion on this -- I followed what we did for `WithVolume` -- https://github.com/mongodb/mongodb-kubernetes-operator/blob/44c20125f4eb7cff9525e6bfb8259a255c2319c2/pkg/kube/podtemplatespec/podspec_template.go#L127, looks like we skip setting it if it exists. Also users, might configure their own volume, so I am mildly in favour of not overriding.

chatton: SGTM

chatton (approved):
LGTM

PR #785, opened 2021-11-22
CLOUDP-105706: Arbiters go into own Statefulset

rodrigovalin:
# Summary

- Deploys arbiters to a new (different) StatefulSet to be able to scale both arbiters and members independently.
- RFC is here (private): https://github.com/10gen/ops-manager-kubernetes/pull/2064

### DIFF
--- a/api/v1/mongodbcommunity_types.go
+++ b/api/v1/mongodbcommunity_types.go
@@ -497,8 +497,8 @@ func (m MongoDBCommunity) GetScramUsers() []scram.User {
 
 func (m MongoDBCommunity) IsStillScaling() bool {
 	return scale.IsStillScaling(m) || scale.IsStillScaling(automationConfigReplicasScaler{
-		desired: m.DesiredArbiters(),
 		current: m.CurrentArbiters(),
+		desired: m.DesiredArbiters(),
 	})
 }
 
@@ -507,8 +507,8 @@ func (m MongoDBCommunity) IsStillScaling() bool {
 // current number.
 func (m MongoDBCommunity) AutomationConfigMembersThisReconciliation() int {
 	return scale.ReplicasThisReconciliation(automationConfigReplicasScaler{
-		desired: m.Spec.Members,
 		current: m.Status.CurrentMongoDBMembers,
+		desired: m.Spec.Members,
 	})
 }
 
@@ -521,9 +521,11 @@ func (m MongoDBCommunity) AutomationConfigArbitersThisReconciliation() int {
 	if m.AutomationConfigMembersThisReconciliation() < m.Spec.Members {
 		return m.Status.CurrentMongoDBArbiters
 	}
+
 	return scale.ReplicasThisReconciliation(automationConfigReplicasScaler{
-		desired: m.Spec.Arbiters,
-		current: m.Status.CurrentMongoDBArbiters,
+		desired:                m.Spec.Arbiters,
+		current:                m.Status.CurrentMongoDBArbiters,
+		forceIndividualScaling: true,
 	})
 }
 
@@ -628,6 +630,10 @@ func (m MongoDBCommunity) CurrentReplicas() int {
 	return m.Status.CurrentStatefulSetReplicas
 }
 
+func (m MongoDBCommunity) ForcedIndividualScaling() bool {
+	return false
+}
+
 func (m MongoDBCommunity) DesiredArbiters() int {
 	return m.Spec.Arbiters
 }
@@ -649,15 +655,17 @@ func (m MongoDBCommunity) GetMongoDBVersionForAnnotation() string {
 
 func (m *MongoDBCommunity) StatefulSetReplicasThisReconciliation() int {
 	return scale.ReplicasThisReconciliation(automationConfigReplicasScaler{
-		desired: m.DesiredReplicas(),
-		current: m.CurrentReplicas(),
+		desired:                m.DesiredReplicas(),
+		current:                m.CurrentReplicas(),
+		forceIndividualScaling: false,
 	})
 }
 
 func (m *MongoDBCommunity) StatefulSetArbitersThisReconciliation() int {
 	return scale.ReplicasThisReconciliation(automationConfigReplicasScaler{
-		desired: m.DesiredArbiters(),
-		current: m.CurrentArbiters(),
+		desired:                m.DesiredArbiters(),
+		current:                m.CurrentArbiters(),
+		forceIndividualScaling: true,
 	})
 }
 
@@ -698,7 +706,8 @@ func (m MongoDBCommunity) LogsVolumeName() string {
 }
 
 type automationConfigReplicasScaler struct {
-	current, desired int
+	current, desired       int
+	forceIndividualScaling bool
 }
 
 func (a automationConfigReplicasScaler) DesiredReplicas() int {
@@ -707,6 +716,10 @@ func (a automationConfigReplicasScaler) DesiredReplicas() int {
 
 func (a automationConfigReplicasScaler) CurrentReplicas() int {
 	return a.current
+}
+
+func (a automationConfigReplicasScaler) ForcedIndividualScaling() bool {
+	return a.forceIndividualScaling
 }
 
 // +kubebuilder:object:root=true
--- a/api/v1/mongodbcommunity_types.go
+++ b/api/v1/mongodbcommunity_types.go
@@ -391,6 +391,9 @@ type MongoDBCommunityStatus struct {
 	CurrentStatefulSetReplicas int `json:"currentStatefulSetReplicas"`
 	CurrentMongoDBMembers      int `json:"currentMongoDBMembers"`
 
+	CurrentStatefulSetArbitersReplicas int `json:"currentStatefulSetArbitersReplicas"`
+	CurrentMongoDBArbiters             int `json:"currentMongoDBArbiters"`
+
 	Message string `json:"message,omitempty"`
 }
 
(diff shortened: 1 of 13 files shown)


controllers/replica_set_controller.go
```diff
@@ -390,6 +404,14 @@ func (r *ReplicaSetReconciler) ensureService(mdb mdbv1.MongoDBCommunity) error {
 		r.log.Infof("The service already exists... moving forward: %s", err)
 		return nil
 	}
+	svc = buildService(mdb)
```
rodrigovalin, line 407: parameterize `buildService` to be able to build the arbiters service without having to patch it before creation

controllers/replica_set_controller.go
```diff
@@ -407,6 +429,23 @@ func (r *ReplicaSetReconciler) createOrUpdateStatefulSet(mdb mdbv1.MongoDBCommun
 	return nil
 }
 
+func (r *ReplicaSetReconciler) createOrUpdateStatefulSetForArbiters(mdb mdbv1.MongoDBCommunity) error {
```
rodrigovalin, line 432: add function docs

controllers/replica_set_controller.go
```diff
@@ -407,6 +429,23 @@ func (r *ReplicaSetReconciler) createOrUpdateStatefulSet(mdb mdbv1.MongoDBCommun
...
+	set := appsv1.StatefulSet{}
+	name := types.NamespacedName{Name: mdb.Name + "-arb", Namespace: mdb.Namespace}
```
rodrigovalin, line 434: see if we can remove this `+ "-arb"` thing from there

pkg/automationconfig/automation_config_builder.go
```diff
@@ -14,8 +14,9 @@ import (
 type Topology string
 
 const (
-	ReplicaSetTopology Topology = "ReplicaSet"
-	maxVotingMembers   int      = 7
+	ReplicaSetTopology    Topology = "ReplicaSet"
+	maxVotingMembers      int      = 7
```
rodrigovalin, line 18: can't set arbiters to no votes, fix this

api/v1/mongodbcommunity_types.go
```diff
@@ -492,12 +495,37 @@ func (m MongoDBCommunity) GetScramUsers() []scram.User {
 	return users
 }
 
+func (m MongoDBCommunity) IsStillScaling() bool {
```
irajdeep, line 498: Can we add a comment to this function about what it does?

api/v1/mongodbcommunity_types.go
```diff
@@ -492,12 +495,37 @@ func (m MongoDBCommunity) GetScramUsers() []scram.User {
...
 		current: m.Status.CurrentMongoDBMembers,
+		desired: m.Spec.Members,
+	})
+}
+
+// AutomationConfigArbitersThisReconciliation determines the correct number of
+// automation config replica set arbiters based on our desired number, and our
+// current number.
+//
+// Will not update arbiters until members have reached desired number.
```
irajdeep, line 519: > Will not update arbiters until members have reached desired number.

Is this a requirement for the agent to function properly?

api/v1/mongodbcommunity_types.go
```diff
@@ -602,6 +639,18 @@ func (m MongoDBCommunity) CurrentReplicas() int {
 	return m.Status.CurrentStatefulSetReplicas
 }
 
+func (m MongoDBCommunity) ForcedIndividualScaling() bool {
```
irajdeep, line 642: This method always returns false - Is this needed?

api/v1/mongodbcommunity_types.go
rodrigovalin: Yes, better described here: https://github.com/10gen/ops-manager-kubernetes/blob/master/docs/rfcs/0000-supporting-arbiters.md#scaling-of-arbiters

api/v1/mongodbcommunity_types.go
rodrigovalin: For `automationConfigReplicasScaler` this method is different. The meaning is: "with Replica set members, you can Scale 0 -> 3 at once (only at the beginning), with Arbiters you never want that, they always scale one-by-one".

The controller can create a Replica Set with 3 members at once, but after they have been created, the arbiters will be added one by one, because after creation, scaling needs to be done per member.

Let me know if that explains the situation

controllers/replica_set_controller.go
```diff
@@ -407,6 +435,23 @@ func (r *ReplicaSetReconciler) createOrUpdateStatefulSet(mdb mdbv1.MongoDBCommun
 	return nil
 }
 
+func (r *ReplicaSetReconciler) createOrUpdateStatefulSetForArbiters(mdb mdbv1.MongoDBCommunity) error {
```
irajdeep, line 438: Wondering if we can re-use `createOrUpdateStatefulSet` instead of adding a different function for arbiters?

pkg/automationconfig/automation_config_builder.go
```diff
@@ -221,22 +236,36 @@ func (b *Builder) Build() (AutomationConfig, error) {
...
+			//
+			replicaSetIndex = arbitersStartingIndex + processIndex
+		}
+
+		fcv := versions.CalculateFeatureCompatibilityVersion(b.mongodbVersion)
+		if b.fcv != "" {
+			fcv = b.fcv
+		}
 
+		// TODO: make a builder function for this struct
```
irajdeep, line 259: Do we still want to add a builder as a part of this pr?

pkg/automationconfig/automation_config_test.go
```diff
@@ -65,8 +65,9 @@ func TestBuildAutomationConfig(t *testing.T) {
 func TestBuildAutomationConfigArbiters(t *testing.T) {
 	// Test no arbiter (field specified)
 	noArbiters := 0
+	noMembers := 4
 	ac, err := NewBuilder().
-		SetMembers(4).
+		SetMembers(noMembers).
```
irajdeep, line 70: maybe rename from `noMembers` to `numMembers`?

pkg/automationconfig/automation_config_test.go
```diff
@@ -91,78 +92,71 @@ func TestBuildAutomationConfigArbiters(t *testing.T) {
...
 		SetArbiters(noArbiters).
 		Build()
 
 	assert.NoError(t, err)
 
 	rs = ac.ReplicaSets[0]
 	for i, member := range rs.Members {
-		if i < noArbiters {
-			assert.True(t, member.ArbiterOnly, "The first two members should be arbiters")
+		if i < noMembers {
```
irajdeep, line 123: same as the above comment regarding rename.

irajdeep (commented):
Looks good, added a few comments, lmk what you think.

api/v1/mongodbcommunity_types.go
irajdeep: sounds good.

api/v1/mongodbcommunity_types.go
irajdeep: sounds good, maybe we can add it as a comment in the function for posterity.

controllers/replica_set_controller.go
```diff
@@ -384,12 +402,22 @@ func (r *ReplicaSetReconciler) deployMongoDBReplicaSet(mdb mdbv1.MongoDBCommunit
...
+
+	if mdb.Spec.Arbiters > 0 {
+		svc = buildService(mdb, true)
+		err = r.client.Create(context.TODO(), &svc)
+		if err != nil && apiErrors.IsAlreadyExists(err) {
+			r.log.Infof("Arbiters service already exists... moving forward: %s", err)
+			return nil
+		}
+	}
+
```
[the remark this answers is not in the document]
chatton: I will throw one more idea into the mix with the goal of keeping things as simple as possible

```golang
services := []corev1.Service{mdbSvc}
if mdb.Spec.Arbiters > 0 {
    services = append(services, buildService(mdb, true))
}
for _, svc := range services {
    // create them
}

```

controllers/replica_set_controller.go
```diff
@@ -448,13 +501,24 @@ func buildAutomationConfig(mdb mdbv1.MongoDBCommunity, auth automationconfig.Aut
...
+// If `isArbiter` is true, the function will create a Service suitable for the
+// Arbiter's StatefulSet.
+func buildService(mdb mdbv1.MongoDBCommunity, isArbiter bool) corev1.Service {
 	label := make(map[string]string)
-	label["app"] = mdb.ServiceName()
+	if isArbiter {
+		label["app"] = mdb.ArbiterServiceName()
+	} else {
+		label["app"] = mdb.ServiceName()
+	}
```
chatton, line 513: [nit] to keep things as flat as possible we could do

```golang
name := mdb.ServiceName()
if isArbiter {
    name = mdb.ArbiterServiceName()
}
label["app"] = name

```

will leave it to you! This is just my personal preference.

controllers/replica_set_controller.go
```diff
@@ -561,6 +630,14 @@ func buildStatefulSetModificationFunction(mdb mdbv1.MongoDBCommunity) statefulse
 	)
 }
 
+func buildArbitersModificationFunction(mdb mdbv1.MongoDBCommunity) statefulset.Modification {
+	return statefulset.Apply(
+		statefulset.WithReplicas(int(mdb.Spec.Arbiters)),
+		statefulset.WithServiceName(mdb.Name+"-arb-svc"),
```
chatton, line 636: we have a function for the arbiter service name right?

controllers/replica_set_controller.go
```diff
@@ -561,6 +630,14 @@ func buildStatefulSetModificationFunction(mdb mdbv1.MongoDBCommunity) statefulse
 	)
 }
 
+func buildArbitersModificationFunction(mdb mdbv1.MongoDBCommunity) statefulset.Modification {
+	return statefulset.Apply(
+		statefulset.WithReplicas(int(mdb.Spec.Arbiters)),
```
chatton, line 635: this should be using the function which takes scaling into account right? Rather than spec.Arbiters directly.

pkg/automationconfig/automation_config.go
```diff
@@ -135,17 +135,16 @@ type ReplicaSetMember struct {
...
-	votes := 1
-	priority := 1
+	votes := 0
+	priority := 0
 
-	isArbiter := totalVotesSoFar < numberArbiters
-
-	if totalVotesSoFar > maxVotingMembers {
-		votes = 0
-		priority = 0
```
chatton, line 148: is this logic lost?

edit: NVM I see it is handled elsewhere!

test/e2e/mongodbtests/mongodbtests.go
```diff
@@ -42,6 +42,17 @@ func StatefulSetBecomesReady(mdb *mdbv1.MongoDBCommunity, opts ...wait.Configura
 	return statefulSetIsReady(mdb, defaultOpts...)
 }
 
+// StatefulSetBecomesReady ensures that the underlying stateful set
```
chatton, line 45: incorrect function name in comment

chatton (commented):
Looking great, I just had a few questions.

Also, we should make sure to update the release notes with this change.

pkg/automationconfig/automation_config_builder.go
rodrigovalin: Not as part of this PR. If you see in this function we have 2 ways of building 2 structs:

1. `Process` is built manually and then with some getters (SetPort, SetStoragePath)
2. `ReplicaSetMember` is built with a `newReplicaSetMember` functin, with parameters
3. The *standard* we have been doing is to use a Builder (like Secret.Builder)

The TODO is about the combination of strategies in this function; will add more context to the TODO in fact 👍

controllers/replica_set_controller.go
rodrigovalin: thanks for catching this!

rodrigovalin:
@chatton I moved the buildService logic around and all the changes are in [this commit](https://github.com/mongodb/mongodb-kubernetes-operator/pull/785/commits/e1e0d399de59b7a28e7bfc06cfff16d73337a04f). Let me know what you think.

rodrigovalin:
@irajdeep @chatton addressed all comments and rebased on master. Ready for another review.

irajdeep (approved):
LGTM 👍

chatton (approved):
LGTM! Nice work with this implementation.

PR #798, opened 2021-12-01
Change interface for automation config

bznein:
This PR includes changes to make the automation config package and the construction usable by enterprise with the new vault changes

### DIFF
--- a/controllers/construct/mongodbstatefulset.go
+++ b/controllers/construct/mongodbstatefulset.go
@@ -89,6 +89,9 @@ type MongoDBStatefulSetOwner interface {
 
 	// GetMongodConfiguration returns the MongoDB configuration for each member.
 	GetMongodConfiguration() map[string]interface{}
+
+	// NeedsAutomationConfigVolume returns whether the statefuslet needs to have a volume for the automationconfig.
+	NeedsAutomationConfigVolume() bool
 }
 
 // BuildMongoDBReplicaSetStatefulSetModificationFunction builds the parts of the replica set that are common between every resource that implements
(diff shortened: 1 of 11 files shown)


controllers/construct/mongodbstatefulset.go
```diff
@@ -89,6 +89,9 @@ type MongoDBStatefulSetOwner interface {
 
 	// GetMongodConfiguration returns the MongoDB configuration for each member.
 	GetMongodConfiguration() map[string]interface{}
+
+	// NeedsAutomationConfigVolume returns whether the statefuslet needs to have a volume for the automationconfig.
+	NeedsAutomationConfigVolume() bool
```
bznein, line 94: Community always returns `true`, but on AppDB it depends whether or not we are using vault

controllers/construct/mongodbstatefulset.go
```diff
@@ -180,11 +188,11 @@ func BuildMongoDBReplicaSetStatefulSetModificationFunction(mdb MongoDBStatefulSe
 }
 
 func BaseAgentCommand() string {
-	return "agent/mongodb-agent -cluster=" + clusterFilePath + " -healthCheckFilePath=" + agentHealthStatusFilePathValue + " -serveStatusPort=5000"
+	return "agent/mongodb-agent -healthCheckFilePath=" + agentHealthStatusFilePathValue + " -serveStatusPort=5000"
```
bznein, line 191: We need to extract this from the BaseCommand (reason will be explained in an enterprise PR)

pkg/kube/secret/secret.go
```diff
@@ -13,6 +13,7 @@ import (
 )
 
 type Getter interface {
+	IsNotFound(err error) bool
```
bznein, line 16: I decided to add this as method for the interface. This way community will just wrap it with `apiErrors.IsNotFound()` while enterprise will add additional logic based on message returned by the vault secret

pkg/readiness/headless/headless.go
```diff
@@ -15,9 +19,26 @@ import (
...
+
+		if err != nil {
+			return false, err
+		}
+
+		targetVersion, err = strconv.ParseInt(string(data), 10, 64)
+
+		if err != nil {
+			return false, err
+		}
```
bznein, line 41: If there is no secret (e.g AppDB Vault), it means we are mounting it as a configmap and we check from there

pkg/kube/client/client.go
```diff
@@ -103,6 +104,10 @@ func (c client) GetSecret(objectKey k8sClient.ObjectKey) (corev1.Secret, error)
 	return s, nil
 }
 
+func (c client) IsNotFound(err error) bool {
```
irajdeep, line 107: Q: Why is this wrapping needed?

pkg/kube/secret/secret.go
irajdeep: ah, I see -- I already added it to master. do you think we still need this?

pkg/kube/client/client.go
bznein: it is not exactly needed, but it allows us to pass a secret client which implements the same without having to duplicate the custom `NotFound` code in a lot of places

irajdeep: hmm, I merged a change that checks for secretNotFound -- basically the same logic as enterprise -- since we need it for scram secret creation. Do you think we can re-use it?

irajdeep (approved):
Nice job getting this working -- added one non blocking comment

pkg/kube/client/client.go
bznein: sure, I'll do it

PR #806, opened 2021-12-13
Cloudp 107436 enable partial merging of custom automation config

chatton:
This PR adds scaffolding for applying an automation config override. In this PR, the only field that gets merged is the process.disabled field. This will enable [CLOUDP-68868](https://jira.mongodb.org/browse/CLOUDP-68868) to be resolved.

Sample CR:

~~apiVersion: mongodbcommunity.mongodb.com/v1
kind: MongoDBCommunity
metadata:
  name: example-mongodb
spec:
  members: 3
  automationConfigOverride:
    automationConfig:
      processes:
        - name: example-mongodb-1
          disabled: true~~

```yaml
---
kind: MongoDBCommunity
metadata:
  name: example-mongodb
spec:
  members: 3
  automationConfig:
      processes:
        - name: example-mongodb-1
          disabled: true
```

This will disable the process with name `example-mongodb-1`.

In the future we can handle merging of all automation config fields.

I am not including documentation changes as I believe we should include documentation only when we support all fields.


api/v1/mongodbcommunity_types.go
```diff
@@ -197,6 +201,31 @@ type AuthenticationRestriction struct {
...
+// UnmarshalJSON will decode the data into the wrapped map
+func (m *AutomationConfigWrapper) UnmarshalJSON(data []byte) error {
+	return json.Unmarshal(data, &m.AutomationConfig)
+}
+
+func (m *AutomationConfigWrapper) DeepCopy() *AutomationConfigWrapper {
+	return &AutomationConfigWrapper{
+		AutomationConfig: m.AutomationConfig,
+	}
+}
```
chatton, line 227: this is the same boiler plate required by sts override.

config/samples/mongodb.com_v1_mongodbcommunity_disabled_process_cr.yaml
```diff
@@ -0,0 +1,39 @@
+---
+apiVersion: mongodbcommunity.mongodb.com/v1
+kind: MongoDBCommunity
+metadata:
+  name: example-mongodb
+spec:
+  members: 3
+  automationConfigOverride:
```
rodrigovalin, line 8: why do we have `automationConfigOverride.automationConfig`, a 2-level attribute? are you planning on adding any kind of siblings to `automationConfig` attribute in the future?

api/v1/mongodbcommunity_types.go
```diff
@@ -101,6 +101,10 @@ type MongoDBCommunitySpec struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
 	// +nullable
 	AdditionalMongodConfig MongodConfiguration `json:"additionalMongodConfig,omitempty"`
+
+	// AutomationConfigOverride is merged on top of the operator created automation config. Processes are merged
+	// by name.
+	AutomationConfigOverride *AutomationConfigOverride `json:"automationConfigOverride,omitempty"`
```
rodrigovalin, line 107: Should we indicate here that only `process.Disabled` is supported right now?

pkg/util/merge/merge_automationconfigs_test.go
```diff
@@ -0,0 +1,38 @@
+package merge
+
+import (
+	"github.com/mongodb/mongodb-kubernetes-operator/pkg/automationconfig"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestMergeAutomationConfigs(t *testing.T) {
```
rodrigovalin, line 9: I think we could add a test where the override changes a process that does not exist in the original automation config. This can be achieved by an automation config override with a different name, or with more elements, where the ones with modifications are over the len() of the original automationconfig.

rodrigovalin (approved):
A few comments that might be relevant, but not blockers.

config/samples/mongodb.com_v1_mongodbcommunity_disabled_process_cr.yaml
chatton: this was because of what I thought was a limitation, but I think actually we can keep it at 1 level depth. I will experiment with it.

api/v1/mongodbcommunity_types.go
chatton: added comment

PR #813, opened 2021-12-14
CLOUDP-104426: multiple users issue

bznein:
Arising from #764: adding some validation to make sure that `scramCredentialsSecretName` is unique among users


chatton (approved):
LGTM!

PR #816, opened 2021-12-14
CLOUDP-105593: Agent image to ubi8-minimal

bznein:
Updates the agent image to use ubi8-minimal instead of ubi7.

### DIFF
--- a/scripts/dev/templates/agent/Dockerfile.ubi
+++ b/scripts/dev/templates/agent/Dockerfile.ubi
@@ -8,5 +8,5 @@ RUN microdnf install -y  --disableplugin=subscription-manager curl \
     && microdnf upgrade -y  \
     && rm -rf /var/lib/apt/lists/*
 
-RUN microdnf remove perl-IO-Socket-SSL procps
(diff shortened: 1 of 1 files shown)


scripts/dev/templates/agent/Dockerfile.ubi
```diff
@@ -1,10 +1,12 @@
...
 {% block packages -%}
-RUN yum install -y  --disableplugin=subscription-manager -q curl \
-    hostname nss_wrapper --exclude perl-IO-Socket-SSL procps \
-    && yum upgrade -y -q \
+RUN microdnf install -y  --disableplugin=subscription-manager curl \
+    hostname nss_wrapper tar gzip procps\
+    && microdnf upgrade -y  \
     && rm -rf /var/lib/apt/lists/*
+
+RUN microdnf remove perl-IO-Socket-SSL
```
bznein, line 11: We do the same in other places: `microdnf` doesn't have an `exclude` option, so we install and then remove

chatton (approved):
LGTM! Nice work figuring this one out!

Issue #839, opened 2022-01-05
Unable to change port


vikomte:
I have tried every configuration examples from https://docs.mongodb.com/kubernetes-operator/master/reference/k8s-operator-specification/ but no matter what i do, the port of the running ReplicaSet still is 27017.

I have tried adding spec.additionalMongodConfig.net.port, spec.net.port, spec.configSrv.net.port, spec.configSrv.additionalMongodConfig.net.port, and many more..

Example:
```
apiVersion: mongodbcommunity.mongodb.com/v1
kind: MongoDBCommunity
metadata:
  name: mongodb-test
  namespace: mongodb
  labels:
    app: mongodb-test

spec:
  members: 2
  type: ReplicaSet
  version: "4.2.6"
  security:
    authentication:
      modes: ["SCRAM"]
  additionalMongodConfig:
    net:
      port: 44852
...
```

I understood that i can't change the default service port of 27017 which is not a problem because i can overide it.

It is a problem to be unable to change the mongodb port thought.

I have searched hours to try getting this working.

qrr1995:
The doc you have referred is for `MongoDB Enterprise Kubernetes Operator`  https://github.com/mongodb/mongodb-enterprise-kubernetes ， maybe it didn't work for the MongoDB community operator?

PR #848, opened 2022-01-12
Make mongod port configurable

mircea-cosbuc:
This addresses #839 and moves `mongod` configurations handled by the operator to the `MongodConfiguration` type.

### All Submissions:

### DIFF
--- a/api/v1/mongodbcommunity_types.go
+++ b/api/v1/mongodbcommunity_types.go
@@ -295,7 +295,7 @@ func (m MongodConfiguration) GetDBDataDir() string {
 }
 
 // GetDBPort returns the port that should be used for the mongod process.
-// If not port is specified, the default port of 27017 will be used.
+// If port is not specified, the default port of 27017 will be used.
 func (m MongodConfiguration) GetDBPort() int {
 	// When passed as a number "net.port" is unmarshalled into a float64
 	return int(objx.New(m.Object).Get("net.port").Float64(defaultDBPort))
--- a/api/v1/mongodbcommunity_types.go
+++ b/api/v1/mongodbcommunity_types.go
@@ -29,6 +29,8 @@ type Type string
 const (
 	ReplicaSet       Type   = "ReplicaSet"
 	defaultDBForUser string = "admin"
+	defaultDataDir          = "/data"
+	defaultDBPort           = 27017
 )
 
 type Phase string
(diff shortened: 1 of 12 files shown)


pkg/automationconfig/automation_config_builder.go
```diff
@@ -17,6 +17,7 @@ const (
 	ReplicaSetTopology    Topology = "ReplicaSet"
 	maxVotingMembers      int      = 7
 	arbitersStartingIndex int      = 100
+	defaultPort           int      = 27017
```
irajdeep, line 20: The `defaultPort` is also declared at `mongodbcommunity_types.go`, is it possible to consolidate it to only one place?

api/v1/mongodbcommunity_types.go
```diff
@@ -276,6 +278,29 @@ func (m *MongodConfiguration) DeepCopy() *MongodConfiguration {
...
+	return m
+}
+
+// GetDBDataDir returns the db path which should be used.
+func (m MongodConfiguration) GetDBDataDir() string {
+	return objx.New(m.Object).Get("storage.dbPath").Str(defaultDataDir)
+}
+
+// GetDBPort returns the port that should be used for the mongod process.
+// If not port is specified, the default port of 27017 will be used.
```
irajdeep, line 298:
```suggestion
// If port is not specified, the default port of 27017 will be used.
```

controllers/construct/build_statefulset_test.go
```diff
@@ -78,25 +78,6 @@ func TestMongod_Container(t *testing.T) {
 	})
 }
 
-func TestGetDbPath(t *testing.T) {
```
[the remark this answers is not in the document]
mircea-cosbuc: Yes, that test covers it. We could break that test into multiple ones if we have more `mongod` configuration options we want to handle.

pkg/automationconfig/automation_config_builder.go
irajdeep: We have been trying to avoid using `util` package for a while now, basically, we are moving out of the util package in enterprise since it's very unclear what those collections of packages entail. 

I still this this should be declared at one place to avoid any potential bug later, but will leave it to you to decided the preferred location.

irajdeep: Just noticed we already import `automationconfig` package in `mongodbcommunity_types.go`, so maybe we can reuse this constant in mongodbcommunity_types.go ?

controllers/construct/mongodbstatefulset.go
```diff
@@ -88,7 +86,7 @@ type MongoDBStatefulSetOwner interface {
 	LogsVolumeName() string
 
 	// GetMongodConfiguration returns the MongoDB configuration for each member.
-	GetMongodConfiguration() map[string]interface{}
+	GetMongodConfiguration() mdbv1.MongodConfiguration
```
chatton, line 89: this interface is used in enterprise (for deploying the AppDB). I think this will still be fine, but next time we update the community operator version library in the enterprise repo we can address this.

chatton (approved):
Reviewed together in person, LGTM! 

Thanks for moving some of those. values on the `MongodConfiguration`, I think that makes a lot more sense.

I think this is good to merge once @irajdeep's comments are resolved.

irajdeep (approved):
LGTM 👍

controllers/construct/mongodbstatefulset.go
mircea-cosbuc: It might also be a good reason to synchronise the way we're handling the `mongod` options in the two operators.

PR #854, opened 2022-01-17
Remove Unused Github Action

chatton:
### All Submissions:

This PR removes a github action that is no longer required (this was failing on forks but is not an actual test.)


irajdeep:
@chatton the e2e tests are failing -- unsure if related

Issue #818, opened 2021-12-14
Operator is appending args


xadereq:
**What did you do to encounter the bug?**
1. Create mongo deployment with custom container in template
2. Set some args in custom container
3. Apply manifest
4. Change previously args for container for other value
5. Operator is appending args list in custom container, not overwriting it. I had to change stateful set manually to get rid of arguments applied previously.

**What did you expect?**
Arguments should be overwrited instead of being appended

**What happened instead?**
as above

**Operator Information**
 - Operator Version: latest (0.7.2)
 - MongoDB Image used (5.0.5)

**Kubernetes Cluster Information**
 - Distribution: kubeadm
 - Version: 1.21.1
 - Image Registry location (quay, or an internal registry): quay

PR #862, opened 2022-01-20
CLOUDP-107815: Do not append stale command/args to containers

irajdeep:
### Summary:

Fixes:  https://github.com/mongodb/mongodb-kubernetes-operator/issues/818
This patch prevents the operator to merge command/args in containers.  It uses the one specified in the CR as the source of truth(if specified). The problem with the merging is highlighted in the linked issue. If a user removes a command/arg from a container the operator since it is currently merging the new commands/args with the old ones, it is keeping the removed args lurking around as well.

### All Submissions:

### DIFF
--- a/pkg/util/merge/merge_test.go
+++ b/pkg/util/merge/merge_test.go
@@ -182,11 +182,11 @@ func TestMergeContainer(t *testing.T) {
 		)
 		mergedContainer := Container(defaultContainer, overrideContainer)
 		assert.Equal(t, overrideContainer.Name, mergedContainer.Name, "Name was overridden, and should be used.")
-		assert.Equal(t, []string{"a", "b", "c", "d", "f", "e"}, mergedContainer.Command, "Command was specified in both, so results should be merged.")
+		assert.Equal(t, []string{"d", "f", "e"}, mergedContainer.Command, "Command was specified in both, so results should be merged.")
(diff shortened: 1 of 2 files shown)


chatton (approved):
Changes look good to me, just to double check, this same logic happens for init containers and containers right?

If this is the case then LGTM!

irajdeep:
> this same logic happens for init containers and containers right?

@chatton that is right.

pkg/util/merge/merge_test.go
```diff
@@ -182,11 +182,11 @@ func TestMergeContainer(t *testing.T) {
 		)
 		mergedContainer := Container(defaultContainer, overrideContainer)
 		assert.Equal(t, overrideContainer.Name, mergedContainer.Name, "Name was overridden, and should be used.")
-		assert.Equal(t, []string{"a", "b", "c", "d", "f", "e"}, mergedContainer.Command, "Command was specified in both, so results should be merged.")
+		assert.Equal(t, []string{"d", "f", "e"}, mergedContainer.Command, "Command was specified in both, so results should be merged.")
```
mircea-cosbuc, line 185: Think we should change the comment as well to illustrate that we'd be overriding the default command.

mircea-cosbuc (approved):
LGTM with a small change for clarity.

pkg/util/merge/merge_test.go
irajdeep: Good catch, fixed.

PR #880, opened 2022-02-04
Fix Secret creation with username with rfc-1123 non-allowed-chars.

### LINKED ISSUE #879: Credentials secret creation fails when user name contains underscore
I defined a user name with _underscore_ in it ("_").

**What did you expect?**
That the credentials `Secret` would have been created with dash ("-") instead of underscore ("_") characters. 

**What happened instead?**
I defined a user name with _underscore_ in it. The credentials `Secret` tried to be created will have `metadata.name` with characters not replaced and with a value which is not conform to the RFC 1123. 

**Logs**
```
controllers/replica_set_controller.go:236     Could not update connection string secrets: Secret "myservice-mongodb5-admin-prometheus_exporter" is invalid: metadata.name: Invalid value: "myservice-mongodb5-admin-prometheus_exporter": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*') 
```

**Operator Information**
 - Operator version: 0.7.2
 - Operator image registry: Quay
 - MongoDB Image used: docker.io/mongo:5.0.4

 - Distribution: none
 - Version:  v1.21.4

**Additional context**
MongoDB manifest snippet:

```
kind: MongoDBCommunity
metadata:
  name: myservice-mongodb5
spec:
  [...]
  users:
    name: prometheus_exporter
    passwordSecretRef:
      name: myservice-mongodb5-prometheus-exporter-user-password
    roles:
    - db: admin
      name: clusterMonitor
    - db: local
      name: read
    scramCredentialsSecretName: myservice-mongodb5-prometheus-exporter-user-scram
  version: 5.0.4
```

rodrigovalin:
- Fixes #879 

Username with underscore in name will prevent Secret from being created.

Open questions:

- Should we add other characters to be replaced as well?

### DIFF
--- a/pkg/authentication/scram/scram.go
+++ b/pkg/authentication/scram/scram.go
@@ -3,6 +3,7 @@ package scram
 import (
 	"encoding/base64"
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 
@@ -339,5 +340,6 @@ func convertMongoDBUserToAutomationConfigUser(secretGetUpdateCreateDeleter secre
 
 // GetConnectionStringSecretName returns the name of the secret where the operator stores the connection string for current user
 func (u User) GetConnectionStringSecretName(mdb Configurable) string {
-	return fmt.Sprintf("%s-%s-%s", mdb.NamespacedName().Name, u.Database, u.Username)
+	username := strings.ReplaceAll(u.Username, "_", "-")
+	return fmt.Sprintf("%s-%s-%s", mdb.NamespacedName().Name, u.Database, username)
 }
--- a/pkg/authentication/scram/scram.go
+++ b/pkg/authentication/scram/scram.go
@@ -3,6 +3,7 @@ package scram
 import (
 	"encoding/base64"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


irajdeep:
We should try to make any non-compatible `RFC 1123` compatible, in addition to this(underscore) special case. We have had a discussion around this, in this ticket: https://jira.mongodb.org/browse/CLOUDP-95810

rodrigovalin:
I think we should be good with transforming `: / ? # [ ] @` into `-` as described in https://docs.mongodb.com/manual/reference/connection-string/. Do you agree that that solution is sufficient?

irajdeep:
> I think we should be good with transforming `: / ? # [ ] @` into `-` as described in https://docs.mongodb.com/manual/reference/connection-string/. Do you agree that that solution is sufficient?

I think this would make sense. In addition to this, we should deal it the way we handle `scramCredentialsSecretName` which has the same problem:
1. Ask the user to provide an optional secret ref name for storing the credential.
2. Perform a DNS1123 name validation at the custom resource level if it is specified.

rodrigovalin:
Considering that we have an internal ticket for the bigger fix, we could fix this problem with this preliminary solution right away; and then work on the other implementation once that ticket has been triagged.

irajdeep:
Yeah, for now, we can transform  `: / ? # [ ]` @ into `-` and get this merged and revisit the optional secretRef later.

rodrigovalin:
@irajdeep ready for another review

irajdeep (approved):
LGTM

PR #888, opened 2022-02-09
Accepts `metrics.prometheus` endpoint.

rodrigovalin:
# Summary 

Using latest agent, we can pass a `spec.metrics.prometheus` configuration that will result in an automation_config change that *enables a `/metrics` endpoint that can be used with Prometheus.*

# Dependencies

We need to publish a new agent version before these changes are even recognized by the agents.

### DIFF
--- a/api/v1/mongodbcommunity_types.go
+++ b/api/v1/mongodbcommunity_types.go
@@ -107,9 +107,9 @@ type MongoDBCommunitySpec struct {
 	// by name. Currently Only the process.disabled field is supported.
 	AutomationConfigOverride *AutomationConfigOverride `json:"automationConfig,omitempty"`
 
-	// Metrics configurations.
+	// Prometheus configurations.
 	// +optional
-	Metrics *Metrics `json:"metrics,omitempty"`
+	Prometheus *Prometheus `json:"prometheus,omitempty"`
 }
 
 // ReplicaSetHorizonConfiguration holds the split horizon DNS settings for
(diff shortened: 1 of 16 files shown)


priyolahiri (approved):
LGTM

api/v1/mongodbcommunity_types.go
```diff
@@ -128,6 +132,34 @@ type CustomRole struct {
 	AuthenticationRestrictions []AuthenticationRestriction `json:"authenticationRestrictions,omitempty"`
 }
 
+type Metrics struct {
```
irajdeep, line 135: Q: Any reason to wrap this in Metrics struct instead of using Prometheus struct directly?

pkg/automationconfig/automation_config.go
```diff
@@ -185,6 +186,31 @@ type Auth struct {
...
+	ServiceDiscovery string `json:"serviceDiscovery"`
+}
+
+func NewDefaultPrometheus(username string) Prometheus {
+	return Prometheus{
+		Enabled:          true,
+		Username:         username,
+		Scheme:           "http",
+		Mode:             "opsManager",
+		ListenAddress:    "0.0.0.0:9216",
```
irajdeep, line 208: Should we hardcode the port here? can it be changed on some usecase?

irajdeep (approved):
LGTM -- added some minor comments.

api/v1/mongodbcommunity_types.go
rodrigovalin: No reason in particular, maybe preparing for other options in the future, to go besides Prometheus; but I think that posibility is unlikely.

I don't have a strong opinion. Maybe it is unecessary at this point. I will remove

pkg/automationconfig/automation_config.go
rodrigovalin: It can be changed via `prometheus.port`

Issue #874, opened 2022-02-02
TLS certificate content update fails to rollout automation config update


maxgio92:
**What did you do to encounter the bug?**
Steps to reproduce the chicken-and-egg behaviour:
- Update a certificate TLS (e.g. adding a new SAN); the `Secret` will be named the same, the content is changed.
- mongod fails to find the certificate file, as the `Secret` generated by the operator has a key named as the hashed cert PEM file and that hash is changed as content did. Insted, the configuration is not already updated with new x509 certificate file path.
- The configuration won't be updated until mongod (and the mongo-agent) is ready ([as for](https://github.com/mongodb/mongodb-kubernetes-operator/blob/master/controllers/replica_set_controller.go#L377))

**What did you expect?**
An update of the automation config, before the StatefulSet rollout happens (as like [this](https://github.com/mongodb/mongodb-kubernetes-operator/blob/master/controllers/replica_set_controller.go#L388)), otherwise I need to manually fix the configuration with right filepaths (`Secret`'s hashed certificate filename key).

**What happened instead?**
The automation configuration is not updated until the StatefulSet is ready, which won't until the automation configuration will be.

**Operator Information**
 - Operator version: 0.7.2
 - Operator image registry: Quay
 - MongoDB Image used: docker.io/mongo:5.0.4

**Kubernetes Cluster Information**
 - Distribution: none
 - Version:  v1.21.4

**Additional context**

**`MongoDBCommunity` manifest**
```
apiVersion: mongodbcommunity.mongodb.com/v1
kind: MongoDBCommunity
metadata:
  name: my-service-mongodb5
spec:
  members: 4
  replicaSetHorizons:
  <my_horizon_DNS_records>
  security:
    authentication:
      modes:
      - SCRAM
    tls:
      caConfigMapRef:
        name: my-service-mongodb5-ca
      certificateKeySecretRef:
        name: my-service-mongodb5-tls
      enabled: true
  statefulSet:
    spec:
      template:
        metadata:
          labels:
            app.kubernetes.io/instance: my-service-mongodb5
          topologySpreadConstraints:
          - labelSelector:
              matchLabels:
                app.kubernetes.io/instance: my-service-mongodb5
            maxSkew: 1
            topologyKey: failure-domain.beta.kubernetes.io/zone
            whenUnsatisfiable: DoNotSchedule
  type: ReplicaSet
  users:
    <my_users>
  version: 5.0.4
```

[...]

PR #905, opened 2022-02-24
CLOUDP-111968: Update automation config when updating TLS

mircea-cosbuc:
### All Submissions:

This might address issues like https://github.com/mongodb/mongodb-kubernetes-operator/issues/874 and similar, when the mongod process is trying to start with a stale automation config.

### DIFF
--- a/config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
+++ b/config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
@@ -5,12 +5,6 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.4.1
-    service.binding/type: 'mongodb'
-    service.binding/provider: 'community'
-    service.binding: 'path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret'
-    service.binding/connectionString: 'path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret,sourceKey=connectionString.standardSrv'
-    service.binding/username: 'path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret,sourceKey=username'
-    service.binding/password: 'path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret,sourceKey=password'
   creationTimestamp: null
   name: mongodbcommunity.mongodbcommunity.mongodb.com
 spec:
--- a/config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
+++ b/config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
@@ -5,6 +5,12 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.4.1
+    service.binding/type: 'mongodb'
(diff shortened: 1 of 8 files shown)


config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
```diff
@@ -5,12 +5,6 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.4.1
-    service.binding/type: 'mongodb'
-    service.binding/provider: 'community'
-    service.binding: 'path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret'
-    service.binding/connectionString: 'path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret,sourceKey=connectionString.standardSrv'
-    service.binding/username: 'path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret,sourceKey=username'
-    service.binding/password: 'path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret,sourceKey=password'
```
irajdeep, line 13: Why did this get removed, it was added in a recent PR to support service-binding operator I think.

test/e2e/replica_set_tls_rotate_delete_sts/replica_set_tls_rotate_delete_sts_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	t.Run("Create MongoDB Resource", mongodbtests.CreateMongoDBResource(&mdb, ctx))
+	t.Run("Basic tests", mongodbtests.BasicFunctionality(&mdb))
+	t.Run("Wait for TLS to be enabled", tester.HasTlsMode("requireSSL", 60, WithTls(mdb)))
+	t.Run("Test Basic TLS Connectivity", tester.ConnectivitySucceeds(WithTls(mdb)))
+	t.Run("Ensure Authentication", tester.EnsureAuthenticationIsConfigured(3, WithTls(mdb)))
+	t.Run("Test TLS required", tester.ConnectivityFails(WithoutTls()))
+
+	t.Run("MongoDB is reachable while certificate is rotated", func(t *testing.T) {
+		t.Run("Delete Statefulset", mongodbtests.DeleteStatefulSet(&mdb))
```
irajdeep, line 58: So the way this test is different from the existing rotate cert test is that we delete the statefulset while rotating the certificate right?

controllers/replica_set_controller.go
```diff
@@ -301,6 +301,10 @@ func (r *ReplicaSetReconciler) ensureTLSResources(mdb mdbv1.MongoDBCommunity) er
 		if err := ensureTLSSecret(r.client, mdb); err != nil {
 			return errors.Errorf("could not ensure TLS secret: %s", err)
 		}
+		r.log.Infof("Ensuring the automation config is up to date with the latest TLS secret")
+		if _, err := r.ensureAutomationConfig(mdb); err != nil {
+			return errors.Errorf("failed to sync automation config with latest TLS secret")
+		}
```
irajdeep, line 307: Do we need this call explicitly now? Q: How was the AC updated when the cert is rotated currently(in the existing cert rotation test).

mircea-cosbuc: If the pods are running already it's not a problem, because they get restarted by the reconcile loop after the automation config is updated.

irajdeep (approved):
LGTM

PR #923, opened 2022-03-15
CLOUDP-95810: Normalize generated connection string secret name

mircea-cosbuc:
### All Submissions:
This patch adds a normalisation step to the connection string secret generated by the operator and also introduces an optional field for deterministically naming this secret through the `users[i].connectionStringSecretName` field.


api/v1/mongodbcommunity_types.go
```diff
@@ -373,6 +379,15 @@ func (m MongoDBUser) GetScramCredentialsSecretName() string {
 	return fmt.Sprintf("%s-%s", m.ScramCredentialsSecretName, "scram-credentials")
 }
 
+// GetConnectionStringSecretName gets the connection string secret name provided by the user or generated
+// from the SCRAM user configuration.
+func (m MongoDBUser) GetConnectionStringSecretName(resourceName string) string {
```
rodrigovalin, line 384: Isn't `resourceName` available in the receiving `MongoDBUser`? Can you skip the `resourceName` parameter?

controllers/mongodb_users.go
```diff
@@ -44,9 +44,20 @@ func (r ReplicaSetReconciler) updateConnectionStringSecrets(mdb mdbv1.MongoDBCom
 		if err != nil {
 			return err
 		}
+		secretName := user.ConnectionStringSecretName
```
rodrigovalin, line 47: I think this should be `GetConnectionStringSecretName()` instead of you'll not be respecting the custom name of the secret?

rodrigovalin (commented):
I'm not sure why is `GetConnectionStringSecretName()` not used consistently across the whole patch. Mostly it is `ConnectionStringSecretName` which, being optional, can be left blank, in which case the default should be used instead.

api/v1/mongodbcommunity_types.go
mircea-cosbuc: It's not available in the `MongoDBUser` indeed.

controllers/mongodb_users.go
mircea-cosbuc: The controllers are working with `scram.User` objects which are created with https://github.com/mongodb/mongodb-kubernetes-operator/blob/CLOUDP-95810_normalize-secret-name/api/v1/mongodbcommunity_types.go#L578 . At this point `user.ConnectionStringSecretName` contains the correct secret name (configured or generated).

rodrigovalin (approved):
LGTM

pkg/util/names/names.go
```diff
@@ -0,0 +1,32 @@
+package names
+
+import (
+	"regexp"
+	"strings"
+
```
irajdeep, line 6: Preferably we can create this `name` package outside `util`. At some point we should remove packages from inside `util` -- the way we currently use is an anti-pattern.

irajdeep (approved):
LGTM

pkg/util/names/names.go
mircea-cosbuc: It might make sense to just have this in [mongodbcommunity_types.go](https://github.com/mongodb/mongodb-kubernetes-operator/blob/master/api/v1/mongodbcommunity_types.go) in this case.

mircea-cosbuc: No, it would still be a utility function, but its only purpose right now is to normalise this particular secret name and it is called when generating `scram.Users` from a `MongoDBCommunity` resource.

irajdeep: got it, makes sense to move there then.

PR #995, opened 2022-06-07
CLOUDP-119891: Change port of running cluster

### LINKED ISSUE #945: Modify port error
1. Deploy the operator and cluster as normal
2. Modify the port of the configuration file from the original default 27017 to 27018

**What did you expect?**
The cluster state is running and the new port 27018 is in effect

**What happened instead?**
The cluster status is running, the port is still 27017, and mongodb-agent reports an error
It says 【 Cannot simultaneously change multiple ports in the same deployment item. Please do one of them at a time.Both process = mongodb-0 and process = mongodb-1 changed ports.】

**Screenshots**
image
image
image

**Operator Information**
 - Operator v0.7.3
 - MongoDB 4.4.13

 - Distribution
 - Version
 - Image Registry location (quay, or an internal registry)

**Additional context**
cr.yaml
```
---
kind: MongoDBCommunity
metadata:
  name: mongodb
spec:
  members: 5
  type: ReplicaSet
  version: "4.4.13"
  security:
    authentication:
      modes: ["SCRAM"]
  users:
    - name: shuaib
      db: admin
      passwordSecretRef: # a reference to the secret that will be used to generate the user's password
        name: my-user-password
      roles:
        - name: clusterAdmin
          db: admin
        - name: userAdminAnyDatabase
          db: admin
  additionalMongodConfig:
    net:
      port: 27018
      maxIncomingConnections: 100000
    replication:
      oplogSizeMB: 2048

  statefulSet:
    spec:
      template:
        spec:
          initContainers:
            - name: mongod-posthook
              imagePullPolicy: IfNotPresent
            - name: mongodb-agent-readinessprobe
              imagePullPolicy: IfNotPresent
          containers:
            - name: mongod
              imagePullPolicy: IfNotPresent
              resources:
                limits:
                  cpu: "8"
                  memory: 16G
                requests:
                  cpu: "8"
                  memory: 16G
            - name: mongodb-agent
              imagePullPolicy: IfNotPresent
              resources:
                limits:
                  cpu: "0.2"
                  memory: 250M
                requests:
                  cpu: "0.2"
                  memory: 200M


# the user credentials will be generated from this secret
# once the credentials are generated, this secret is no longer required
---
apiVersion: v1
kind: Secret
metadata:
  name: my-user-password
type: Opaque
stringData:
  password: echo1106

```

If possible, please include:
 - `kubectl describe` output
 - yaml definitions for your objects
 - log files for the operator and database pods

lsierant:
This patchset improves reconfiguration of MongoDB's ports in the replica set.
MongoDB Agent does not support changing port of more than one process at a time in a replica set.

Changes:
 * Introduced ReplicaSetPortManager to determine processes' ports in Service and Automation Config. 
 * Fixed handling nested map format (e.g. `net.port`) when unmarshalling `MongodConfiguration`
 * Improved handling ports both as ints and float64 in `MongodConfiguration` and `AutomationConfig`
 * Set explicit linux/amd64 platform when building agent-image, which fixes local builds on M1 macs.

Port change process:
 * When the port (`newPort`) is customised, but there aren't any pods yet, then all processes are configured with the `newPort` in automation config. This works as before.
 * When the cluster is running and the port is changed, each process need to be configured one by one:
  * ReplicaSetPortManager changes port in automation config only when all pods reached its goal states.
  * Until all processes are configured with `newPort` the service is configured with two ports: `oldPort` and `newPort`.
  * When all pods are changed one by one, the service is reconfigured to have only `newPort`.

closes #945
### All Submissions:


config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
```diff
@@ -5,12 +5,6 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.4.1
-    service.binding/type: 'mongodb'
-    service.binding/provider: 'community'
-    service.binding: 'path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret'
-    service.binding/connectionString: 'path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret,sourceKey=connectionString.standardSrv'
-    service.binding/username: 'path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret,sourceKey=username'
-    service.binding/password: 'path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret,sourceKey=password'
   creationTimestamp: null
```
irajdeep, line 8: This might have been caused by the commit hook -- but please revert this change.

controllers/replica_set_port_manager.go
```diff
@@ -0,0 +1,109 @@
+package controllers
```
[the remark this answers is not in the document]
irajdeep: > This probably should be in pkg/ which would allow it to be used in the enterprise operator

Indeed 

I wouldn't mind putting this under `automationconfig` package for now -- since it computes the ports for AC.

controllers/replicaset_controller_test.go
```diff
@@ -296,6 +300,177 @@ func TestService_usesCustomMongodPortWhenSpecified(t *testing.T) {
...
+	namespacedName := mdb.NamespacedName()
+
+	const oldPort = automationconfig.DefaultDBPort
+	const newPort = 8000
+
+	mgr := client.NewManager(&mdb)
+
+	r := NewReconciler(mgr)
+
+	t.Run("Prepare cluster and change port", func(t *testing.T) {
```
slaskawi, line 344: Just an idea (probably for the future) - if we used an [event recorder](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/recorder) in the controller and record when do we change ports, the test would be much simpler. We'd just check if proper events were emitted in proper order.

pkg/automationconfig/automation_config.go
```diff
@@ -60,6 +70,22 @@ func (p *Process) SetPort(port int) *Process {
 	return p.SetArgs26Field("net.port", port)
 }
 
+func (p *Process) GetPort() int {
+	if p.Args26 == nil {
+		return 0
+	}
+
+	// Args26 map could be manipulated in memory, e.g. via SetPort - then it will be as int,
```
slaskawi, line 78: It's the second time I'm reading this comment, so it makes me scratch my head - shouldn't we be using defensive copying when unmarshaling the configuration?

test/e2e/e2eutil.go
```diff
@@ -127,7 +127,7 @@ func NewTestMongoDB(ctx *Context, name string, namespace string) (mdbv1.MongoDBC
 										Name: "mongodb-agent",
 										Resources: corev1.ResourceRequirements{
 											Limits: map[corev1.ResourceName]resource.Quantity{
-												"cpu":    resource.MustParse("0.1"),
+												"cpu":    resource.MustParse("1.0"),
```
slaskawi, line 130: Quite a bit of change! Is it required?

slaskawi (commented):
The changes look great! I inlined a couple of questions (most of them to educate me). None of them is a blocker though.

api/v1/mongodbcommunity_types.go
```diff
@@ -339,8 +352,24 @@ func (m MongodConfiguration) GetDBDataDir() string {
...
 // If port is not specified, the default port of 27017 will be used.
 func (m MongodConfiguration) GetDBPort() int {
-	// When passed as a number "net.port" is unmarshalled into a float64
-	return int(objx.New(m.Object).Get("net.port").Float64(float64(automationconfig.DefaultDBPort)))
+	portValue := objx.New(m.Object).Get("net.port")
+
+	// Underlying map could be manipulated in memory, e.g. via SetDBPort - then it will be as int,
+	// or it could be deserialized from JSON and then integer in an untyped map will be deserialized as float64.
+	// It's behavior of https://pkg.go.dev/encoding/json#Unmarshal that is converting JSON integers as float64.
+	if portValue.IsInt() {
```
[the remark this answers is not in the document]
lsierant: `portValue` here is a  *objx.Value (https://github.com/stretchr/objx/blob/master/value.go#L10), so it is unfortunately not easily castable

api/v1/mongodbcommunity_types.go
```diff
@@ -339,8 +352,24 @@ func (m MongodConfiguration) GetDBDataDir() string {
...
+	} else if portValue.IsFloat64() {
+		return int(portValue.Float64(float64(automationconfig.DefaultDBPort)))
+	}
+
+	return automationconfig.DefaultDBPort
+}
+
+// SetDBPort ensures that port is stored as float64
+func (m MongodConfiguration) SetDBPort(port int) MongodConfiguration {
+	m.SetOption("net.port", float64(port))
```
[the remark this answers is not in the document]
lsierant: I also don't like it. We definitely should have only one type here. Ints are there only when we manipulate these maps in process, most probably in unit tests. I tried to get rid of ints there but that would require fixing some other existing tests. I tried to be as much noninvasive and backwards compatible with this change as possible as `MongodConfiguration` and `AutomationConfigs` are also heavily used in enterprise operator codebase.

api/v1/mongodbcommunity_types.go
```diff
@@ -339,8 +352,24 @@ func (m MongodConfiguration) GetDBDataDir() string {
...
+	// Underlying map could be manipulated in memory, e.g. via SetDBPort - then it will be as int,
+	// or it could be deserialized from JSON and then integer in an untyped map will be deserialized as float64.
+	// It's behavior of https://pkg.go.dev/encoding/json#Unmarshal that is converting JSON integers as float64.
+	if portValue.IsInt() {
+		return portValue.Int(automationconfig.DefaultDBPort)
+	} else if portValue.IsFloat64() {
+		return int(portValue.Float64(float64(automationconfig.DefaultDBPort)))
+	}
+
+	return automationconfig.DefaultDBPort
```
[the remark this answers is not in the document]
lsierant: Yes, definitely. But now it's not easy to log from inside these structures. So probably we would need to return error here and this would be an API change. So I decided to just be backwards compatible.
I would rather opt for validation routine which would check if everything is OK and not validate ad hoc in getters.

config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
lsierant: I need to check why we have this in every commit, kinda annoying.

controllers/replica_set_controller.go
```diff
@@ -442,17 +442,20 @@ func (r *ReplicaSetReconciler) deployMongoDBReplicaSet(mdb mdbv1.MongoDBCommunit
...
+		return err
+	}
 
-	svc := &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: mdb.Namespace}}
+	svc := &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: mdb.ServiceName(), Namespace: mdb.Namespace}}
 	op, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, svc, func() error {
 		resourceVersion := svc.ResourceVersion // Save resourceVersion for later
-		*svc = buildService(mdb)
+		*svc = r.buildService(mdb, processPortManager)
 		svc.ResourceVersion = resourceVersion
```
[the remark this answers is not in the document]
lsierant: I think because service builder does not populate this field and in docs there is:
```
	// Clients must treat these values as opaque and passed unmodified back to the server.
```
https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L171
But I could be mistaken.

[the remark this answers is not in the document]
lsierant: Isn't this field changed automatically by the api server when anything was changed in values?

controllers/replica_set_port_manager_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	generatePortMap := func(ports ...int) map[string]int {
+		portMap := map[string]int{}
+		for i, port := range ports {
+			portMap[podName(i).Name] = port
+		}
+		return portMap
+	}
+
+	testCases := []testCase{
```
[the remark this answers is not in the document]
lsierant: That's very nice idea!

controllers/replicaset_controller_test.go
```diff
@@ -296,6 +300,177 @@ func TestService_usesCustomMongodPortWhenSpecified(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name.Name,
+			Namespace: name.Namespace,
+			Annotations: map[string]string{
+				"agent.mongodb.com/version": versionStr,
+			},
+		},
+	}
+
+	err := c.Create(context.TODO(), &pod)
```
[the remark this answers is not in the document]
lsierant: If course it's not good. I thing handling proper context and canceling/timeouts should be a point in our tech debt. Right now we have context.TODO() basically everywhere. Let's do this outside this patchset, ok?

controllers/replicaset_controller_test.go
lsierant: That's nice idea, but wouldn't this make this test to be too coupled with implementation? We don't need to know if we change the service or the pod first but the end result.
Also here we use fake api clients, so we don't have any api machinery in unit tests.

pkg/automationconfig/automation_config.go
lsierant: What do you mean by "using defensive copying" here? 
The comment here is intentionally copied from MongodConfiguration. I lost too many hours chasing bugs with these ints and floats! As said before - it'd be good to get rid of ints!

test/e2e/e2eutil.go
lsierant: This is only in e2e tests, so I didn't give it much thought. 
When run in GHA and on local non M1 mac this was sufficient. Unfortunately on macs there is additional QEMU layer, so everything is less performant and with 0.1 limit is just hanging.
This change shouldn't be invasive, i.e. shouldn't cause pods to be unschedulable as it's only limit, not request.

controllers/replica_set_port_manager.go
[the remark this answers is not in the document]
lsierant: @irajdeep We cannot move it easily into automationconfig pkg because it's using pod state logic from agent package. And we have a nasty import cycle there.
But I think it fits in agent package as it's logic is also related to mongodb-agent limitation around changing port one by one. Also core logic uses routine for checking goal states and that is also in agent pkg.
Moved to agent for now. Let me know if it's ok.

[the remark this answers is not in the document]
irajdeep: > Moved to agent for now.

Sounds good.

controllers/replica_set_controller.go
```diff
@@ -461,6 +464,31 @@ func (r *ReplicaSetReconciler) ensureService(mdb mdbv1.MongoDBCommunity) error {
...
+	}
+
+	sts, err := r.client.GetStatefulSet(mdb.NamespacedName())
+	if err != nil {
+		if !apiErrors.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to get StatefulSet: %s", err)
+		}
+
+		// sts's name is needed for GetAllDesiredMembersPodState logic
+		sts.Namespace = mdb.Namespace
```
irajdeep, line 480: Should this logic be executed if err != nil?

pkg/agent/agent_readiness.go
```diff
@@ -53,6 +54,33 @@ func AllReachedGoalState(sts appsv1.StatefulSet, podGetter pod.Getter, desiredMe
 	return true, nil
 }
 
+func GetAllDesiredMembersPodState(sts appsv1.StatefulSet, podGetter pod.Getter, desiredMemberCount, targetConfigVersion int, log *zap.SugaredLogger) ([]PodState, error) {
```
irajdeep, line 57: Please add a comment to this method as to what it does

controllers/replica_set_controller.go
```diff
@@ -461,6 +464,31 @@ func (r *ReplicaSetReconciler) ensureService(mdb mdbv1.MongoDBCommunity) error {
 	return err
 }
 
+func (r *ReplicaSetReconciler) createProcessPortManager(mdb mdbv1.MongoDBCommunity) (*agent.ReplicaSetPortManager, error) {
```
irajdeep, line 467: Please add a method to this comment as what it does.

pkg/agent/agent_readiness.go
```diff
@@ -53,6 +54,33 @@ func AllReachedGoalState(sts appsv1.StatefulSet, podGetter pod.Getter, desiredMe
...
+	var podStates []PodState
```
irajdeep, line 58: nit: some optimization, since we know that the size of this slice would be we can pre-allocate here instead of appending.

pkg/agent/replica_set_port_manager.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/mongodb/mongodb-kubernetes-operator/pkg/automationconfig"
+	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
+)
+
+type ReplicaSetPortManager struct {
+	log              *zap.SugaredLogger
+	expectedPort     int
+	currentPodStates []PodState
+	currentAC        automationconfig.AutomationConfig
```
irajdeep, line 13: Do we really need to store the entire automation config here, can be just store the processes which is needed I guess?

pkg/agent/replica_set_port_manager.go
```diff
@@ -0,0 +1,108 @@
...
+			Name: "mongodb",
+		},
+	}
+
+	r.log.Debugf("Port change in progress, setting service ports: %+v", servicePorts)
+
+	return servicePorts
+}
+
+func (r *ReplicaSetPortManager) calculateExpectedPorts() (map[string]int, bool, int) {
```
irajdeep, line 61: This method needs a comment as to what `calculateExpectedPorts` means.

test/e2e/replica_set_mongod_port_change/replica_set_mongod_port_change_test.go
```diff
@@ -0,0 +1,53 @@
...
+	t.Run("Test Basic Connectivity", tester.ConnectivitySucceeds())
+	t.Run("AutomationConfig has the correct version", mongodbtests.AutomationConfigVersionHasTheExpectedVersion(&mdb, 1))
+	t.Run("Mongod setting net.port has been set", tester.EnsureMongodConfig("net.port", int32(automationconfig.DefaultDBPort)))
+	t.Run("Service has the correct port", mongodbtests.ServiceUsesCorrectPort(&mdb, int32(automationconfig.DefaultDBPort)))
+	t.Run("Wait for MongoDB to finish setup cluster", mongodbtests.MongoDBReachesRunningPhase(&mdb))
+	t.Run("Change port of running cluster", mongodbtests.ChangePort(&mdb, 40333))
+	t.Run("Wait for MongoDB to start changing port", mongodbtests.MongoDBReachesPendingPhase(&mdb))
+	t.Run("Wait for MongoDB to finish changing port", mongodbtests.MongoDBReachesRunningPhase(&mdb))
+	t.Run("Mongod setting net.port has been set", tester.EnsureMongodConfig("net.port", int32(40333)))
+	t.Run("Service has the correct port", mongodbtests.ServiceUsesCorrectPort(&mdb, int32(40333)))
```
irajdeep, line 52: It would be prudent to do a `ConnectivitySucceeds` test here as well.

irajdeep (commented):
Nice work on this one! Added few comments inline.

controllers/replicaset_controller_test.go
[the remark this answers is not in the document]
slaskawi: Of course! Totally agree!

controllers/replica_set_controller.go
lsierant: This is executed only if err IsNotFound. But if you're asking I'll look if it can be improved.

api/v1/mongodbcommunity_types.go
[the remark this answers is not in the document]
slaskawi: Thanks for clarification!

I've just stumbled upon this from the [manual page](https://github.com/stretchr/objx#reading-data) of the `objx`:

```
m := objx.MustFromJSON(`{"name": "Mat", "age": 30}`)

// Get the details
name := m.Get("name").Str()
age := m.Get("age").Int()
```

Since we're using objx this way or another, maybe it's better to use its own unmarshaling and `.Int()` methods?

But I'm leaving the final choice to you.

api/v1/mongodbcommunity_types.go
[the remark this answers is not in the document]
slaskawi: This is actually a nice idea! I like it!

api/v1/mongodbcommunity_types.go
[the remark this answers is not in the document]
slaskawi: Fair enough! Thanks for the explanation.

controllers/replica_set_controller.go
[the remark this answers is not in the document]
slaskawi: I tried to play with `kubectl create` and `kubectl replace` (the `kubectl apply` is a bad choice as it records its own annotation with the last seen configuration) and it seems those values are always replaced by the API Server. I would need to dig deeper into it but at a first glance, calling `svc.ResourceVersion = resourceVersion` has no effect.

controllers/replicaset_controller_test.go
slaskawi: There are test event recorder implementation available as well. The idea is to treat it as a "flight recorder". You execute a test, wait till the end result in some sort of `eventually` call and once the result is fine, you analyse the events if things happened in a proper order. This is quite a popular approach in core OpenShift operators.

test/e2e/e2eutil.go
slaskawi: ooh yes, I'm sorry, it's limits and I'm blind!

pkg/automationconfig/automation_config.go
slaskawi: By defensive copy I mean that you replace the internal map whenever you mutate it (just clone it, update one of the values and replace it). This way, whatever caller has a reference to that map, it has a guarantee it won't change.

api/v1/mongodbcommunity_types.go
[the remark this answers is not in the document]
lsierant: From what I see in the source code it's not doing anything more than regular json.Unmarshall as objx.Map does not have UnmarshallJSON method. They only add tryConvertFloat64() as a postProcess step so you don't get float64 from integer fields. But this will only add fuel to the fire as we expect to get floats in many places.

pkg/agent/agent_readiness.go
lsierant: Added. 
Btw. I also spotted that we don't need statefulSet here - only namespace and name. So I have also refactored it a bit. But I haven't changed the signature of `AllReachedGoalState` as it is used in enterprise.

controllers/replica_set_controller.go
lsierant: Done.

pkg/agent/agent_readiness.go
lsierant: Done.

pkg/agent/replica_set_port_manager.go
lsierant: Good catch. We need only processes!

lsierant: Done

pkg/agent/replica_set_port_manager.go
lsierant: Done. Also added some more docs also in code as the process is not straightforward at first glance.

api/v1/mongodbcommunity_types.go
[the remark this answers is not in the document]
slaskawi: Thanks for the explanation. Understood!

pkg/automationconfig/automation_config.go
slaskawi: :thumbsup:

lsierant: Done

controllers/replica_set_port_manager.go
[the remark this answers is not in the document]
lsierant: Done

config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
lsierant: Done

controllers/replica_set_controller.go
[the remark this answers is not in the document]
lsierant: @slaskawi Would you like to investigate it further? For now in this PR I'll leave it as is.

controllers/replica_set_port_manager_test.go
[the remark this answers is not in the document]
lsierant: Done

pkg/automationconfig/automation_config.go
lsierant: Done

controllers/replica_set_controller.go
lsierant: I have removed it completely.

test/e2e/replica_set_mongod_port_change/replica_set_mongod_port_change_test.go
lsierant: Done.

.action_templates/jobs/tests.yaml
```diff
@@ -54,3 +54,7 @@ tests:
           distro: ubi
         - test-name: replica_set_mount_connection_string
           distro: ubi
+        - test-name: replica_set_mongod_port_change
+          distro: ubi
+        - test-name: replica_set_mongod_port_change_with_arbiters
```
irajdeep, line 59: Curious why do we need 2 tests here, can we just run 1 e2e test which has both MongoDB deployment and arbiters?

lsierant: My thought process was that configuration with arbiters and without them is different, so we should have e2e tests to both scenarios. But if I think about it now it is unnecessary as arbiters functionality is just an extension, so one suit will suffice. Removed it.

test/e2e/replica_set_mongod_port_change_with_arbiters/replica_set_mongod_port_change_with_arbiters_test.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	connectivityTests := func(t *testing.T) {
+		fmt.Printf("connectionStringForUser: %s\n", mongodbtests.GetConnectionStringForUser(mdb, scramUser))
+		t.Run("Test Basic Connectivity with generated connection string secret",
+			tester.ConnectivitySucceeds(WithURI(mongodbtests.GetConnectionStringForUser(mdb, scramUser))))
+
+		// FIXME after port change in the service mongodb+srv connection stopped working!
```
lsierant, line 47: I have confirmed that this is most probably an issue with e2e tests/mongodb go driver only. I've confirmed that SRV query after port change is returning the new port correctly. Also mongosh connection to the replicaset using mongodb-srv connection string from secret is working correctly.

controllers/replica_set_controller.go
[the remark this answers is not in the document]
slaskawi: No, it's fine to leave it as is in this PR. If we ever wanted to change this, it should have it's own PR.

PR #1019, opened 2022-06-21
CLOUDP-125628 security context update

slaskawi:
This Pull Request removes `securityContext` set on container level and moves all the previously used settings to the pod level. Note, that previously the pod level settings didn't make any effect as they were overriden by the container level.

### All Submissions:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,11 +32,8 @@ all: manager
 
 # Run unit tests
 TEST ?= ./pkg/... ./api/... ./cmd/... ./controllers/... ./test/e2e/util/mongotester/...
-ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
 test: generate fmt vet manifests
-	mkdir -p ${ENVTEST_ASSETS_DIR}
-	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.7.0/hack/setup-envtest.sh
-	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test $(TEST) -coverprofile cover.out
+	go test $(TEST) -coverprofile cover.out
 
 # Build manager binary
 manager: generate fmt vet
--- a/controllers/construct/build_statefulset_test.go
+++ b/controllers/construct/build_statefulset_test.go
@@ -1,6 +1,8 @@
 package construct
 
 import (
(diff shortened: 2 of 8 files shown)


pkg/kube/podtemplatespec/podspec_template.go
```diff
@@ -160,10 +160,11 @@ func WithSecurityContext(securityContext corev1.PodSecurityContext) Modification
 	}
 }
 
-// DefaultPodSecurityContext returns the default pod security context with FsGroup = 2000
+// DefaultPodSecurityContext returns the default pod security context with uid 2000 and non root
 func DefaultPodSecurityContext() corev1.PodSecurityContext {
-	fsGroup := int64(2000)
```
slaskawi, line 165: This is required!!!

pkg/kube/podtemplatespec/podspec_template.go
```diff
@@ -160,10 +160,12 @@ func WithSecurityContext(securityContext corev1.PodSecurityContext) Modification
 	}
 }
 
-// DefaultPodSecurityContext returns the default pod security context with FsGroup = 2000
```
slaskawi, line 163: Correct description!

controllers/construct/build_statefulset_test.go
```diff
@@ -87,6 +88,7 @@ func assertStatefulSetIsBuiltCorrectly(t *testing.T, mdb mdbv1.MongoDBCommunity,
 	assert.Equal(t, mongodbDatabaseServiceAccountName, sts.Spec.Template.Spec.ServiceAccountName)
 	assert.Len(t, sts.Spec.Template.Spec.Containers[0].Env, 4)
 	assert.Len(t, sts.Spec.Template.Spec.Containers[1].Env, 1)
+	assert.Equal(t, podtemplatespec.DefaultPodSecurityContext(), *sts.Spec.Template.Spec.SecurityContext)
```
lsierant, line 91: Is it possible that `sts.Spec.Template.Spec.SecurityContext` is nil here? It's good to have require.NotNil before dereferencing if it's a possibility. If it's nil the panic is polluting the logs.

Makefile
```diff
@@ -32,11 +32,8 @@ all: manager
 
 # Run unit tests
 TEST ?= ./pkg/... ./api/... ./cmd/... ./controllers/... ./test/e2e/util/mongotester/...
-ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
 test: generate fmt vet manifests
-	mkdir -p ${ENVTEST_ASSETS_DIR}
```
mircea-cosbuc, line 37: Nice catch, I thought this was used by the test actions.

mircea-cosbuc (approved):
LGTM

controllers/construct/build_statefulset_test.go
slaskawi: In this  unit test, it's not. However, you have a very good point, I should add a test testing a managed security (so that it will be nil). Good spot!

slaskawi: Added more tests in this area (and a requested nil/non-nil check).

PR #1034, opened 2022-06-30
(DOCSP-18367) Updating secure deployment doc to reflect simplified procedure using cert-manager integration.

lmkerbey-mdb:
[JIRA](https://jira.mongodb.org/browse/DOCSP-18367)

@mircea-cosbuc Ready for your technical review.


docs/secure.md
```diff
@@ -8,99 +8,72 @@
...
-   kubectl create secret tls <tls-secret-name> --cert=server.crt --key=server.key --namespace <namespace>
+   helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --set installCRDs=true
    ```
 
-### Procedure
-
-To secure connections to MongoDBCommunity resources using TLS:
-
-1. Add the following fields to the MongoDBCommunity resource definition:
+1. Perform an upgrade to the MongoDBCommunity resource you wish to
```
mircea-cosbuc, line 48: This would rather be `Create a TLS secured MongoDBCommunity` resource, since there's none created before this step.

mircea-cosbuc (approved):
LGTM with one comment

mircea-cosbuc (approved):
LGTM

docs/secure.md
```diff
@@ -8,99 +8,68 @@
...
-     version: "4.2.7"
-     security:
-       tls:
-         enabled: true
-         certificateKeySecretRef:
-           name: <tls-secret-name>
-         caConfigMapRef:
-           name: <tls-ca-configmap-name>
-   ```
+   - Then, use `mongo` to connect over TLS:
```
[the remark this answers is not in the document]
jwilliams-mongo: command should be similar, but names of the options / syntax might be slightly different. See the [MongoDB Shell docs](https://www.mongodb.com/docs/mongodb-shell/reference/options/#tls-options) for more info.

docs/secure.md
```diff
@@ -8,99 +8,68 @@
...
 
-1. Apply the configuration to Kubernetes:
    ```
-   kubectl apply -f <example>.yaml --namespace <my-namespace>
+   mongo --tls --tlsCAFile /var/lib/tls/ca/ca.crt --tlsCertificateKeyFile /var/lib/tls/server/*.pem --host <mongodb-replica-set>.<mongodb-replica-set>-svc.<namespace>.svc.cluster.local
```
jwilliams-mongo, line 71: question: will the ca file and the cert key file be mounted to these locations on the replica set pods?

docs/secure.md
```diff
@@ -8,99 +8,68 @@
```
jwilliams-mongo, line 8: question: do we also need to document how to rotate certificates? been a while since I looked at this, but I think you can do this by changing these values: https://github.com/mongodb/helm-charts/blob/main/charts/community-operator/values.yaml#L86-L87

jwilliams-mongo (commented):
@lmkerbey-mdb content looks good % a couple of minor comments and a few questions for engineering

docs/secure.md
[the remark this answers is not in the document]
lmkerbey-mdb: Compared with `mongosh`, looks like the syntax is the same in this case.

jwilliams-mongo (approved):
@lmkerbey-mdb LGTM % one small correction

PR #1039, opened 2022-07-06
Tool for editing cluster config secret

lsierant:
With `scripts/dev/edit_cluster_config.sh` we can easily edit the cluster config in the secret that is mounted into mongodb-agents' pods. 
It's handy when there is a need to tinker with the automation agent directly. 

### All Submissions:

### DIFF
--- a/scripts/dev/edit_cluster_config.sh
+++ b/scripts/dev/edit_cluster_config.sh
@@ -1,5 +1,16 @@
 #!/usr/bin/env bash
 
+# This script is intended for editing the raw cluster configuration that is used by the automation agents.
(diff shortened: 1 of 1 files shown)


scripts/dev/edit_cluster_config.sh
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env bash
+
```
slaskawi, line 2: Would be good to describe what it does here and even create some md file.

lsierant: Done. Is it OK?

irajdeep (approved):
LGTM

slaskawi (approved):
LGTM

PR #1064, opened 2022-08-04
CLOUDP-131647: Update to stateful set ignored existing volumes

lsierant:
In 0.7.3 there were changes to TLS secrets. Upgrading from 0.7.2 to 0.7.3, 0.7.4 was broken because the underlying secret in stateful set volume wasn't updated. It was because the code for updating volumes explicitly ignored any updates to volumes if they already exist, even if they contain important changes.

### All Submissions:

### DIFF
--- a/controllers/mongodb_tls_test.go
+++ b/controllers/mongodb_tls_test.go
@@ -8,7 +8,7 @@ import (
 
 	mdbv1 "github.com/mongodb/mongodb-kubernetes-operator/api/v1"
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/automationconfig"
-	"github.com/mongodb/mongodb-kubernetes-operator/pkg/kube/client"
+	kubeClient "github.com/mongodb/mongodb-kubernetes-operator/pkg/kube/client"
 	mdbClient "github.com/mongodb/mongodb-kubernetes-operator/pkg/kube/client"
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/kube/configmap"
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/kube/secret"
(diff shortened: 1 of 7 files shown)


test/e2e/replica_set_operator_upgrade/replica_set_operator_upgrade_test.go
```diff
@@ -41,19 +41,76 @@ func TestReplicaSetOperatorUpgrade(t *testing.T) {
...
 	err = setup.DeployOperator(config, "mdb", false, false)
 	assert.NoError(t, err)
 
 	// Perform the basic tests
 	t.Run("Basic tests", mongodbtests.BasicFunctionality(&mdb, true))
 }
+
+// TestReplicaSetOperatorUpgradeFrom0_7_2 is intended to be run locally not in CI.
+// It simulates deploying cluster using community operator 0.7.2 and then upgrading it using newer version.
+func TestReplicaSetOperatorUpgradeFrom0_7_2(t *testing.T) {
```
lsierant, line 63: As a followup to this I would make this test to be running on CI. But this needs some more work, as it would require installing CRDs for 0.7.2 and then master inside the test, which afaik we do not currently do.

mircea-cosbuc (approved):
This looks good to me, not sure why the TLS Upgrade test is failing though.

controllers/mongodb_tls_test.go
```diff
@@ -23,10 +23,10 @@ func TestStatefulSet_IsCorrectlyConfiguredWithTLS(t *testing.T) {
 	mdb := newTestReplicaSetWithTLS()
 	mgr := client.NewManager(&mdb)
 
-	client := mdbClient.NewClient(mgr.GetClient())
-	err := createTLSSecret(client, mdb, "CERT", "KEY", "")
+	cli := mdbClient.NewClient(mgr.GetClient())
```
irajdeep, line 26: Q: Why this rename IMO `client` was more descriptive.

irajdeep:
Nice work, This change should also have an entry in release notes.

controllers/mongodb_tls_test.go
lsierant, line 26: client is shadowing package `client`, e.g. used in line 24. I was getting warnings about it. So either we alias package or change var name. Yes `cli` is not a great name :(

lsierant: @irajdeep I mistakenly deleted your Q: why `cli` because `client` was more descriptive

irajdeep:
>client is shadowing package client, e.g. used in line 24. I was getting warnings about it. So either we alias package or change var name. Yes cli is not a great name :(

Let's rename the package name then.

irajdeep (approved):
LGTM

PR #1078, opened 2022-08-15
Bump K8s dependencies

irajdeep:
### All Submissions:


go.mod
```diff
@@ -14,42 +14,51 @@ require (
 	go.mongodb.org/mongo-driver v1.10.1
 	go.uber.org/zap v1.21.0
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0
-	k8s.io/api v0.23.6
-	k8s.io/apimachinery v0.23.6
-	k8s.io/client-go v0.23.6
-	sigs.k8s.io/controller-runtime v0.11.2
+	k8s.io/api v0.24.3
```
slaskawi, line 17: We need to be careful with this one. It means, we're bumping the supported Kube's version to `1.24`. I'm OK with this but just wanted to make it visible. 

Also, could you please check if we mention the Kube's version in our manual pages or helm charts? This would also need to be (probably) updated.

irajdeep: > It means, we're bumping the supported Kube's version to 1.24.

Indeed, that's correct.
> Also, could you please check if we mention the Kube's version in our manual pages or helm charts?

This is something we currently do only for enterprise operator and not the community. Probably we should do for both.

slaskawi: 👌

slaskawi (approved):

PR #1091, opened 2022-08-22
CLOUDP-130491: Daily rebuilds of the operator image

lsierant:
This PR contains changes enabling daily rebuilds of `mongodb-kubernetes-operator` image as part of daily rebuild process in the enterprise operator repository. Operator image is now built in the similar way as mongodb-agent, leveraging daily rebuild pipeline in the enterprise repository.

Changes:
* Moved operator docker templates to its own `scripts/dev/templates/operator` directory.
* Changed docker build process of the operator to be performed in two steps.
  * Build `context` image which contains the operator binary.
  * Build target operator image from ubi8:latest as a base and copy the operator binary from context image. Dockerfile for this step is published to S3 storage containing dockerfiles for each supported version. 
* Moved golang builder image version into one place - into release.json.
* Added support for storing released versions `mongodb-kubernetes-operator` in  `add_supported_release.py` 

### All Submissions:


inventories/operator-inventory.yaml
```diff
@@ -14,24 +14,44 @@ images:
...
-
+        template_file_extension: operator
         inputs:
-          - builder
-          - builder_image
           - base_image
 
         output:
-          - dockerfile: scripts/dev/templates/Dockerfile.ubi-$(inputs.params.version_id)
+          - dockerfile: scripts/dev/templates/operator/Dockerfile.operator-$(inputs.params.version_id)
```
slaskawi, line 43: So it seems we're now only using UBI images. If that is the case, would you mind to mention this in the release notes?

release.json
```diff
@@ -1,5 +1,6 @@
 {
-  "mongodb-kubernetes-operator": "0.7.5",
+  "golang-builder-image": "golang:1.18.5",
```
slaskawi, line 2: Actually, could we use `golang:1.18` here and never worry about the micro updates?

On one hand I know the build reproducibility argument but on the other hand, this is the fastest way to ensure we're compatible with the latest micro of the Golang. I think this outweighs the build reproducibility argument.

scripts/dev/templates/operator/Dockerfile.builder
```diff
@@ -0,0 +1,20 @@
...
+
+RUN mkdir /workspace
+WORKDIR /workspace
+
+COPY go.mod go.sum ./
+COPY cmd/manager/main.go cmd/manager/main.go
+COPY api/ api/
+COPY controllers/ controllers/
+COPY pkg/ pkg/
+COPY build/bin/ build/bin/
```
slaskawi, line 12: This is very fragile. Whenever we introduce a new directory, we'll be scratching our heads why it's not included in the image by default.

Here's an alternative way to do it: https://github.com/openshift/cert-manager-operator/blob/master/images/ci/Dockerfile#L2-L3

scripts/dev/templates/operator/Dockerfile.builder
```diff
@@ -0,0 +1,20 @@
...
+
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager cmd/manager/main.go
+
+FROM busybox
```
slaskawi, line 16: `scratch` would work better here.

scripts/dev/templates/versionhook/Dockerfile.builder
```diff
@@ -1,4 +1,5 @@
-FROM golang:1.18.5 as builder
+ARG builder_image
+FROM ${builder_image} as builder
 
 ENV GO111MODULE=on
```
slaskawi, line 4: Not related to your changes but since you have this "under your axe" - could you please delete those unnecessary Go env. variables?

scripts/dev/templates/operator/Dockerfile.operator
```diff
@@ -0,0 +1,18 @@
+ARG imagebase
+FROM ${imagebase} AS builder
+
+FROM {{base_image}}
+
+ENV OPERATOR=manager \
+    USER_UID=1001 \
```
slaskawi, line 7: Can we change this to `2000` and align with the SecurityContext we already have in the code? https://github.com/mongodb/mongodb-kubernetes-operator/blob/e8406136eea04ac52d48ac232f6ebd1104b76acc/pkg/kube/podtemplatespec/podspec_template.go#L167

scripts/ci/add_supported_release.py
```diff
@@ -71,8 +76,17 @@ def main() -> int:
     parser.add_argument(
         "--image-name", help="image to add a new supported version", type=str
     )
+    parser.add_argument(
```
slaskawi, line 79: Does this align with the Enterprise `pipeline.py`?

release.json
```diff
@@ -1,5 +1,6 @@
...
+  "mongodb-kubernetes-operator": "0.0.1",
```
slaskawi, line 3: Why this is `0.0.1`? I suspect this is either a leftover or you wanted to leave it like this for testing. If that is so, you might also use `0.7.5+<SHA from git>`. In semantic versioning everything after the plus sign is treated as build id. Alternatively, if you're using this for testing, you might want to use the minus sign (so that this will be treated as pre-release and won't replace the original release).

slaskawi (changes requested):

inventories/operator-inventory.yaml
lsierant: It hasn't changed. We haven't built ubuntu in the community at all.

release.json
lsierant: Yes, I like the idea!

release.json
lsierant: Yes, this shouldn't be here. Thanks for the idea with SHA!

scripts/ci/add_supported_release.py
lsierant: These two unaligned if you compare them. It's not worth the cost - we'll soon get rid of adding versions to the external mongodb database and switch to simple file in repo. But this is the next step I need to convince the team on :)

scripts/dev/templates/operator/Dockerfile.builder
lsierant: Yes, I know. It's good point. I'll change it. The funny thing is that I just mostly copied the old dockerfile here :)

scripts/dev/templates/operator/Dockerfile.builder
lsierant: I wanted to preserve the ability to run the context image, e.g. for investigating what's inside. If we can do this using scratch (which I doubt) then I could change it. Otherwise I would leave it as busybox is ~1MB.

lsierant: Also this is something that was before.

release.json
```diff
@@ -1,4 +1,5 @@
 {
+  "golang-builder-image": "golang:1.18",
```
mircea-cosbuc, line 2: 👍 Nice, this should make it easier to manage golang version upgrade

mircea-cosbuc (approved):

lsierant:
Agreed with @slaskawi to merge without re-review.

slaskawi (commented):
LGTM

PR #1103, opened 2022-08-31
CLOUDP-132091: Add TLS and arbiters to operator upgrade test

mircea-cosbuc:
This patch enables the operator upgrade test in the e2e suite and adds TLS and arbiter settings to the resource in the test.


lsierant (approved):
LGTM! with a minor thing

irajdeep (approved):

PR #1196, opened 2023-01-18
[log] fix log output for pods that don't exist yet

nammn:
### Changes:
* When checking whether our agents are ready or not we log when a Pod is not ready. There is a logging bug when the pod is not running yet. This means we are checking the goal state of a pod that does not exist yet. This will lead to the following output:
```
DEBUG   agent/agent_readiness.go:102    The Pod '' doesn't have annotation 'agent.mongodb.com/version' yet  {"ReplicaSet": "default/mdb0"}
```
*  When the pod is not around (yet) there is no goalState that needed to be checked, so we can skip that check where the wrong log came from.
*  Fixes some comments

### All Submissions:

### DIFF
--- a/pkg/agent/agent_readiness.go
+++ b/pkg/agent/agent_readiness.go
@@ -81,7 +81,11 @@ func GetAllDesiredMembersAndArbitersPodState(namespacedName types.NamespacedName
 		p, err := podGetter.GetPod(podNamespacedName)
 		if err != nil {
 			if apiErrors.IsNotFound(err) {
+				// we can skip below iteration and check for our goal state since the pod is not available yet
 				podState.Found = false
+				podState.ReachedGoalState = false
+				podStates[i] = podState
+				continue
 			} else {
 				return nil, err
 			}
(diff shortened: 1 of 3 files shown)


pkg/agent/agent_readiness.go
```diff
@@ -81,7 +81,11 @@ func GetAllDesiredMembersAndArbitersPodState(namespacedName types.NamespacedName
 		p, err := podGetter.GetPod(podNamespacedName)
 		if err != nil {
 			if apiErrors.IsNotFound(err) {
+				// we can skip below iteration and check for our goal state since the pod is not available yet
 				podState.Found = false
+				podState.ReachedGoalState = false
+				podStates[i] = podState
```
nammn, line 87: Alternatively (which reduces LoC), one can make a podName check here: https://github.com/mongodb/mongodb-kubernetes-operator/blob/2f089f9fb61c7c2e33dbba5465f294c6c78df031/pkg/agent/agent_readiness.go#L104-L108 but this only checks whether the podName is empty. Here we definitely know whether the error is not found.

mircea-cosbuc (approved):
LGTM

PR #1198, opened 2023-01-19
[fix] fixes a sts controller race condition with the mongodb-agent

nammn:
## Context
During version changes the following happens:
- sts is ready
- we write a new automation config 
- agent receives new automation config
- `mongodb-agents` will start to delete pods in a rolling update fashion (not the agents itself, but the result is the same)
- primary will be replaced last (rolling update)
- every pods readiness probe fails (we do have some time until we call it unhealthy)
- depending on the time it took a pod can get unhealthy as the readiness probe fails. 
- if the pod got unhealthy, the sts controller will not be able to start up a new pod that we just recently killed

=> deadlock

## Solution/Changes
- mark the probe/container as ready during the upgrade step
- some comment fixes
- also dump event list to know who killed what 

### All Submissions:

### DIFF
--- a/cmd/readiness/main.go
+++ b/cmd/readiness/main.go
@@ -87,7 +87,7 @@ func isPodReady(conf config.Config) (bool, error) {
 
 // hasDeadlockedSteps returns true if the agent is stuck on waiting for the other agents
 func hasDeadlockedSteps(health health.Status) bool {
-	currentStep := findGivenStep(health.ProcessPlans)
+	currentStep := findCurrentStep(health.ProcessPlans)
 	if currentStep != nil {
 		// this means we are waiting for external factors before we are able to continue our update. If we expect the node to be up, then we can assume that it is ready.
 		// To match the rest of the code, a more proper way would be to look at all previous steps and
@@ -99,13 +99,13 @@ func hasDeadlockedSteps(health health.Status) bool {
 	return false
 }
 
-// findGivenStep returns the step which seems to be run by the Agent now. The step is always in the last plan
+// findCurrentStep returns the step which seems to be run by the Agent now. The step is always in the last plan
 // (see https://github.com/10gen/ops-manager-kubernetes/pull/401#discussion_r333071555) so we iterate over all the steps
 // there and find the last step which has "Started" non nil
 // (indeed this is not the perfect logic as sometimes the agent doesn't update the 'Started' as well - see
 // 'health-status-ok.json', but seems it works for finding deadlocks still
 // noinspection GoNilness
-func findGivenStep(processStatuses map[string]health.MmsDirectorStatus) *health.StepStatus {
+func findCurrentStep(processStatuses map[string]health.MmsDirectorStatus) *health.StepStatus {
 	var currentPlan *health.PlanStatus
 	if len(processStatuses) == 0 {
 		// Seems shouldn't happen but let's check anyway - may be needs to be changed to Info if this happens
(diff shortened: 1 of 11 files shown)


controllers/construct/mongodbstatefulset.go
```diff
@@ -245,7 +245,7 @@ func DefaultReadiness() probes.Modification {
 	return probes.Apply(
 		probes.WithExecCommand([]string{readinessProbePath}),
 		probes.WithFailureThreshold(40),
-		probes.WithInitialDelaySeconds(5),
+		probes.WithInitialDelaySeconds(15),
```
nammn, line 248: note: if that helps, this is not the solution, just a measure to figure out the cause

cmd/readiness/testdata/health-status-deadlocked-with-prev-config.json
```diff
@@ -0,0 +1,159 @@
...
+                {
+                  "step": "CheckRsCorrect",
+                  "stepDoc": "Check that replica set configuration is correct",
+                  "isWaitStep": false,
+                  "started": "2023-01-19T17:38:33.622960067Z",
+                  "completed": "2023-01-19T17:38:33.623363973Z",
+                  "result": "success"
+                },
+                {
+                  "step": "WaitCanUpdate",
```
nammn, line 111: Our test fail here. Agents receive a new automation config with above step. By current definition they are never in `inGoalState` since they are waiting for the update and therefore `unready`. Depending on how fast the agent is able to realise updates the pod can get unhealthy and therefore the cluster. Fully stopping the update due to sts controller not continuing

cmd/readiness/main.go
```diff
@@ -89,6 +89,11 @@ func isPodReady(conf config.Config) (bool, error) {
 func hasDeadlockedSteps(health health.Status) bool {
 	currentStep := findCurrentStep(health.ProcessPlans)
 	if currentStep != nil {
+		// this means we are waiting for external factors before we are able to continue our update. If we expect the node to be up, then we can assume that it is ready.
+		// To match the rest of the code, a more proper way would be to look at all previous steps and
+		if currentStep.Step == "WaitCanUpdate" && currentStep.Started != nil && currentStep.Completed == nil && currentStep.Result == "wait" {
```
nammn, line 94: alternative: adding `WaitCanUpdate` to parts of the deadlock risky steps

cmd/readiness/testdata/health-status-deadlocked-with-prev-config.json
nammn: I verified the following:
- all agents receive the new automation config containing the `WaitCanUpdate`
- agents are getting replaced one after other and therefore `lose` the `WaitCanUpdate` step, since they were replaced
- the primary got replaced last, giving it more time to get unhealthy

cmd/readiness/main.go
mircea-cosbuc: It makes sense to me to move this to the risky steps. Is it a check that we might want to handle differently in the future and it's worth keeping separate?

mircea-cosbuc (approved):
LGTM, very nice catch!

cmd/readiness/main.go
nammn: Done in 3c1576a !


The only "con"/difference I see is the 15 second health check: https://github.com/mongodb/mongodb-kubernetes-operator/blob/3c1576a6c6987f1ae5e07493f0fa58862fd80f79/cmd/readiness/main.go#L145

which means that our readinessProbe will "fail" for a max of 10 seconds until the time has gone. Which should not impact the version change process as after those 10 seconds we will return "ready" probes

irajdeep (approved):
LGTM

PR #1229, opened 2023-02-16
CLOUDP-156372 Improve Service merging

slaskawi:
### All Submissions:

### DIFF
--- a/pkg/util/merge/merge_service_spec.go
+++ b/pkg/util/merge/merge_service_spec.go
@@ -5,13 +5,70 @@ import (
 )
 
 // ServiceSpec merges two ServiceSpecs together.
-// The implementation does not override:
-// - labels/selectors
-// - cluster IPs
+// The implementation merges Selector instead of overriding it.
 func ServiceSpec(defaultSpec, overrideSpec corev1.ServiceSpec) corev1.ServiceSpec {
-	mergedSpec := overrideSpec
+	mergedSpec := defaultSpec
 	mergedSpec.Selector = StringToStringMap(defaultSpec.Selector, overrideSpec.Selector)
-	mergedSpec.ClusterIPs = defaultSpec.ClusterIPs
-	mergedSpec.ClusterIP = defaultSpec.ClusterIP
+
+	if len(overrideSpec.Ports) != 0 {
+		mergedSpec.Ports = overrideSpec.Ports
+	}
+
+	if len(overrideSpec.Type) != 0 {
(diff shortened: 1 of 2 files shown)


pkg/util/merge/merge_test.go
```diff
@@ -101,25 +101,11 @@ func TestMergeServices(t *testing.T) {
 				Selector: map[string]string{"test1": "true", "test2": "true"},
```
nammn, line 101: how do we handle selectors with the same key but different values? Should we add a test for this?

nammn (approved):

pkg/util/merge/merge_service_spec.go
```diff
@@ -5,13 +5,70 @@ import (
...
-// - cluster IPs
+// The implementation merges Selector instead of overriding it.
 func ServiceSpec(defaultSpec, overrideSpec corev1.ServiceSpec) corev1.ServiceSpec {
-	mergedSpec := overrideSpec
+	mergedSpec := defaultSpec
 	mergedSpec.Selector = StringToStringMap(defaultSpec.Selector, overrideSpec.Selector)
-	mergedSpec.ClusterIPs = defaultSpec.ClusterIPs
-	mergedSpec.ClusterIP = defaultSpec.ClusterIP
+
+	if len(overrideSpec.Ports) != 0 {
```
lsierant, line 13: I'm not sure this is the right approach, because this would prevent us from actually setting zero-value in override. 
For IP or Type field this is not a problem, but field like PublishNotReadyAddresses would never set it to false.

slaskawi: That's a fair point. We could leave this as it was before. The only problem is that you won't be able to do cascading overrides in MongoDBMulti, such as:

1. use `PublishNotReadyAddresses` as `true` by default
2. override type and set it to `loadbalancer` in the `spec`
3. override individual Load Balancer IPs in specific clusterSpecs. 

You will have to always specify all 3 properties (`PublishNotReadyAddresses`, Load Balancer type and Load Balancer IPs) in the clusterSpec. I think that's fine, although - a bit counter intuitive. But if anyone complains about it, we could enhance this in the future.

lsierant: I'm slightly for less convenient approach, but without surprises, so as it was before.

lsierant (approved):

slaskawi:
Closing this as the current behavior should be sufficient.

PR #1271, opened 2023-04-13
CLOUDP-169740 Enable use of the EA images

slaskawi:
# Summary

This Pull Request makes possible to run EA Server images with the Operator.

### DIFF
--- a/controllers/replica_set_controller.go
+++ b/controllers/replica_set_controller.go
@@ -579,8 +579,11 @@ func guessEnterprise(mdb mdbv1.MongoDBCommunity) bool {
 			}
 		}
 	}
-	if strings.Contains(overriddenImage, construct.OfficialMongodbEnterpriseServerImageName) {
-		return true
+	if len(overriddenImage) > 0 {
+		if strings.Contains(overriddenImage, construct.OfficialMongodbEnterpriseServerImageName) {
+			return true
+		}
+		return false
 	}
 	return os.Getenv(construct.MongodbImageEnv) == construct.OfficialMongodbEnterpriseServerImageName
 }
--- a/controllers/replica_set_controller.go
+++ b/controllers/replica_set_controller.go
@@ -541,6 +541,7 @@ func buildAutomationConfig(mdb mdbv1.MongoDBCommunity, auth automationconfig.Aut
 	}
 
 	return automationconfig.NewBuilder().
+		IsEnterprise(os.Getenv(construct.MongodbImageEnv) == construct.OfficialMongodbEnterpriseServerImageName).
 		SetTopology(automationconfig.ReplicaSetTopology).
 		SetName(mdb.Name).
 		SetDomain(domain).
--- a/controllers/replica_set_controller.go
+++ b/controllers/replica_set_controller.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
+	"strings"
 
 	"github.com/imdario/mergo"
 	"github.com/stretchr/objx"
@@ -541,7 +543,7 @@ func buildAutomationConfig(mdb mdbv1.MongoDBCommunity, auth automationconfig.Aut
 	}
 
 	return automationconfig.NewBuilder().
-		IsEnterprise(os.Getenv(construct.MongodbImageEnv) == construct.OfficialMongodbEnterpriseServerImageName).
+		IsEnterprise(guessEnterprise(mdb)).
 		SetTopology(automationconfig.ReplicaSetTopology).
 		SetName(mdb.Name).
 		SetDomain(domain).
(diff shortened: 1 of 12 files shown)


pkg/automationconfig/automation_config_builder.go
```diff
@@ -156,6 +162,9 @@ func (b *Builder) AddVersion(version MongoDbVersionConfig) *Builder {
 
 func (b *Builder) SetMongoDBVersion(version string) *Builder {
 	b.mongodbVersion = version
+	if b.isEnterprise {
```
slaskawi, line 165: I'm scratching my head if this is the best approach. Modifying something in a setter is mostly considered as a bad practice, but this is really what is happening here - the proper version for enterprise has the `-ent` suffix. Without it, it's considered as community, which does not reflect the true. 

That was the main argument for doing this in a setter. WDYT?

nammn: As long as there is no different way to mark a image as enterprise (i.e. through a setting) I think thats ok

controllers/replica_set_controller.go
```diff
@@ -541,6 +541,7 @@ func buildAutomationConfig(mdb mdbv1.MongoDBCommunity, auth automationconfig.Aut
 	}
 
 	return automationconfig.NewBuilder().
+		IsEnterprise(os.Getenv(construct.MongodbImageEnv) == construct.OfficialMongodbEnterpriseServerImageName).
```
nammn, line 544: This actually might not be true, because our environment variable (at least in enterprise)  does not have to be the source of truth, since we can define a wrapper in our crd. 💀 

For internal reference: https://github.com/10gen/ops-manager-kubernetes/blob/25eacd7470ef05bf699e4eea185757c482a36614/api/v1/om/appdb_types.go#L258-L275 + https://github.com/10gen/ops-manager-kubernetes/blob/a68a1f281e297eec893ea936c9f2316722aefc8b/docker/mongodb-enterprise-tests/tests/opsmanager/withMonitoredAppDB/om_appdb_configure_all_images.py#L36-L51

pkg/automationconfig/automation_config_builder.go
slaskawi: This still bothers me - let me rework this piece of the code.

slaskawi: Addressed. Closing.

controllers/replica_set_controller.go
slaskawi: I implemented a piece of logic with `guessEnterprise`. It checks if there's an override via env. variable and if it's missing - it tries to guess the value. I think with this approach we should be on the safe side as you can always override to bring it to desired behavior.

controllers/replica_set_controller.go
```diff
@@ -559,6 +562,29 @@ func buildAutomationConfig(mdb mdbv1.MongoDBCommunity, auth automationconfig.Aut
 		Build()
 }
 
+func guessEnterprise(mdb mdbv1.MongoDBCommunity) bool {
```
nammn, line 565: On enterprise we will need something similar. We could make this re-usable by changing the signature to take a wrapper of containers instead

controllers/replica_set_controller.go
```diff
@@ -559,6 +562,29 @@ func buildAutomationConfig(mdb mdbv1.MongoDBCommunity, auth automationconfig.Aut
...
+	overrideAssumption, err := strconv.ParseBool(os.Getenv(construct.MongoDBAssumeEnterpriseEnv))
```
nammn, line 566: I don't understand this logic, where would we ever set this env var?

controllers/replicaset_controller_test.go
```diff
@@ -173,7 +173,103 @@ func TestStatefulSet_IsCorrectlyConfigured(t *testing.T) {
...
+			setArgs: func(t *testing.T) {
+				t.Setenv(construct.MongodbRepoUrl, "some_other_repo.com/some_other_org")
+				t.Setenv(construct.MongodbImageEnv, "mongodb-community-server")
+			},
+			mdb:                mdbv1.MongoDBCommunity{},
+			expectedEnterprise: false,
+		},
+		// This one is a corner case. We don't expect users to fall here very often as there are
+		// dedicated variables to control this type of behavior.
+		"Enterprise with StatefulSet override": {
```
nammn, line 237: can we add a test case for the other way around? Setting env var to enterprise, but sts wrapper to non enterprise

controllers/replica_set_controller.go
```diff
@@ -559,6 +562,29 @@ func buildAutomationConfig(mdb mdbv1.MongoDBCommunity, auth automationconfig.Aut
...
+	if len(containers) > 0 {
+		for _, c := range containers {
+			if c.Name == construct.MongodbName {
+				if len(c.Image) > 0 {
+					overriddenImage = c.Image
+				}
+			}
+		}
+	}
+	if strings.Contains(overriddenImage, construct.OfficialMongodbEnterpriseServerImageName) {
```
nammn, line 582: shouldn't we return false if we have an override which is **not** enterprise? Imagine the testcase which I mention further below

controllers/replica_set_controller.go
slaskawi: So this is a backdoor that we might want to set at the Operator Deployment Spec level (in env. vars). It's not documented anywhere, it just prevents the operator from guessing and enforces the certain behavior. It's here just in case we do the guessing wrong.

controllers/replicaset_controller_test.go
slaskawi: This is great spot! Indeed, there's a bug in the implementation but I also had to add an explicit check if the overridden image is really there. Please re-check the code.

controllers/replica_set_controller.go
slaskawi: Answering this here https://github.com/mongodb/mongodb-kubernetes-operator/pull/1271#discussion_r1171775852

controllers/replica_set_controller.go
slaskawi: I think guessing is not necessary in the enterprise, where we rely only on the enterprise server image. So I believe we'll always set `true` there.

controllers/replica_set_controller.go
nammn: I see, that makes sense!

slaskawi: Yes, definitely! I will be looking into the internal docs shortly!

controllers/replica_set_controller.go
slaskawi: This has been answered internally.

nammn (approved):
LGTM! Thanks for the changes!

PR #1324, opened 2023-07-04
CLOUDP-186802: Add connection string options 

### LINKED ISSUE #1134: Additional flags to autogenerated connection string
I deployed a simple MongoDB ReplicaSet as example:

```yml
kind: MongoDBCommunity
metadata:
  name: temp-mongodb-replicaset
spec:
  type: ReplicaSet
  members: 3
  version: 6.0.2
  security:
    authentication:
      modes:
        - SCRAM
      ignoreUnknownUsers: false
  users:
    - name: temp-mongodb-admin
      db: admin
      passwordSecretRef:
        name: mongodb-kubernetes-operator-admin-creds
      roles:
        - name: clusterAdmin
          db: admin
        - name: userAdminAnyDatabase
          db: admin
      scramCredentialsSecretName: temp-mongodb-admin
      connectionStringSecretName: temp-mongodb-admin-connection-string
  additionalMongodConfig:
  statefulSet:
    spec:
      selector:
        matchLabels:
          component: mongodb
          replicaset: "true"
      template:
        metadata:
          labels:
            component: mongodb
            replicaset: "true"
        spec:
          containers:
            - name: mongod
              resources:
                requests:
                  cpu: 500m
                  memory: 1Gi
                limits:
                  cpu: 1
                  memory: 2Gi
            - name: mongodb-agent
              resources:
                requests:
                  cpu: 50m
                  memory: 64Mi
                limits:
                  cpu: 100m
                  memory: 128Mi
              readinessProbe:
                failureThreshold: 50
                initialDelaySeconds: 10
          affinity:
            podAntiAffinity:
              preferredDuringSchedulingIgnoredDuringExecution:
                - weight: 100
                  podAffinityTerm:
                    topologyKey: kubernetes.io/hostname
                    labelSelector:
                      matchExpressions:
                        - key: app
                          operator: In
                          values:
                            - mongodb
      volumeClaimTemplates:
        - metadata:
            name: data-volume
          spec:
            accessModes:
              - ReadWriteOnce
            storageClassName: ssd-csi-expandable
            resources:
              requests:
                storage: 3Gi
        - metadata:
            name: logs-volume
          spec:
            accessModes:
              - ReadWriteOnce
            storageClassName: ssd-csi-expandable
            resources:
              requests:
                storage: 2Gi
  prometheus:
    username: metrics
    passwordSecretRef:
      name: mongodb-kubernetes-operator-metrics-creds
```

As I got a secret named `temp-mongodb-admin-connection-string` with this connection string as content: `mongodb+srv://temp-mongodb-admin:***@temp-mongodb-replicaset-svc.temp.svc.cluster.local/admin?replicaSet=temp-mongodb-replicaset&ssl=false`

Is it possible to add flags to the connection string through the operator? For example in MongoDBCommunity CRD?

Flags like `readPreference` would be really important.

**Operator Information**
 - Operator Version:   0.7.6
 - MongoDB Image used:   docker.io/mongo:6.0.2

 - Distribution/Version:   v1.21.14-gke.2700
 - Image Registry location (quay, or an internal registry):   quay

lucian-tosa:
### Example:

```yaml
  kind: MongoDBCommunity
  metadata:
    name: example-mongodb
  spec:
    members: 3
    type: ReplicaSet
    version: "6.0.5"
    security:
      authentication:
        modes: ["SCRAM"]
    users:
      - name: my-user
        db: admin
        passwordSecretRef:
          name: my-user-password
        roles:
          - name: clusterAdmin
            db: admin
          - name: userAdminAnyDatabase
            db: admin
        additionalConnectionStringConfig:
          readPreference: secondary
    additionalMongodConfig:
    additionalConnectionStringConfig:
      readPreference: primary
  ```
- In this example, the connection string for ``my-user`` will have as a connection string option ``readPreference=secondary`` by overriding the resource option. 
- Any options included will **not** be validated and using the connection strings with incorrect options will fail.

Closes #1134 

### All Submissions:

### CODE REVIEW

File: api/v1/mongodbcommunity_types.go [hunk 1]
```diff
@@ -736,29 +761,92 @@ func (m *MongoDBCommunity) AutomationConfigArbitersThisReconciliation() int {
...
+	}
+	return optionsString
+}
+
 // MongoURI returns a mongo uri which can be used to connect to this deployment
 func (m *MongoDBCommunity) MongoURI(clusterDomain string) string {
-	return fmt.Sprintf("mongodb://%s/?replicaSet=%s", strings.Join(m.Hosts(clusterDomain), ","), m.Name)
+	optionsString := m.GetOptionsString()
+
+	return fmt.Sprintf("mongodb://%s/?replicaSet=%s%s", strings.Join(m.Hosts(clusterDomain), ","), m.Name, optionsString)
```
mircea-cosbuc, line 823: Since the `replicaSet` option is hardcoded in the connection string, what happens if it's provided as an additional option? Does it need to be filtered out of the computed `optionsString` ?

lucian-tosa: Indeed, it will be duplicated in the connection string. The connection string would still work as any duplicate would be simply ignored. However, it makes sense that it should be filtered out.

mircea-cosbuc: Would the connection string use the first option? We should make sure that is the case since we require using the `replicaSet` created with the operator through the automation config.

lucian-tosa: Yes, it uses the first option.

api/v1/mongodbcommunity_types.go
```diff
@@ -119,6 +119,69 @@ type MongoDBCommunitySpec struct {
 	// Prometheus configurations.
 	// +optional
 	Prometheus *Prometheus `json:"prometheus,omitempty"`
+
+	// Additional options to be appended to the connection string. These options apply to the entire resource and to each user.
+	// +kubebuilder:validation:Type=object
+	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +nullable
+	AdditionalConnectionStringConfig MapWrapper `json:"additionalConnectionStringConfig,omitempty"`
```
slaskawi, line 128: Would it be simpler if we put a `map[string]interface{}` directly there?

api/v1/mongodbcommunity_types.go
```diff
@@ -119,6 +119,69 @@ type MongoDBCommunitySpec struct {
...
+
+// MarshalJSON defers JSON encoding to the wrapped map
+func (m *MapWrapper) MarshalJSON() ([]byte, error) {
+	return json.Marshal(m.Object)
+}
+
+// UnmarshalJSON will decode the data into the wrapped map
+func (m *MapWrapper) UnmarshalJSON(data []byte) error {
+	if m.Object == nil {
+		m.Object = map[string]interface{}{}
```
slaskawi, line 147: With this code, does it make sense to use `// +nullable` in the type definition?

api/v1/mongodbcommunity_types_test.go
```diff
@@ -28,6 +30,61 @@ func TestMongoDB_MongoURI(t *testing.T) {
 	assert.Equal(t, mdb.MongoURI("my.cluster"), "mongodb://my-rs-0.my-rs-svc.my-namespace.svc.my.cluster:40333,my-rs-1.my-rs-svc.my-namespace.svc.my.cluster:40333/?replicaSet=my-rs")
 }
 
+func TestMongoDB_MongoURI_With_Options(t *testing.T) {
+	mdb := newReplicaSet(2, "my-rs", "my-namespace")
```
slaskawi, line 34: Might be nicer with parametric test, such as this one: https://github.com/mongodb/mongodb-kubernetes-operator/blob/master/test/e2e/replica_set_arbiter/replica_set_arbiter_test.go#L26

slaskawi (approved):
Added a few questions and a suggestion but overall - great work 💪 LGTM!

api/v1/mongodbcommunity_types.go
lucian-tosa: I am not completely sure that it's related. I believe that `// +nullable` allows the field `AdditionalConnectionStringConfig` itself to be **nil**, and I don't think it's related to the `Object` field of `MapWrapper`, which could be **nil** anyways. I may be wrong. However, I agree that removing the marker would be better.

api/v1/mongodbcommunity_types_test.go
lucian-tosa: Done! Although, for these short unit tests, it feels a bit overkill.

api/v1/mongodbcommunity_types.go
```diff
@@ -736,29 +761,96 @@ func (m *MongoDBCommunity) AutomationConfigArbitersThisReconciliation() int {
...
+//
+// Takes into account both user options and resource options.
+// User options will override any existing options in the resource.
+func (m *MongoDBCommunity) GetUserOptionsString(user scram.User) string {
+	generalOptionsMap := m.Spec.AdditionalConnectionStringConfig.Object
+	userOptionsMap := user.ConnectionStringOptions
+	optionValues := make([]string, len(generalOptionsMap)+len(userOptionsMap))
+	i := 0
+	for key, value := range userOptionsMap {
+		if key != "replicaSet" && key != "ssl" && key != "tls" {
```
mircea-cosbuc, line 800: It would be nice to extract the protected options in a constant for easier maintenance.

api/v1/mongodbcommunity_types.go
lucian-tosa: Upon further testing, I notice that having this marker allows for using the SetOption method without actually checking if the map is nil or not. This way, it helps to avoid any nil pointer exceptions. So having it does make sense.

api/v1/mongodbcommunity_types.go
lucian-tosa: Done! However, Go doesn't allow constant maps or any other reference type for that matter. So I used a variable map, since having multiple string constants wouldn't be easily maintainable.

api/v1/mongodbcommunity_types.go
slaskawi: And what's the error then (out of curiosity)? I think it just requires a policy how to merge entries but I might be wrong.

api/v1/mongodbcommunity_types.go
slaskawi: Thanks for checking!

api/v1/mongodbcommunity_types.go
lucian-tosa: I did not try that actually. But according to this open issue: https://github.com/kubernetes-sigs/controller-tools/issues/636 and the comments that were already in the code, it should not work.

slaskawi:
Perfect! Thanks for checking!

PR #1332, opened 2023-07-12
CLOUDP-189434 Using Wait Steps instead of risky Steps

slaskawi:
## Summary

This Pull Request changes the core deadlock detection algorithm and replaces it with detecting Wait Steps (the steps with `isWaitStep: true`. This loosens up the way the Probe behaves and makes it less restrictive. 

The end user impact is completely opposite now. In the worst case scenario, we will report the database to be ready (instead of not ready and blocking the rollout as we've seen before). The end user experience is much better as it may experience a short outage instead of full database blackout. 

This Pull Request also introduces name refactoring and aligns it with the field names in JSON file. This way, it's less confusing when discussing this with a wider Team.

## Technical details

A lot of this design is related to this conversation: https://github.com/mongodb/mongodb-kubernetes-operator/pull/1332

## Testing results in both Community and EA

* All tests against Community seem to be fine. I can successfully delete any number of Pods (being careful not to lose quorum, which causes a short outage - until a new quorum is established - see RAFT algorithm)
* All tests against EA seem to be fine. When introducing AutomationConfig changes and deleting Pods during the Rolling Update, it's easy to lose quorum. But even then, the system recovers in a few seconds. 

All tests against the Enterprise are successful [link](https://github.com/10gen/ops-manager-kubernetes/pull/2985).

### DIFF
--- a/cmd/readiness/main.go
+++ b/cmd/readiness/main.go
@@ -136,7 +136,12 @@ func findCurrentStep(processStatuses map[string]health.MmsDirectorStatus) *healt
 	return lastStartedStep
 }
 
-// isWaitStep return true is the Agent is currently waiting for something.
+// isWaitStep returns true is the Agent is currently waiting for something to happen.
+//
+// Most of the time, the Agent waits for an initialization by other member of the cluster. In such case,
+// holding the rollout does not improve the overall system state. Even if the probe returns true too quickly
+// the worst thing that can happen is a short service interruption, which is still better than full service outage.
+//
 // The 15 seconds explanation:
 //   - The status file is written every 10s but the Agent processes steps independently of it
 //   - In order to avoid reacting on a newly added wait Step (as they can naturally go away), we're giving the Agent
--- a/cmd/readiness/main.go
+++ b/cmd/readiness/main.go
@@ -10,7 +10,6 @@ import (
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/readiness/config"
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/readiness/headless"
 	"github.com/mongodb/mongodb-kubernetes-operator/pkg/readiness/health"
-	"github.com/mongodb/mongodb-kubernetes-operator/pkg/util/contains"
 	"go.uber.org/zap/zapcore"
 	"gopkg.in/natefinch/lumberjack.v2"
 
@@ -25,12 +24,9 @@ const (
 	mongodNotReadyIntervalMinutes = time.Minute * 1
 )
 
-var riskySteps []string
 var logger *zap.SugaredLogger
 
 func init() {
-	riskySteps = []string{"WaitAllRsMembersUp", "WaitRsInit", "WaitCanUpdate"}
-
 	// By default, we log to the output (convenient for tests)
 	cfg := zap.NewDevelopmentConfig()
 	log, err := cfg.Build()
@@ -45,8 +41,7 @@ func init() {
 // The logic depends on if the pod is a standard MongoDB or an AppDB one.
 // - If MongoDB: then just the 'statuses[0].IsInGoalState` field is used to learn if the Agent has reached the goal
 // - if AppDB: the 'mmsStatus[0].lastGoalVersionAchieved' field is compared with the one from mounted automation config
-// Additionally if the previous check hasn't returned 'true' the "deadlock" case is checked to make sure the Agent is
-// not waiting for the other members.
+// Additionally if the previous check hasn't returned 'true' an additional check for wait steps is being performed
 func isPodReady(conf config.Config) (bool, error) {
 	healthStatus, err := parseHealthStatus(conf.HealthStatusReader)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


cmd/readiness/main.go
```diff
@@ -25,12 +24,10 @@ const (
 	mongodNotReadyIntervalMinutes = time.Minute * 1
 )
 
-var riskySteps []string
```
slaskawi, line 28: To be removed

cmd/readiness/main.go
```diff
@@ -73,33 +69,35 @@ func isPodReady(conf config.Config) (bool, error) {
...
 	}
 
 	// Fallback logic: the agent is not in goal state and got stuck in some steps
-	if !inGoalState && hasDeadlockedSteps(healthStatus) {
+	if !inGoalState && isOnWaitingStep(healthStatus) {
+		logger.Info("The Agent is on wait Step. Returning ready.")
 		return true, nil
 	}
 
+	logger.Info("Reached the end of the check. Returning not ready.")
```
slaskawi, line 82: A note for the future - we might want to consider building a logging context with debug information and once e return not ready, we could append why. Potentially this could be implemented withing this or next ticket.

cmd/readiness/main.go
```diff
@@ -138,12 +137,18 @@ func findCurrentStep(processStatuses map[string]health.MmsDirectorStatus) *healt
...
+//   - In order to avoid reacting on a newly added wait Step (as they can naturally go away), we're giving the Agent
+//     at least 15 sends to spend on that Step.
+//   - This hopefully prevents the Probe from flipping False to True too quickly.
+func isWaitStep(status *health.StepStatus) bool {
 	// Some logic behind 15 seconds: the health status file is dumped each 10 seconds, so we are sure that if the agent
 	// has been in the step for 10 seconds - this means it is waiting for the other hosts, and they are not available
 	fifteenSecondsAgo := time.Now().Add(time.Duration(-15) * time.Second)
-	if contains.String(riskySteps, status.Step) && status.Completed == nil && status.Started.Before(fifteenSecondsAgo) {
-		logger.Infof("Indicated a possible deadlock, status: %s, started at %s but hasn't finished "+
+	if status.IsWaitStep && status.Completed == nil && status.Started.Before(fifteenSecondsAgo) {
```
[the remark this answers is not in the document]
slaskawi: Thanks a lot - I'll perform all the test in both Community and Enterprise.

> I'm not sure what changes that would be in AC that would cause new plans to be generated

This one is very, very tricky. I only managed to artificially test it by handcrafting the changes (and it worked fine). The thing is that under certain circumstances, the Automation Agent might completely recompute the plan. Then, the new plan is appended to the (therefore, it's the last one). So far I've seen only once such situation - someone seemed to delete the automation agent user when rolling out new deployment. However, I failed to recreate this scenario with more recent versions of the Agent.

cmd/readiness/main.go
slaskawi: On the second thought - we return not ready when we don't fall into any other condition in the checking loop. The only thing we could log there is - 👆 statement. And this doesn't make sense.

cmd/readiness/main.go
```diff
@@ -55,8 +50,8 @@ func isPodReady(conf config.Config) (bool, error) {
 	}
 
 	// The 'statuses' file can be empty only for OM Agents
-	if len(healthStatus.Healthiness) == 0 && !isHeadlessMode() {
-		logger.Info("'statuses' is empty. We assume there is no automation config for the agent yet.")
+	if len(healthStatus.Statuses) == 0 && !isHeadlessMode() {
+		logger.Info("'statuses' is empty. We assume there is no automation config for the agent yet. Returning ready.")
```
slaskawi, line 54: Change to debug

cmd/readiness/main.go
```diff
@@ -138,12 +136,18 @@ func findCurrentStep(processStatuses map[string]health.MmsDirectorStatus) *healt
...
+//     at least 15 sends to spend on that Step.
+//   - This hopefully prevents the Probe from flipping False to True too quickly.
+func isWaitStep(status *health.StepStatus) bool {
 	// Some logic behind 15 seconds: the health status file is dumped each 10 seconds, so we are sure that if the agent
 	// has been in the step for 10 seconds - this means it is waiting for the other hosts, and they are not available
 	fifteenSecondsAgo := time.Now().Add(time.Duration(-15) * time.Second)
-	if contains.String(riskySteps, status.Step) && status.Completed == nil && status.Started.Before(fifteenSecondsAgo) {
-		logger.Infof("Indicated a possible deadlock, status: %s, started at %s but hasn't finished "+
+	if status.IsWaitStep && status.Completed == nil && status.Started.Before(fifteenSecondsAgo) {
+		logger.Infof("Indicated a wait Step, status: %s, started at %s but hasn't finished "+
```
slaskawi, line 150: Change to debug

cmd/readiness/main.go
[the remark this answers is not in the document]
slaskawi: I performed all the test you mentioned. They all seem to be fine and at the same time, the probe seems to properly guard the cluster from rolling multiple members at once.

cmd/readiness/main.go
```diff
@@ -138,12 +136,18 @@ func findCurrentStep(processStatuses map[string]health.MmsDirectorStatus) *healt
 	return lastStartedStep
 }
 
-func isDeadlocked(status *health.StepStatus) bool {
+// isWaitStep return true is the Agent is currently waiting for something.
```
nammn, line 139: Can we add a bit more context what a `waitstep` is or why it matters here to return "true"?

slaskawi: Done

nammn (approved):
LGTM, thanks for finding this!

Issue #1344, opened 2023-08-02
Endless Rolling Restarts when using Tolerations


siegenthalerroger:
**What did you do to encounter the bug?**

Had a MongoDB resource with v0.8.0 like below (incomplete) and then upgraded to v0.8.1.

```
apiVersion: mongodbcommunity.mongodb.com/v1
kind: MongoDBCommunity
metadata:
  name: mongodb
  namespace: demo-mongodb
spec:
  members: 2
  type: ReplicaSet
  version: "6.0.8"
  statefulSet:
    spec:
      template:
        metadata:
          labels:
            variant: global
        spec:
          containers:
            - ... (resource configuration for agent + mongod, aswell as mongo-exporter sidecar)
          tolerations:
            - key: "optimised-for"
              operator: "Equal"
              value: "memory"
              effect: "PreferNoSchedule"
```

**What did you expect?**

The cluster to continue to operate with the newer operator.

**What happened instead?**

There was an endless rolling restart loop between the two replicas. 
 **Downgrading to 0.8.0 rectified the issue**. 

I triaged this to be due to the operator repeatedly adding the "toleration" defined above to the StatefulSet, causing a state difference which k8s reconciled by triggering a rolling update. I suspect this bug was introduced by #1314.

Following a cleaned-up variant of the operator-reconciled mongo StatefulSet:

<details>
```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: mongodb
  namespace: demo-mongodb
  uid: 1922f590-e379-4bff-8362-8a7a5cbfc352
  resourceVersion: '242788943'
  generation: 2798
  creationTimestamp: '2023-08-02T11:41:33Z'
  ownerReferences:
    - apiVersion: mongodbcommunity.mongodb.com/v1
      kind: MongoDBCommunity
      name: mongodb
      uid: 3bbe2f15-b9d8-4c26-8507-eb538927bcce
      controller: true
      blockOwnerDeletion: true
  managedFields:
    - manager: manager
      operation: Update
      apiVersion: apps/v1
      time: '2023-08-02T11:41:33Z'
      fieldsType: FieldsV1
      fieldsV1:
        f:metadata:
          f:ownerReferences:
            .: {}
            k:{"uid":"3bbe2f15-b9d8-4c26-8507-eb538927bcce"}: {}
        f:spec:
          f:podManagementPolicy: {}
          f:replicas: {}
          f:revisionHistoryLimit: {}
[...]

PR #1346, opened 2023-08-04
CLOUDP-193669: Make overriden tolerations take precedence

mircea-cosbuc:
This patch addresses a bug where overriding tolerations in the StatefulSet spec would end up in an endless reconciliation loop. 

Addresses:  
- https://github.com/mongodb/mongodb-kubernetes-operator/issues/1344
- https://github.com/mongodb/mongodb-kubernetes-operator/issues/1349


slaskawi (approved):
LGTM!

One thing you might think of is whether this fix deserves an e2e test. To me personally, a unit test would be sufficient. But I'm leaving this up to you - great work!

mircea-cosbuc:
@slaskawi The STS override e2e tests were updated to include tolerations.

pkg/kube/podtemplatespec/podspec_template_test.go
```diff
@@ -247,6 +247,140 @@ func TestMergeEnvironmentVariables(t *testing.T) {
...
+					Operator: corev1.TolerationOpEqual,
+				},
+			},
+			overrideTolerations: []corev1.Toleration{},
+			expectedTolerations: []corev1.Toleration{},
+		},
+		{
+			// Overriding toleration should replace a nil original toleration.
+			name:               "Overriding when default toleration is nil",
+			defaultTolerations: nil,
```
lucian-tosa, line 304: Would it make sense to also test when defaultTolerations is an empty array?

test/e2e/statefulset_arbitrary_config/statefulset_arbitrary_config_test.go
```diff
@@ -50,4 +67,20 @@ func TestStatefulSetArbitraryConfig(t *testing.T) {
 	t.Run("Container has been merged by name", mongodbtests.StatefulSetContainerConditionIsTrue(&mdb, "mongodb-agent", func(container corev1.Container) bool {
 		return container.ReadinessProbe.TimeoutSeconds == 100
 	}))
+	t.Run("Tolerations have been added correctly", mongodbtests.StatefulSetConditionIsTrue(&mdb, func(sts appsv1.StatefulSet) bool {
```
lucian-tosa, line 70: It seems that this test doesn't check whether tolerations are added only once. Since this was the initial issue, maybe this deserves an e2e. Correct me if I'm wrong.

pkg/kube/podtemplatespec/podspec_template_test.go
```diff
@@ -247,6 +247,140 @@ func TestMergeEnvironmentVariables(t *testing.T) {
...
+				},
+				{
+					Key:      "key1",
+					Value:    "value2",
+					Operator: corev1.TolerationOpExists,
+				},
+			},
+		},
+		{
+			// Overriding toleration should replace any original toleration.
```
lucian-tosa, line 331: Is this the correct behaviour? As I noticed, the behaviour before the bug was introduced was more of "merge" instead of a total override.

mircea-cosbuc: Yes, this is the direction of this PR. Merging the tolerations only makes sense if the operator sets tolerations when creating the statefulsets, which it doesn't since it has no bias about the cluster topology. We want whatever is in the override to be reflected in the statefulset.

test/e2e/statefulset_arbitrary_config/statefulset_arbitrary_config_test.go
mircea-cosbuc: Good point, this should actually check if the tolerations are identical since only the pods get the default kubernetes tolerations added to them so it's not necessary to iterate and check if they are contained.

pkg/kube/podtemplatespec/podspec_template_test.go
mircea-cosbuc: This is identical to the case when `defaultTolerations` is a non-empty array, since we discard everything there. Should we change the logic, that test might make sense.

lucian-tosa:
LGTM

lucian-tosa (approved):

PR #1348, opened 2023-08-09
CLOUDP 189433 AC recovery

slaskawi:
This is a hypothetical fix for CLOUDP-189433. However it's hard to come up with scenarios when it happens. The only scenario I found is scaling up and disabling TLS at the same time.


slaskawi:
This makes no sense in community. Closing.

PR #1350, opened 2023-08-09
Bump and release agent version to 12.0.25.7724-1

lsierant:
* Bumped agent to `12.0.25.7724-1`
* mongodb-tools to `100.7.4`

### DIFF
--- a/release.json
+++ b/release.json
@@ -4,7 +4,7 @@
   "version-upgrade-hook": "1.0.7",
   "readiness-probe": "1.0.15",
   "mongodb-agent": {
-    "version": "12.0.24.7719-1",
-    "tools_version": "100.7.0"
+    "version": "12.0.25.7724-1",
+    "tools_version": "100.7.4"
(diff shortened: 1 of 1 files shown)


mircea-cosbuc (approved):

release.json
```diff
@@ -4,7 +4,7 @@
   "version-upgrade-hook": "1.0.7",
   "readiness-probe": "1.0.15",
   "mongodb-agent": {
-    "version": "12.0.24.7719-1",
-    "tools_version": "100.7.0"
+    "version": "12.0.25.7724-1",
```
lucian-tosa, line 7: LGTM! Does this need an entry in the release notes for updated images? Or should that be done when we release the operator?

lucian-tosa (approved):

release.json
lsierant: We do this at release time. But thanks for noticing it!

PR #1370, opened 2023-08-23
Returning ready if there are no plans in headless mode

lsierant:
Sometimes the agent is not returning IsInGoalState when mongod is scaled down. 
This change prevents the operator from being stuck and allows to progress and remove the pod from the statefulset.

If there are no plans after removing the process from automation config and IsInGoalState=false, the target version was -1 and it wasn't possible to recover automatically from this situation.

### DIFF
--- a/pkg/readiness/headless/headless.go
+++ b/pkg/readiness/headless/headless.go
@@ -67,19 +67,11 @@ func readCurrentAgentInfo(health health.Status, targetVersion int64) int64 {
 		zap.S().Debugf("Automation Config version: %d, Agent last version: %d", targetVersion, v.LastGoalStateClusterConfigVersion)
 		return v.LastGoalStateClusterConfigVersion
 	}
-	// The edge case: if the scale down operation is happening and the member + process are removed
-	// from the Automation Config - the Agent just doesn't write the 'mmsStatus' at all so there is no indication of
-	// the version it has achieved (though health file contains 'IsInGoalState=true')
-	// Let's return the desired version in case if the Agent is in goal state and no plans exist in the health file
-	for _, v := range health.Statuses {
-		if v.IsInGoalState {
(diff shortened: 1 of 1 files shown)


config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
```diff
@@ -1,15 +1,10 @@
...
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.3
-    service.binding: path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret
-    service.binding/connectionString: path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret,sourceKey=connectionString.standardSrv
-    service.binding/password: path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret,sourceKey=password
-    service.binding/provider: community
-    service.binding/type: mongodb
-    service.binding/username: path={.metadata.name}-{.spec.users[0].db}-{.spec.users[0].name},objectType=Secret,sourceKey=username
```
irajdeep, line 12: revert

irajdeep (approved):

slaskawi (approved):
LGTM but please create a JIRA for it.

config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
nammn: This happens if you have an old version under bin/controller-gen.

My suggestion:
- delete it
- run `make controller-gen` to retrieve the one with the correct version

PR #1479, opened 2024-01-25
fix confusing error message

nammn:
### All Submissions:

removing error message in case that file does not exist. We do want to use it in case it exist, with this we have the opportunity to connect to OM via monitoring.

### DIFF
--- a/controllers/construct/mongodbstatefulset.go
+++ b/controllers/construct/mongodbstatefulset.go
@@ -121,9 +121,6 @@ type MongoDBStatefulSetOwner interface {
 
 	// NeedsAutomationConfigVolume returns whether the statefulset needs to have a volume for the automationconfig.
 	NeedsAutomationConfigVolume() bool
-
-	// IsAppDB returns whether the statefulset owner is AppDB.
-	IsAppDB() bool
 }
 
 // BuildMongoDBReplicaSetStatefulSetModificationFunction builds the parts of the replica set that are common between every resource that implements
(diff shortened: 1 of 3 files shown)


controllers/construct/mongodbstatefulset.go
```diff
@@ -64,7 +64,7 @@ const (
 	automationAgentLogOptions = " -logFile ${AGENT_LOG_FILE} -maxLogFileDurationHrs ${AGENT_MAX_LOG_FILE_DURATION_HOURS} -logLevel ${AGENT_LOG_LEVEL}"
 
 	MongodbUserCommand = `current_uid=$(id -u)
-AGENT_API_KEY="$(cat /mongodb-automation/agent-api-key/agentApiKey)"
+AGENT_API_KEY="$(cat /mongodb-automation/agent-api-key/agentApiKey 2>/dev/null)"
```
mircea-cosbuc, line 67: We reference this const for AppDB construction as well. What's the interaction there?

controllers/construct/mongodbstatefulset.go
```diff
@@ -64,6 +64,17 @@ const (
 	automationAgentLogOptions = " -logFile ${AGENT_LOG_FILE} -maxLogFileDurationHrs ${AGENT_MAX_LOG_FILE_DURATION_HOURS} -logLevel ${AGENT_LOG_LEVEL}"
 
 	MongodbUserCommand = `current_uid=$(id -u)
+declare -r current_uid
```
nammn, line 67: We could merge with below var, but this seems easy enough once we need to do one change in the future

controllers/construct/mongodbstatefulset.go
```diff
@@ -225,7 +236,12 @@ func BaseAgentCommand() string {
 	return "agent/mongodb-agent -healthCheckFilePath=" + agentHealthStatusFilePathValue + " -serveStatusPort=5000"
 }
 
-func AutomationAgentCommand() []string {
+// AutomationAgentCommand withAgentAPIKeyExport detects whether we want to deploy this agent with the agent api key exported
+// it can be used to register the agent with OM.
+func AutomationAgentCommand(withAgentAPIKeyExport bool) []string {
```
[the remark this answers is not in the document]
nammn: related: https://github.com/10gen/ops-manager-kubernetes/pull/3364/files

controllers/construct/mongodbstatefulset.go
```diff
@@ -64,6 +64,17 @@ const (
...
+sed -e "s/^mongodb:/builder:/" /etc/passwd > /tmp/passwd
+echo "mongodb:x:$(id -u):$(id -g):,,,:/:/bin/bash" >> /tmp/passwd
+export NSS_WRAPPER_PASSWD=/tmp/passwd
+export LD_PRELOAD=libnss_wrapper.so
+export NSS_WRAPPER_GROUP=/etc/group
+fi
+
+`
+	MongodbUserCommandWithAPIKeyExport = `current_uid=$(id -u)
 AGENT_API_KEY="$(cat /mongodb-automation/agent-api-key/agentApiKey)"
```
lsierant, line 78: we might do this also by swallowing stderr to /dev/null.

lsierant (approved):
One comment to consider, but in any way LGTM

controllers/construct/mongodbstatefulset.go
nammn: Initially I implemented it that way: https://github.com/mongodb/mongodb-kubernetes-operator/pull/1479/commits/6526377527089b4d9f598761c22deb57f4d7047a but then @irajdeep had a great comment that he thinks it would be nice to make sure that we can see the error message in appdb in case monitoring does not work for any reason.

PR #1480, opened 2024-01-29
CLOUDP-226357 - fix logRotate & add local dev imp & update examples

nammn:
### All Submissions:

## Changes

**Make local development easier**
makes switching between enterprise operator and community easier if not relying on a test but applying yamls directly:
- patching the service accounts to use local image registries
- sourcing the env var

**Fix LogRotate red herring bug** 
- In case one did not setup a LogRotate configuration, the operator would default to neutral/default values and send them to the agent
- depending on the agent version/behaviour of parsing those default values it can error out
- solution is to make sure to only send them if it has been configured in the first place
- fixes a red herring error message

**Updating misleading examples where we override the app selector**


Makefile
```diff
@@ -101,11 +102,21 @@ install: manifests helm install-crd ## Install CRDs into a cluster
 install-crd:
 	kubectl apply -f config/crd/bases/mongodbcommunity.mongodb.com_mongodbcommunity.yaml
 
-install-chart:
+install-chart: uninstall-crd
```
nammn, line 105: before being able to install the helm chart we need to make sure that crds are not around. They will be installed by the chart anyway

Makefile
```diff
@@ -101,11 +102,21 @@ install: manifests helm install-crd ## Install CRDs into a cluster
...
 	$(HELM) upgrade --install $(STRING_SET_VALUES) $(RELEASE_NAME_HELM) $(HELM_CHART) --namespace $(NAMESPACE) --create-namespace
 
+install-chart-local-operator: uninstall-crd
+	$(HELM) upgrade --install $(STRING_SET_VALUES_LOCAL) $(RELEASE_NAME_HELM) $(HELM_CHART) --namespace $(NAMESPACE) --create-namespace
+
+prepare-local-dev: generate-env-file install-chart-local-operator setup-sas
+
+# patches all sas to use the local-image-registry
```
nammn, line 113: this makes sure if we are using a remote registry, that we can access it

pkg/automationconfig/automation_config.go
```diff
@@ -127,15 +127,15 @@ type LogRotate struct {
...
-	LogRotate                   AcLogRotate `json:"LogRotate,omitempty"`
+	Name                        string       `json:"name"`
+	Disabled                    bool         `json:"disabled"`
+	HostName                    string       `json:"hostname"`
+	Args26                      objx.Map     `json:"args2_6"`
+	FeatureCompatibilityVersion string       `json:"featureCompatibilityVersion"`
+	ProcessType                 ProcessType  `json:"processType"`
+	Version                     string       `json:"version"`
+	AuthSchemaVersion           int          `json:"authSchemaVersion"`
+	LogRotate                   *AcLogRotate `json:"LogRotate,omitempty"`
```
nammn, line 138: the pointer ensures that the json marshaller only adds the value if it exists.

scripts/dev/get_e2e_env_vars.py
```diff
@@ -33,6 +33,8 @@ def _get_e2e_test_envs(dev_config: DevConfig) -> Dict[str, str]:
         "MONGODB_REPO_URL": dev_config.mongodb_image_repo_url,
         "HELM_CHART_PATH": os.path.abspath("./helm-charts/charts/community-operator"),
         "MDB_IMAGE_TYPE": dev_config.image_type,
+        "MDB_LOCAL_OPERATOR": dev_config.local_operator,
+        "KUBECONFIG": dev_config.kube_config,
```
mircea-cosbuc, line 37: Nice!

mircea-cosbuc (approved):
LGTM

PR #1224, opened 2023-02-06
remove panic of missing health status.json

nammn:
### All Submissions:

* [ ] Have you opened an Issue before filing this PR?
* [x] Have you signed our [CLA](https://www.mongodb.com/legal/contributor-agreement)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
* [ ] Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).

## Reason
We panic on purpose on errors to have the information in the pod description.

But we always panic due to missing the health status file, since the file will can be missing during early of the pod startup. Making this a red herring in the describe.

```
│   Warning  Unhealthy  15m   kubelet            Readiness probe failed: panic: open /var/log/mongodb-mms-automation/agent-health-status.json: no such file │
│  or directory                                                                                                                                             │
│                                                                                                                                                           │
│ goroutine 1 [running]:                                                                                                                                    │
│ main.main()                                                                                                                                               │
│            /build/main.go:217 +0x19a                                                                                                                      │
│   Warning  Unhealthy  12m (x10 over 15m)  kubelet  Readiness probe failed:
```

PR #1504, opened 2024-03-07
fix red herring in the readinessProbe

nammn:
### All Submissions:

### Summary
- fixes the red herring that the agent-health-status file is missing
- continue of fix: https://github.com/mongodb/mongodb-kubernetes-operator/pull/1224, we missed a few places. This PR instead suggests to test for the file as early as possible in the code
- it can be missing, since the agent requires some time until it populates that file
- while it is still populating the probe should just return unready and it will eventually work

### DIFF
--- a/cmd/readiness/main.go
+++ b/cmd/readiness/main.go
@@ -221,12 +221,15 @@ func main() {
 		panic(err)
 	}
 
+	initLogger(config.GetLogger())
+
 	healthStatusFilePath := config.GetEnvOrDefault(config.AgentHealthStatusFilePathEnv, config.DefaultAgentHealthStatusFilePath)
 	file, err := os.Open(healthStatusFilePath)
 	// The agent might be slow in creating the health status file.
 	// In that case, we don't want to panic to show the message
 	// in the kubernetes description. That would be a red herring, since that will solve itself with enough time.
 	if err != nil {
+		logger.Errorf("health status file not avaible yet: %s ", err)
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/readiness/main.go
```diff
@@ -221,7 +221,16 @@ func main() {
...
 	}
 
-	cfg, err := config.BuildFromEnvVariables(clientSet, isHeadlessMode())
+	healthStatusFilePath := config.GetEnvOrDefault(config.AgentHealthStatusFilePathEnv, config.DefaultAgentHealthStatusFilePath)
+	file, err := os.Open(healthStatusFilePath)
+	// The agent might be slow in creating the health status file.
+	// In that case, we don't want to panic to show the message
+	// in the kubernetes description. That would be a red herring, since that will solve itself with enough time.
+	if err != nil {
+		os.Exit(1)
```
lsierant, line 230: can we log something here to indicate what happened?

nammn: I was thinking about the same, but then I would need to move the logger initialization out to fit the logger: https://github.com/mongodb/mongodb-kubernetes-operator/blob/a7812ff0728930f2fb35ae4a9776b80a86e2e64b/pkg/readiness/config/config.go#L59-L64

I **could** do this 🤔

slaskawi: How about panicing with an error? This way it should be visible in the `kubectl get events` and `kubectl describe pod`.

slaskawi: Thanks for the explanation @nammn ! I believe it's still worth logging an info why this happened. Without it, investigating corner cases might be simply impossible.

nammn: Done

slaskawi (approved):

PR #1513, opened 2024-03-21
CLOUDP-204105 Conditionally run the upgrade hook

slaskawi:
# Summary

This Pull Requests runs the upgrade hook only if it exists (and it doesn't in Static Architecture)


nammn (approved):
do we need to bump something in the release.json?

slaskawi:
@nammn No, no - it's fine as is.

PR #1533, opened 2024-04-23
Sign images

Julien-Ben:
## Introduction

This PR introduces signature for all container images released for the community operator.

## How it works

After building and pushing an image or manifest to a registry, we fetch its digest using skopeo.
Then we sign this image digest with cosign, and the signature is pushed to the same registry.
We also do a signature verification step against our public key.

All images built during releases (-context) are signed.

## Example 

The readiness probe images with tag `1.0.18` are signed, this is visible with an icon next to the tag name.
https://quay.io/repository/mongodb/mongodb-kubernetes-readinessprobe?tab=tags

## Other changes

Refactor inventories to remove a lot of duplication (architecture as parameter).
Use a custom logger instead of prints.

### DIFF
--- a/.action_templates/jobs/setup.yaml
+++ b/.action_templates/jobs/setup.yaml
@@ -4,7 +4,7 @@ setup:
     fail-fast: false
     matrix:
       include:
-        - pipeline-argument: mongodb-kubernetes-operator
+        - pipeline-argument: operator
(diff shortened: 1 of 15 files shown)


mircea-cosbuc (approved):
LGTM

PR #1558, opened 2024-06-05
CLOUDP-252271 Use gosec linter

slaskawi:
This Pull Request introduces gosec through golangci-lint project. As a side-effect, it removes super-linter integration.

The motivation behind integrating it this way is using a single `//nolint` comment to disable (or mark as false-positive) all linters and not using linter-specific markers (such as `nosec`).

### DIFF
--- a/test/e2e/replica_set_arbiter/replica_set_arbiter_test.go
+++ b/test/e2e/replica_set_arbiter/replica_set_arbiter_test.go
@@ -69,9 +69,9 @@ func TestReplicaSetArbiter(t *testing.T) {
 			resourceName:     "mdb4",
 		},
 	}
-	for testName, _ := range tests {
+	for testName := range tests {
 		t.Run(testName, func(t *testing.T) {
-			testConfig, _ := tests[testName]
+			testConfig := tests[testName]
 			mdb, user := e2eutil.NewTestMongoDB(testCtx, testConfig.resourceName, "")
 			mdb.Spec.Arbiters = testConfig.numberOfArbiters
 			mdb.Spec.Members = testConfig.numberOfMembers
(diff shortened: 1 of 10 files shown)


test/e2e/replica_set_arbiter/replica_set_arbiter_test.go
```diff
@@ -69,9 +69,9 @@ func TestReplicaSetArbiter(t *testing.T) {
 			resourceName:     "mdb4",
 		},
 	}
-	for testName, _ := range tests {
```
nammn, line 72: is the linter complaining over explicit empty assigns?

nammn (approved):
changes lgtm, but I am interested in the reasons why we have those changes

test/e2e/replica_set_operator_upgrade/replica_set_operator_upgrade_test.go
```diff
@@ -73,7 +73,7 @@ func TestReplicaSetOperatorUpgrade(t *testing.T) {
 // TestReplicaSetOperatorUpgradeFrom0_7_2 is intended to be run locally not in CI.
 // It simulates deploying cluster using community operator 0.7.2 and then upgrading it using newer version.
 func TestReplicaSetOperatorUpgradeFrom0_7_2(t *testing.T) {
-	ctx := context.Background()
+	ctx := context.Background() //nolint
```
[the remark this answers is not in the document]
slaskawi: The linter though it's not used anywhere. Obviously a false-positive (as it is used a few lines below - it's passed to other methods).

test/e2e/replica_set_arbiter/replica_set_arbiter_test.go
slaskawi: Yeah. it is. I'm not sure why but apparently it doesn't like when you leave the second one empty in a `for`. Kinda makes sense I guess.

PR #1589, opened 2024-07-18
Improvements for local development setup process

MateiGrigore:
### Summary:

This PR will improve the setup process for local development of the mongodb-kubernetes-operator. The documentation contained outdated and misleading information.  Also some problems with building images and deploying the operator were encountered because of using a local registry. This

The PR adds:
- Documentation changes to improve the local setup process
- Changes to the `Makefile`:
    - Change the name of the parameter to look for the agent image in the `config.json` file to `agent_image`
    - Add `--ignore-not-found` flag to `uninstall-crd` target because of errors encountered when calling this during the `make deploy` command where no crd was installed yet
    - Add `$(IMG_BUILD_ARGS)` to all image building targets to allow the `--insecure` flag to be passed as a paramter
- Changes to the `inventory.yaml` file:
    - Removed default values for `agent_distro` and `tools_distro`
    - Changed the values for `agent_distro` and `tools_distro` in buildargs to be extracted from the params
- Changes to the `pipeline.py` file:
    - Allow for a new argument `insecure` to be passed
- Chnages to the `setup_kind_cluster.sh`:
    - Add the registry config to the nodes

### All Submissions:

### DIFF
--- a/docs/run-operator-locally.md
+++ b/docs/run-operator-locally.md
@@ -18,17 +18,23 @@ Being able to run and build the binary locally can help with faster feedback-cyc
 - Run e2e tests locally
 
 ## Running The Operator locally
-1. Use the dedicated make target which exports the needed environment variables and builds & runs the operator binary
+1. Use the dedicated make target which exports the needed environment variables and builds & runs the operator binary.
 
-```sh
-make run
-```
+   Before doing that you need to add 2 more fields to the `config.json` file from in [contributing.md](contributing.md):
+    - `mdb_local_operator`: needs to be set to `true`
+    - `kubeconfig`: needs to be set to the path of the configuration file mentioned above
+  
+   Then you can run the command:
+  
+    ```sh
+       make run
+    ```
 
 2.  For debugging one can use the following make target, which uses [dlv](https://github.com/go-delve/delve):
 
-```sh
-make debug
-```
+    ```sh
+    make debug
+    ```
 
 ## Running e2e tests with the local operator
 - Our [e2e tests](../test/e2e), contains sub-steps that will install the following helm-chart: [operator.yaml](../helm-charts/charts/community-operator/templates/operator.yaml)
--- a/docs/run-operator-locally.md
+++ b/docs/run-operator-locally.md
@@ -20,9 +20,9 @@ Being able to run and build the binary locally can help with faster feedback-cyc
 ## Running The Operator locally
 1. Use the dedicated make target which exports the needed environment variables and builds & runs the operator binary.
 
-   Before doing that you need to add 2 more fields to the `config.json` file from in [contributing.md](contributing.md):
-    - `mdb_local_operator`: needs to be set to `true`
-    - `kubeconfig`: needs to be set to the path of the configuration file mentioned above
+   Before doing that you need to add 2 more fields to the `config.json` file found in [contributing.md](contributing.md), because the python script looks for them in the file:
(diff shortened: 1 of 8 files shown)


inventory.yaml
```diff
@@ -9,9 +9,6 @@ images:
     vars:
       context: .
       template_context: scripts/dev/templates/agent
-      # Default values but overwritten in pipeline.py
```
Julien-Ben, line 12: Did these default values caused troubles in your setup ?

pipeline.py
```diff
@@ -87,6 +87,7 @@ def build_and_push_image(
     architectures: Set[str],
     release: bool,
     sign: bool,
+    insecure: bool,
```
Julien-Ben, line 90: I think we should also add a default here.

Julien-Ben (approved):
Very nice setup improvements ! 👏 
Left a few comments but LGTM overall
(please wait for others to have a look before merging)

docs/run-operator-locally.md
```diff
@@ -18,17 +18,23 @@ Being able to run and build the binary locally can help with faster feedback-cyc
...
-1. Use the dedicated make target which exports the needed environment variables and builds & runs the operator binary
-
-```sh
-make run
-```
+1. Use the dedicated make target which exports the needed environment variables and builds & runs the operator binary.
+
+   Before doing that you need to add 2 more fields to the `config.json` file from in [contributing.md](contributing.md):
+    - `mdb_local_operator`: needs to be set to `true`
+    - `kubeconfig`: needs to be set to the path of the configuration file mentioned above
```
nammn, line 25: can you clarify this part?

inventory.yaml
nammn: What did exactly happen? What was the error?

scripts/dev/setup_kind_cluster.sh
```diff
@@ -72,10 +72,26 @@ networking:
   serviceSubnet: "${service_network}"
 containerdConfigPatches:
 - |-
-  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:${reg_port}"]
-    endpoint = ["http://${reg_name}:${reg_port}"]
+  [plugins."io.containerd.grpc.v1.cri".registry]
+    config_path = "/etc/containerd/certs.d"
```
nammn, line 76: can you explain to me this change and why this is necessary? In meko we are not doing this and it works fine

nammn: I personally need more information why we are doing this and what it does before approving. I won't block the PR but if it can wait I prefer to let this one open until mircea is back

inventory.yaml
nammn: isn't that because we hardcoded the tools_distr and agent_distro?
But wouldn't it work if you still keep the defaults? 😕 

Nevertheless, works for me!

nammn: yes, we actually had a bug! We pulled that fix out https://github.com/mongodb/mongodb-kubernetes-operator/pull/1596

scripts/dev/setup_kind_cluster.sh
mircea-cosbuc: This was basically updating the local registry setup using the official docs guide: https://kind.sigs.k8s.io/docs/user/local-registry/ . This is better because it's not proxying the registry it actually makes the kind node aware of the localhost registry.

PR #1623, opened 2024-09-28
Changed the order of CreateOrUpdate functions

lsierant:
This PR changes the way we do `CreateOrUpdate` functions. 

It was: Get the Update if exists else Create. 
For some cases after the resource was created in the same reconcile event, another execution will again hit into Create case as the informer caches weren't yet refreshed. And that resulted in errors.

### DIFF
--- a/pkg/kube/configmap/configmap.go
+++ b/pkg/kube/configmap/configmap.go
@@ -85,14 +85,14 @@ func UpdateField(ctx context.Context, getUpdater GetUpdater, objectKey client.Ob
 // CreateOrUpdate creates the given ConfigMap if it doesn't exist,
 // or updates it if it does.
 func CreateOrUpdate(ctx context.Context, getUpdateCreator GetUpdateCreator, cm corev1.ConfigMap) error {
-	_, err := getUpdateCreator.GetConfigMap(ctx, types.NamespacedName{Name: cm.Name, Namespace: cm.Namespace})
-	if err != nil {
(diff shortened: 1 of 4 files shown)


mircea-cosbuc (approved):

Julien-Ben (approved):

MaciejKaras (approved):

lucian-tosa (approved):

pkg/kube/configmap/configmap.go
```diff
@@ -85,14 +85,14 @@ func UpdateField(ctx context.Context, getUpdater GetUpdater, objectKey client.Ob
 // CreateOrUpdate creates the given ConfigMap if it doesn't exist,
 // or updates it if it does.
 func CreateOrUpdate(ctx context.Context, getUpdateCreator GetUpdateCreator, cm corev1.ConfigMap) error {
-	_, err := getUpdateCreator.GetConfigMap(ctx, types.NamespacedName{Name: cm.Name, Namespace: cm.Namespace})
-	if err != nil {
+	if err := getUpdateCreator.UpdateConfigMap(ctx, cm); err != nil {
 		if apiErrors.IsNotFound(err) {
 			return getUpdateCreator.CreateConfigMap(ctx, cm)
+		} else {
```
nammn, line 91: i am a bit late to the party but we could have saved the else and just return, right?

Julien-Ben: When I see such if/else I always wonder if I should leave a comment
Should we try to be consistent in the whole codebase about that ?
We use both currently

PR #1676, opened 2025-02-19
CLOUDP-302068: Refactoring: propagate env vars via parameters

m1kola:
### Summary:

This will help with code reusability.

### All Submissions:

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -3,11 +3,13 @@ package main
 import (
 	"fmt"
 	"os"
+
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 
 	mdbv1 "github.com/mongodb/mongodb-kubernetes-operator/api/v1"
 	"github.com/mongodb/mongodb-kubernetes-operator/controllers"
 	"github.com/mongodb/mongodb-kubernetes-operator/controllers/construct"
+	"github.com/mongodb/mongodb-kubernetes-operator/pkg/util/envvar"
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -56,7 +58,14 @@ func main() {
 		log.Sugar().Fatalf("Failed to configure logger: %v", err)
 	}
 
-	if !hasRequiredVariables(log, construct.AgentImageEnv, construct.VersionUpgradeHookImageEnv, construct.ReadinessProbeImageEnv) {
+	if !hasRequiredVariables(
+		log,
+		construct.MongodbRepoUrlEnv,
+		construct.MongodbImageEnv,
+		construct.AgentImageEnv,
+		construct.VersionUpgradeHookImageEnv,
+		construct.ReadinessProbeImageEnv,
+	) {
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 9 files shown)


cmd/manager/main.go
```diff
@@ -56,7 +58,14 @@ func main() {
 		log.Sugar().Fatalf("Failed to configure logger: %v", err)
 	}
 
-	if !hasRequiredVariables(log, construct.AgentImageEnv, construct.VersionUpgradeHookImageEnv, construct.ReadinessProbeImageEnv) {
+	if !hasRequiredVariables(
+		log,
+		construct.MongodbRepoUrlEnv,
+		construct.MongodbImageEnv,
```
m1kola, line 64: This is probably the most significant change in this PR: rest is moving things around and updating tests where applicable.

These env vars were not required before, but from the code it seems like they actually are required. So I'm adding them here. Would be good to get a sanity check here. Am I missing something? Is it actually possible and reasonable to run without them?

controllers/construct/mongodbstatefulset.go
```diff
@@ -123,7 +126,7 @@ type MongoDBStatefulSetOwner interface {
 // BuildMongoDBReplicaSetStatefulSetModificationFunction builds the parts of the replica set that are common between every resource that implements
 // MongoDBStatefulSetOwner.
 // It doesn't configure TLS or additional containers/env vars that the statefulset might need.
-func BuildMongoDBReplicaSetStatefulSetModificationFunction(mdb MongoDBStatefulSetOwner, scaler scale.ReplicaSetScaler, agentImage string, withInitContainers bool) statefulset.Modification {
+func BuildMongoDBReplicaSetStatefulSetModificationFunction(mdb MongoDBStatefulSetOwner, scaler scale.ReplicaSetScaler, mongodbRepoUrl, mongodbImage, mongodbImageType, agentImage, versionUpgradeHookImage, readinessProbeImage string, withInitContainers bool) statefulset.Modification {
```
m1kola, line 129: Can probably replace `mongodbRepoUrl`, `mongodbImage` and `mongodbImageType` with just `mongodbImage`.

And have `mongodbImage := getMongoDBImage(...)` somewhere higher up the call hierarchy.

controllers/replica_set_controller.go
```diff
@@ -60,7 +60,7 @@ func init() {
 	zap.ReplaceGlobals(logger)
 }
 
-func NewReconciler(mgr manager.Manager) *ReplicaSetReconciler {
+func NewReconciler(mgr manager.Manager, mongodbRepoUrl, mongodbImage, mongodbImageType, agentImage, versionUpgradeHookImage, readinessProbeImage string) *ReplicaSetReconciler {
```
Julien-Ben, line 63: What do you think of creating a structure to hold the controller configuration like you did in MEKO ?

m1kola: @Julien-Ben in MEKO `imageUrls` struct was necessary because we had to handle `RELATED_IMAGE_*` env vars which are not a static set. In here we have a set of images which are known in advance - I think it is a bit easier to follow without a map (without `imageUrls`).

But we might want to reconsider later.

nammn (approved):

Julien-Ben (approved):

PR #1687, opened 2025-03-07
Bump GitHub Actions

m1kola:
actions/cache@v2 is now depricated and fails CI


m1kola:
I don't get it. It still complains about actions/cache v2 but there is no v2 version anywhere.

PR #1690, opened 2025-03-12
CLOUDP-302068: Add a linter to forbid env var usage

m1kola:
### Summary:

We want to avoid spreading the practice of accessing environment variables deep in the call hierarchy. We only want to access env vars in the `main` package or very close to it.

This commit adds a linter to help us with that. It also adds `nolint:forbidigo` exceptions into places where we still unfortunately have access. These places ideally need to change with time.

To find remaining undesired places, from repo root you can `grep` all files for `nolint:forbidigo`. For example: `grep "nolint:forbidigo" -r .`

### All Submissions:

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -44,7 +44,7 @@ func configureLogger() (*zap.Logger, error) {
 func hasRequiredVariables(logger *zap.Logger, envVariables ...string) bool {
 	allPresent := true
 	for _, envVariable := range envVariables {
-		if _, envSpecified := os.LookupEnv(envVariable); !envSpecified {
+		if _, envSpecified := os.LookupEnv(envVariable); !envSpecified { // nolint:forbidigo
 			logger.Error(fmt.Sprintf("required environment variable %s not found", envVariable))
 			allPresent = false
 		}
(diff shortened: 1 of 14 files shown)


Julien-Ben (approved):

cmd/manager/main.go
```diff
@@ -44,7 +44,7 @@ func configureLogger() (*zap.Logger, error) {
 func hasRequiredVariables(logger *zap.Logger, envVariables ...string) bool {
 	allPresent := true
 	for _, envVariable := range envVariables {
-		if _, envSpecified := os.LookupEnv(envVariable); !envSpecified {
+		if _, envSpecified := os.LookupEnv(envVariable); !envSpecified { // nolint:forbidigo
```
nammn, line 47: q: I thought we are fine with all files named `main.go` to keep the env var usage? If yes, could we exclude the whole file instead of single lines? Or is that on purpose here?

nammn (commented):
LGTM! Thanks! One Q

cmd/manager/main.go
m1kola: @nammn I considered doing this, but as far as I understand it is only possible to exclude whole linter. It is not an issue right now because we only use `forbidigo` only for env vars right now. But in the future we decide to use `forbidigo` for something else `main.go` will become a blind spot for this linter.

What do you think - still worth adding an exception for main.go?

nammn: I think its better to skip the whole file for now and in case we add more linters we can revisit this decision.

nammn (approved):

